PR #13, opened 2020-02-09
Android glue

### LINKED ISSUE #10: Unnecessary bindings for target_arch = "armv7"
As I know in rust `target_arch = "arm"` actually corresponding to _armv7_, so second bindings completely useless.
https://github.com/rust-windowing/android-ndk-rs/blob/a8ddfa4f7b510923cd69ded0f7e8492711b0af88/android-ndk-sys/src/lib.rs#L21-L28
Fix me if I'm wrong.

dvc94ch:
The `native-app-glue` provided by the `ndk` which is used by `cargo-apk` is buggy. This PR implements the required parts in rust and rewrites `cargo-apk`, because it's overly complicated and hard to maintain.

- https://github.com/rust-lang/cargo/issues/4881
- https://github.com/rust-lang/cargo/issues/2930
- https://github.com/rust-lang/cargo/issues/5364

Closes #10


dvc94ch:
cc @katyo

android-build-tools/src/apk.rs
```diff
@@ -0,0 +1,191 @@
...
+            path: config.apk(),
+            package_name: config.manifest.package_name.clone(),
+            ndk,
+        }
+    }
+
+    pub fn install(&self) -> Result<(), NdkError> {
+        self.ndk
+            .platform_tool("adb")?
+            .arg("install")
```
katyo: The attempt to reinstall already installed app on older Android will fail with error like the follow:
```
Failure [INSTALL_FAILED_ALREADY_EXISTS]
```
As a simple workaround I suggest to add '-r' flag after 'install' command.

dvc94ch:
Got another batch of refactorings. Want to make sure the API works for the flutter-rs multiapk setup I'm working on first. In terms naming, should we rename cargo-ndk to cargo-apk, as the name cargo-ndk is already taken and archive the android-rs-glue repo?

katyo:
I agree, the cargo-ndk should be named cargo-apk (or cargo-android).

Also I would like to implement some additional features. For example, the ability to extend Native Activity needed for things missing in NDK like IME interfacing and etc.

cargo-ndk/src/apk.rs
```diff
@@ -0,0 +1,113 @@
...
+        Ok(())
+    }
+
+    pub fn build(&self, artifact: &Artifact) -> Result<Apk, NdkError> {
+        let config = Config {
+            ndk: self.ndk.clone(),
+            build_dir: self.build_dir.join(artifact),
+            package_name: package_name(artifact),
+            package_label: artifact.name().to_string(),
+            version_name: self.manifest.version.clone(),
```
katyo: I think we should support setting apk version using `version_code`/`version_name` (from `package.metadata.android`).
In case when this fields is not set we simply generate it from cargo package metadata.

cargo-ndk/src/apk.rs
```diff
@@ -0,0 +1,113 @@
...
+            version_code: 1,
```
katyo: According to the [docs](https://developer.android.com/guide/topics/manifest/manifest-element.html#vcode) this number is used to to determine that one version is more recent than another. So I think it should be either set directly from metadata or generated in more complicated way.

cargo-ndk/src/apk.rs
katyo: Oh it seems this is defaults. Okay.

cargo-ndk/src/apk.rs
katyo: Ok.

cargo-ndk/src/apk.rs
dvc94ch: The reason why the package name and version info can't be set in metadata.android is because it's unclear what artifact they refer to (how can we set the package_name of an example) and in a more complex setup you may want to create multiple apks from the same crate (instead of a fat apk for example) with multiple version codes.

katyo:
I tested android-no-glue example on device with target sdk 18 and it fails with:
```
java.lang.IllegalArgumentException: Unable to load native library: /data/app-lib/rust.android_no_glue-2/libandroid_no_glue.so
```
It seems the native activity does not give extra info about why loading fails.
There is known [workaround](https://stackoverflow.com/questions/22686900/java-lang-illegalargumentexception-unable-to-load-native-library#answer-22729423) to get this extra info.
But it requires the possibility to build some java code.

android-build-tools/src/config.rs
```diff
@@ -0,0 +1,60 @@
...
+    pub(crate) opengles_version: Option<(u8, u8)>,
+    pub(crate) feature: Option<Vec<FeatureConfig>>,
+    pub(crate) permission: Option<Vec<PermissionConfig>>,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+pub(crate) struct FeatureConfig {
+    name: String,
+    required: Option<bool>,
+    version: Option<String>,
```
katyo: Are what the `version` field means?
I cannot find this in docs.

cargo-ndk/src/apk.rs
dvc94ch: How about using `<version_code> = <apk_id><major><minor><patch>` as the default? That would work for most projects, where the project can compute it's `<apk_id>` as the product of it's build matrix: `arch * dpi * target_sdk_version`

android-build-tools/src/config.rs
dvc94ch: nice catch, I looked at the `cargo-apk` sources when writing the code

goddessfreya:
Hmmm, so you folks want to name it `cargo-apk`? I'd have to bug @tomaka to get publishing rights for the crate, so that might take a while.

android-build-tools/src/apk.rs
```diff
@@ -0,0 +1,191 @@
...
+        let file_name = path.file_name().unwrap();
+        let out = self.0.build_dir.join("lib").join(abi);
+        std::fs::create_dir_all(&out)?;
+        std::fs::copy(path, out.join(&file_name))?;
+
+        self.0
+            .build_tool("aapt")?
+            .arg("add")
+            .arg(self.0.unaligned_apk())
+            .arg(format!("lib/{}/{}", abi, file_name.to_str().unwrap()))
```
katyo: Usually you should join the path components using `Path::join`.

dvc94ch: I'm not sure how aapt parses it on windows. It must be exactly this string for it to work, an absolute path for example breaks it

cargo-apk/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "cargo-ndk"
+name = "cargo-apk"
 version = "0.1.0"
```
goddessfreya: Already published, pick a newer version.

dvc94ch:
We probably also need to bug @mb64 for publishing the other crates... Anyone know if he is still alive? :)

cargo-ndk/src/apk.rs
katyo: LGTM.

dvc94ch:
There are two remaining issues. `apksigner` has a `.bat` and not a `.exe` extension on windows and if a command fails we continue as we don't check the exit status anywhere. Other than that are @goddessfreya @katyo ok to merge?

katyo:
@dvc94ch I think it will be good to print some usage/help when __cargo apk__ called without args and/or with '--help' arg.

katyo:
@dvc94ch As I see @mb64 completely inactive on github under last three month. It seems it will be problematic to communicate with him.

dvc94ch:
@katyo I updated cargo-subcommand to use `--help` as the command if none is supplied. And prefixed a custom message when on `--help`.

katyo:
@dvc94ch Hmm, it seems this doesn't work as expected for me.

dvc94ch:
What's your expectation and what is happening?

katyo:
I expect:
```
$ cargo apk --help
cargo-apk helps cargo build apk's for android

<cargo apk subcommands with usage hints>
```
Actual:
```
$ cargo apk --help
cargo-apk helps cargo build apk's for android

<cargo usage help page>
```

android-glue/src/lib.rs
```diff
@@ -0,0 +1,278 @@
...
+            tag.as_ptr() as *const _,
+            msg.as_ptr() as *const _,
+        );
+    }
+}
+
+lazy_static! {
+    static ref NATIVE_WINDOW: RwLock<Option<NativeWindow>> = Default::default();
+    static ref INPUT_QUEUE: RwLock<Option<InputQueue>> = Default::default();
+    static ref CONTENT_RECT: RwLock<Rect> = Default::default();
```
katyo: What the reason you select RwLock rather than Mutex?

katyo:
We should remove unnecessary 'armv7' bindings from _android-ndk-sys_:
```
#[cfg(all(any(target_os = "android", test), target_arch = "armv7"))]
include!("ffi_armv7.rs");
```
See #10

android-ndk-sys/src/lib.rs
```diff
@@ -36,3 +36,7 @@ include!("ffi_i686.rs");
 
 #[cfg(all(any(target_os = "android", test), target_arch = "x86_64"))]
 include!("ffi_x86_64.rs");
+
+#[cfg(target_os = "android")]
+#[link(name = "android")]
+extern "C" {}
```
katyo: What about link with _log_ too?

dvc94ch: log is not required as it's not used in this crate

dvc94ch: log is not required by this crate, it's added in `android-glue`.

android-glue/src/lib.rs
katyo: Actually I don't sure what is the best choise in this case.

dvc94ch: It's a `RwLock` because the input thread is the only one that should write, while multiple consumers can read and block the input thread concurrently. If it's a mutex there is no way of preventing consumers from writing to it.

katyo:
Help looks good to me.

katyo:
I choice target archs like that:
```
build_targets = ["armv7-linux-androideabi"]
```
But it seems this settings does not applied:
```
$ cargo apk build
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
 'lib/arm64-v8a/libandroid_no_glue.so'...
Verifying alignment of ...android/android-no-glue/target/debug/apk/android_no_glue.apk (4)...
      49 AndroidManifest.xml (OK - compressed)
     963 lib/arm64-v8a/libandroid_no_glue.so (OK - compressed)
Verification succesful
```

And secondly, I don't understand how the split can be configured?

katyo:
Actually armv7 bindings should be removed. See my explanation at #10.

android-ndk-sys/src/lib.rs
```diff
@@ -20,14 +20,14 @@ pub mod native_app_glue;
 compile_error!("android-ndk-sys only supports compiling for Android");
 
 #[cfg(any(
-    all(any(target_os = "android", test), target_arch = "arm"),
+    all(
+        any(target_os = "android", test),
+        any(target_arch = "arm", target_arch = "armv7")
```
katyo: You can see the actual compiler specs here:
https://github.com/rust-lang/rust/blob/master/src/librustc_target/spec/arm_linux_androideabi.rs#L15
https://github.com/rust-lang/rust/blob/master/src/librustc_target/spec/armv7_linux_androideabi.rs#L23
The arch for both is "arm" (nothing of that "armv7"). Right?

katyo:
The build targets choicing works well :)

android-glue/src/lib.rs
katyo: Ok, I understand.

katyo:
It seems we quite near to finish :)

dvc94ch:
I'm not convinced about the bindings being superfluous only because they happen to be identical.We should not check the data layout manually, the target triple is the abstraction that the compiler gives us.

android-build-tools/src/ndk.rs
```diff
@@ -0,0 +1,318 @@
...
+            } else {
+                PathBuf::from(ndk_path.ok_or(NdkError::NdkNotFound)?)
+            }
+        };
+
+        let build_tools_dir = sdk_path.join("build-tools");
+        let build_tools_version = std::fs::read_dir(&build_tools_dir)
+            .or(Err(NdkError::PathNotFound(build_tools_dir)))?
+            .filter_map(|path| path.ok())
+            .filter(|path| path.metadata().unwrap().is_dir())
```
katyo: To avoid problems with symlinks instead `path.metadata().unwrap().is_dir()` we should do `path.path().is_dir()`.

android-build-tools/src/ndk.rs
```diff
@@ -0,0 +1,318 @@
...
+            .filter_map(|path| path.file_name().into_string().ok())
+            .filter(|name| name.chars().next().unwrap().is_digit(10))
+            .max()
+            .ok_or(NdkError::BuildToolsNotFound)?;
+
+        let platforms_dir = sdk_path.join("platforms");
+        let platforms: Vec<u32> = std::fs::read_dir(&platforms_dir)
+            .or(Err(NdkError::PathNotFound(platforms_dir)))?
+            .filter_map(|path| path.ok())
+            .filter(|path| path.metadata().unwrap().is_dir())
```
katyo: `path.path().is_dir()`

android-build-tools/src/readelf.rs
```diff
@@ -0,0 +1,113 @@
+use crate::apk::UnalignedApk;
+use crate::error::NdkError;
+use crate::target::Target;
+use std::collections::HashSet;
+use std::io::BufRead;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+pub trait Readelf {
```
katyo: Are the reason to have additional trait for that?
I think that enough add extra impl for `UnalignedApk` itself.

```rust
impl<'a> UnalignedApk<'a> {
    fn add_lib_recursively ...
}
```

android-build-tools/src/apk.rs
```diff
@@ -131,6 +133,47 @@ impl<'a> UnalignedApk<'a> {
         Ok(())
     }
 
+    pub fn add_lib_recursively(
```
katyo: Rust supports multiple impl blocks for struct so it can be stay decomposed (as readelf.rs) to keep code clear.

dvc94ch: well I don't think that works accross files but I might be wrong. if the reader has to look for the method because it's somewhere else, I'm not sure that's an improvement

katyo: It works and I use it to decompose structs with huge number of functions.

katyo:
I implement platform code compilation feature (i.e. java/kotlin).
Is the _android-build-tools_ API stable?
I would like to rebase to current head of this pr.

dvc94ch:
android-build-tools shouldn't change anymore

android-build-tools/src/apk.rs
```diff
@@ -0,0 +1,207 @@
...
+pub struct ApkConfig {
+    pub ndk: Ndk,
+    pub build_dir: PathBuf,
+    pub assets: Option<String>,
+    pub res: Option<String>,
+    pub manifest: Manifest,
+}
+
+impl ApkConfig {
+    pub fn from_config(config: Config, metadata: Metadata) -> Self {
```
katyo: Maybe we should pass `Config` and `Metadata` by reference to avoid cloning.

dvc94ch: why would you want to keep them around? `ApkConfig`'s members are pub and the `Manifest` also is pub.

dvc94ch:
should we rename `android-ndk-sys` to `ndk-sys` and `android-ndk` to `ndk`? `rust-windowing` owns `cargo-apk`.

katyo:
Hmm... sounds not so clear. But I think this is not so matter.

README.md
```diff
@@ -1,18 +1,66 @@
...
- * [`android_logger`](https://crates.io/crates/android_logger) and [`ndk-logger`](https://crates.io/crates/ndk-logger),
-   Android backends for the `log` crate
+pub fn app_main() {
+    println!("hello world");
+}
+```
+
+```sh
+cargo install cargo-apk
+cargo apk run
```
goddessfreya: Alright, so yesterday I tried this, which didn't work:
```
  ~/Do…ts/gfx/eco3/glutin                                                                                                                           SIGINT(2) ↵  69%  
  12:39  17.02.20  cargo apk run --example window
   Compiling glutin v0.23.0-alpha1 (/home/gentz/Documents/gfx/eco3/glutin)
    Finished dev [unoptimized + debuginfo] target(s) in 6.00s
Error: Path "/home/gentz/Documents/gfx/eco3/glutin/target/aarch64-linux-android/debug/examples/libwindow.so" doesn't exist.
```

You mention it needs to be cdylib, so I tried this.

```
  ~/Do…ts/gfx/eco3/glutin                                                                                                                      1 ↵  6.41981s   69%  
  12:39  17.02.20  RUSTFLAGS="--crate-type cdylib" cargo apk run --example window  
   Compiling cfg-if v0.1.10
   Compiling lazy_static v1.4.0
   Compiling scopeguard v1.1.0
   Compiling adler32 v1.0.4
error: `#[panic_handler]` function required, but not found

error: language item required, but not found: `eh_personality`

error: aborting due to 2 previous errors

error: could not compile `cfg-if`.
warning: build failed, waiting for other jobs to finish...
error: `#[panic_handler]` function required, but not found

error: language item required, but not found: `eh_personality`

error: aborting due to 2 previous errors

error: could not compile `scopeguard`.
warning: build failed, waiting for other jobs to finish...
error: build failed
Error: Command 'cargo build --target aarch64-linux-android --example window' had a non-zero exit code.
```

I did some googling, and this gets rid of the `eh_personality` error:
```
  ~/Do…ts/gfx/eco3/glutin                                                                                                                      1 ↵  1.15127s   69%  
  12:40  17.02.20  RUSTFLAGS="--crate-type cdylib -C panic=abort" cargo apk run --example window
   Compiling cfg-if v0.1.10
   Compiling lazy_static v1.4.0
   Compiling scopeguard v1.1.0
   Compiling adler32 v1.0.4
error: `#[panic_handler]` function required, but not found

error: aborting due to previous error

error: could not compile `cfg-if`.
warning: build failed, waiting for other jobs to finish...
error: `#[panic_handler]` function required, but not found

error: aborting due to previous error

error: could not compile `scopeguard`.
warning: build failed, waiting for other jobs to finish...
error: build failed
Error: Command 'cargo build --target aarch64-linux-android --example window' had a non-zero exit code.
```

Implementing panic handler on a function wouldn't fix that error, though, so this readme leaves me kind of stumped :/

dvc94ch: This works for me:
```diff
diff --git a/Cargo.toml b/Cargo.toml
index 3898294..b2c6232 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,3 +74,7 @@ glutin_glx_sys = { version = "0.2.0", path = "../glutin_sys/glutin_glx_sys" }
 parking_lot = "0.10.0"
 glutin_x11_sym = { version = "0.1.0", path = "../glutin_x11_sym" }
 x11-dl = "2.18.4"
+
+[[example]]
+name = "window"
+crate-type = ["cdylib"]
```

dvc94ch: The rustflags env the way you set it gives me the same error

dvc94ch: The problem seems to be that it tries to compile all crates as cdylib, when only the example needs to be cdylib. Trying to compile a proc-macro as cdylib is no good.

katyo:
It seems something wrong with linking.
```
E/dalvikvm(30700): dlopen("/data/app-lib/rust.android_no_glue-1/libandroid_no_glue.so") failed: dlopen failed: empty/missing DT_HASH in "libandroid_no_glue.so" (built with --hash-style=gnu?)
```
This issue actual for android platform 18 and possibly for some other.

dvc94ch:
@katyo are you using the latest commit? There was a regression that could have caused that due to changes in `cargo-subcommand`.

README.md
dvc94ch: also note that it'll crash in any case. we could add support to the ios-example and rename it to mobile-example?

goddessfreya: Well, that sucks. I guess we could follow this suggestion I stole off of reddit: https://www.reddit.com/r/rust/comments/64hlfa/how_to_compile_a_crate_as_dynamic_library/dg2rq64/
I tried the following, but:
```
$ rm android -rf; mkdir android; cat Cargo.toml | sed "s/\.\.\//\.\.\/\.\.\//g" > android/Cargo.toml; ls examples/ | grep rs | awk '{sub(/\..*/,"",$1); print "\n[[example]]\nname = \"" $0 "\"\ncrate-type = [\"cdylib\"]\npath = \"../examples/" $0 ".rs\""}' >> android/Cargo.toml; printf '\n[lib]\npath = "../src/lib.rs"\n' >> android/Cargo.toml; cargo apk run --example window --manifest-path android/Cargo.toml
ERROR: unable to open 'android/target/debug/apk/examples/window-unaligned.apk' as Zip file for writing
    Updating crates.io index
error: no matching package named `ndk` found
location searched: registry `https://github.com/rust-lang/crates.io-index`
required by package `winit v0.21.0 (/home/gentz/Documents/gfx/eco3/winit)`
    ... which is depended on by `glutin v0.23.0-alpha1 (/home/gentz/Documents/gfx/eco3/glutin/android)`
Error: Command 'cargo build --target armv7-linux-androideabi --example window --manifest-path android/Cargo.toml' had a non-zero exit code.
```

```
$ cat android/Cargo.toml         
[package]
name = "glutin"
version = "0.23.0-alpha1"
authors = ["The glutin contributors", "Pierre Krieger <pierre.krieger1708@gmail.com>"]
description = "Cross-platform OpenGL context provider."
keywords = ["windowing", "opengl"]
license = "Apache-2.0"
readme = "../../README.md"
repository = "https://github.com/rust-windowing/glutin"
documentation = "https://docs.rs/glutin"
edition = "2018"
build = "build.rs"

[package.metadata.docs.rs]
features = ["serde_feature"]

[features]
serde_feature = ["winit_types/serde_feature"]

[dependencies]
lazy_static = "1.4.0"
winit_types = { version = "0.1.0", path = "../../winit_types" }
glutin_interface = { version = "0.1.0", path = "../../glutin_interface" }
log = "0.4.8"
libloading = "0.5.2"

[build-dependencies]
gl_generator = "0.14.0"

[dev-dependencies]
winit = { path = "../../winit" }
takeable-option = "0.5.0"
image = "0.22.4"
simple_logger = "1.4.0"

[target.'cfg(all(any(target_os = "linux", target_os = "freebsd", target_os = "dragonfly", target_os = "netbsd", target_os = "openbsd"), target_arch="x86_64"))'.dev-dependencies]
gbm = { path = "../../gbm.rs" }
drm = "0.3.4"

[target.'cfg(target_os = "android")'.dependencies]
glutin_egl_sys = { version = "0.1.4", path = "../../glutin_sys/glutin_egl_sys" }
parking_lot = "0.10.0"

[target.'cfg(any(target_os = "ios", target_os = "macos"))'.dependencies]
objc = "0.2.7"
glutin_gles2_sys = { version = "0.1.3", path = "../../glutin_sys/glutin_gles2_sys" }

[target.'cfg(target_os = "macos")'.dependencies]
cgl = "0.3.2"
cocoa = "0.20.0"
core-foundation = "0.7.0"
core-graphics = "0.19.0"

[target.'cfg(target_os = "windows")'.dependencies]
libloading = "0.5.2"
glutin_wgl_sys = { version = "0.1.3", path = "../../glutin_sys/glutin_wgl_sys" }
glutin_egl_sys = { version = "0.1.5", path = "../../glutin_sys/glutin_egl_sys" }
parking_lot = "0.10.0"

[target.'cfg(target_os = "windows")'.dependencies.winapi]
version = "0.3.8"
features = [
    "winnt",
    "winuser",
    "wingdi",
    "libloaderapi",
]

[target.'cfg(any(target_os = "linux", target_os = "freebsd", target_os = "dragonfly", target_os = "netbsd", target_os = "openbsd"))'.dependencies]
glutin_osmesa_sys = { version = "0.1.0", path = "../../glutin_sys/glutin_osmesa_sys" }
wayland-client = { version = "0.23", features = ["egl", "dlopen"] }
glutin_egl_sys = { version = "0.1.4", path = "../../glutin_sys/glutin_egl_sys" }
glutin_glx_sys = { version = "0.2.0", path = "../../glutin_sys/glutin_glx_sys" }
parking_lot = "0.10.0"
glutin_x11_sym = { version = "0.1.0", path = "../../glutin_x11_sym" }
x11-dl = "2.18.4"

[[example]]
name = "damage"
crate-type = ["cdylib"]
path = "../examples/damage.rs"

[[example]]
name = "egldevice"
crate-type = ["cdylib"]
path = "../examples/egldevice.rs"

[[example]]
name = "fullscreen"
crate-type = ["cdylib"]
path = "../examples/fullscreen.rs"

[[example]]
name = "gbm"
crate-type = ["cdylib"]
path = "../examples/gbm.rs"

[[example]]
name = "headless"
crate-type = ["cdylib"]
path = "../examples/headless.rs"

[[example]]
name = "make_current_rw"
crate-type = ["cdylib"]
path = "../examples/make_current_rw.rs"

[[example]]
name = "multiwindow"
crate-type = ["cdylib"]
path = "../examples/multiwindow.rs"

[[example]]
name = "pixmap"
crate-type = ["cdylib"]
path = "../examples/pixmap.rs"

[[example]]
name = "sharing"
crate-type = ["cdylib"]
path = "../examples/sharing.rs"

[[example]]
name = "transparent"
crate-type = ["cdylib"]
path = "../examples/transparent.rs"

[[example]]
name = "vsync"
crate-type = ["cdylib"]
path = "../examples/vsync.rs"

[[example]]
name = "window"
crate-type = ["cdylib"]
path = "../examples/window.rs"

[lib]
path = "../src/lib.rs"
```

dvc94ch: There may be a way `cargo apk -- rustc --example window -- --crate-type cdylib`. This doesn't currently create an apk but it compiles the example.

goddessfreya: Hmmm, but I was hoping to make an apk.

goddessfreya: Alright, so I applied onto winit:
```patch
diff --git a/Cargo.toml b/Cargo.toml
index a64a0583..e07eef7c 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,9 +35,9 @@ image = "0.22.4"
 simple_logger = "1.4.0"
 
 [target.'cfg(target_os = "android")'.dependencies]
-ndk = "0.1.0"
-ndk-glue = "0.1.0"
-ndk-sys = "0.1.0"
+ndk = { version = "0.1.0", path = "../android-ndk-rs/ndk" }
+ndk-glue = { version = "0.1.0", path = "../android-ndk-rs/ndk-glue" }
+ndk-sys = { version = "0.1.0", path = "../android-ndk-rs/ndk-sys" }
 
 [target.'cfg(target_os = "ios")'.dependencies]
 objc = "0.2.7"
```
Then I ran my long zsh one-liner:
```
$ rm android -rf; mkdir android; cat Cargo.toml | sed "s/\.\.\//\.\.\/\.\.\//g;s/build\.rs/\.\.\/build\.rs/" > android/Cargo.toml; ls examples/ | grep rs | awk '{sub(/\..*/,"",$1); print "\n[[example]]\nname = \"" $0 "\"\ncrate-type = [\"cdylib\"]\npath = \"../examples/" $0 ".rs\""}' >> android/Cargo.toml; printf '\n[lib]\npath = "../src/lib.rs"\n' >> android/Cargo.toml; cargo apk run --example window --manifest-path android/Cargo.toml
ERROR: unable to open 'android/target/debug/apk/examples/window-unaligned.apk' as Zip file for writing
    Updating crates.io index
   Compiling libc v0.2.66
   Compiling cfg-if v0.1.10
   Compiling autocfg v1.0.0
   Compiling log v0.4.8
   Compiling lazy_static v1.4.0
   Compiling autocfg v0.1.7
   Compiling proc-macro2 v0.4.30
   Compiling unicode-xid v0.1.0
   Compiling semver-parser v0.7.0
   Compiling syn v0.15.44
   Compiling proc-macro2 v1.0.8
   Compiling scopeguard v1.1.0
   Compiling khronos_api v3.1.0
   Compiling serde v1.0.104
   Compiling unicode-xid v0.2.0
   Compiling byteorder v1.3.4
   Compiling xml-rs v0.8.0
   Compiling syn v1.0.14
   Compiling rayon-core v1.7.0
   Compiling bitflags v1.2.1
   Compiling num-derive v0.2.5
   Compiling cc v1.0.50
   Compiling adler32 v1.0.4
   Compiling crc32fast v1.2.0
   Compiling ndk-sys v0.1.0 (/home/gentz/Documents/gfx/eco3/android-ndk-rs/ndk-sys)
   Compiling winit_types v0.1.0 (/home/gentz/Documents/gfx/eco3/winit_types)
   Compiling smallvec v1.2.0
   Compiling lzw v0.10.0
   Compiling either v1.5.3
   Compiling jni-sys v0.3.0
   Compiling color_quant v1.0.1
   Compiling android_log-sys v0.1.2
   Compiling instant v0.1.2
   Compiling scoped_threadpool v0.1.9
   Compiling takeable-option v0.5.0
   Compiling num-traits v0.2.11
   Compiling num-integer v0.1.42
   Compiling num-iter v0.1.40
   Compiling num-rational v0.2.3
   Compiling semver v0.9.0
   Compiling crossbeam-utils v0.7.0
   Compiling crossbeam-epoch v0.8.0
   Compiling lock_api v0.3.3
   Compiling inflate v0.4.5
   Compiling glutin_interface v0.1.0 (/home/gentz/Documents/gfx/eco3/glutin_interface)
   Compiling libloading v0.5.2
   Compiling gif v0.10.3
   Compiling rustc_version v0.2.3
   Compiling num_cpus v1.12.0
   Compiling atty v0.2.14
   Compiling parking_lot_core v0.7.0
   Compiling time v0.1.42
   Compiling raw-window-handle v0.3.3
   Compiling quote v0.6.13
   Compiling memoffset v0.5.3
   Compiling gl_generator v0.14.0
   Compiling quote v1.0.2
   Compiling deflate v0.7.20
   Compiling colored v1.9.2
   Compiling parking_lot v0.10.0
   Compiling crossbeam-queue v0.2.1
   Compiling png v0.15.3
   Compiling glutin_egl_sys v0.1.6 (/home/gentz/Documents/gfx/eco3/glutin_sys/glutin_egl_sys)
   Compiling glutin v0.23.0-alpha1 (/home/gentz/Documents/gfx/eco3/glutin/android)
   Compiling chrono v0.4.10
   Compiling toml v0.5.6
   Compiling simple_logger v1.5.0
   Compiling crossbeam-deque v0.7.2
   Compiling derivative v1.0.3
   Compiling proc-macro-crate v0.1.4
   Compiling rayon v1.3.0
   Compiling num_enum_derive v0.4.2
   Compiling tiff v0.3.1
   Compiling num_enum v0.4.2
   Compiling ndk v0.1.0 (/home/gentz/Documents/gfx/eco3/android-ndk-rs/ndk)
   Compiling ndk-glue v0.1.0 (/home/gentz/Documents/gfx/eco3/android-ndk-rs/ndk-glue)
   Compiling winit v0.21.0 (/home/gentz/Documents/gfx/eco3/winit)
   Compiling jpeg-decoder v0.1.18
   Compiling image v0.22.5
warning: function is never used: `main`
  --> ../examples/window.rs:10:4
   |
10 | fn main() {
   |    ^^^^
   |
   = note: `#[warn(dead_code)]` on by default

    Finished dev [unoptimized + debuginfo] target(s) in 3m 54s
ERROR: failed opening/creating 'android/target/debug/apk/examples/window-unaligned.apk' as Zip file
Error: Command '/home/gentz/Android/Sdk/build-tools/29.0.3/aapt add android/target/debug/apk/examples/window-unaligned.apk lib/armeabi-v7a/libwindow.so' had a non-zero exit code.
```
Unfortunately, cargo-apk is outputting the apk to the wrong directory:

```
$ find target/ | grep "unal"
target/debug/apk/examples/window-unaligned.apk

$ find android/target | grep "unal"
```

dvc94ch: why is that the wrong directory? since it builds fat apk's (includes all archs in the same apk) it is not placed in the target directory. if building fat apks is a good idea or not is a different question, but it's cargo-apks original behaviour, so I kept it.

dvc94ch: working on it..

dvc94ch: I think this is resolved now, let me know if it works for you

goddessfreya: I get fetched both repos, git checkout'd the newest version and ran cargo install again. Still seams to be ignoring the set manifest path. 
```
  ~/Do…ts/gfx/eco3/glutin                                                                                                                    101 ↵  0.04256s   70%  
  15:48  17.02.20  cargo apk run --example window --manifest-path android/Cargo.toml --target aarch64-linux-android
ERROR: unable to open 'android/target/debug/apk/examples/window-unaligned.apk' as Zip file for writing
   Compiling glutin v0.23.0-alpha1 (/home/gentz/Documents/gfx/eco3/glutin/android)
    Finished dev [unoptimized + debuginfo] target(s) in 11.42s
ERROR: failed opening/creating 'android/target/debug/apk/examples/window-unaligned.apk' as Zip file
Error: Command '/home/gentz/Android/Sdk/build-tools/29.0.3/aapt add android/target/debug/apk/examples/window-unaligned.apk lib/arm64-v8a/libwindow.so' had a non-zero exit code.

  ~/Do…ts/gfx/eco3/glutin                                                                                                                     1 ↵  11.74608s   70%  
  15:48  17.02.20  find target android/target | grep window-u
target/debug/apk/examples/window-unaligned.apk
```

goddessfreya: Actually, it is not making an apk at all.
```
  ~/Do…ts/gfx/eco3/glutin                                                                                                                          0.04887s   70%  
  15:49  17.02.20  rm target -rf    

  ~/Do…ts/gfx/eco3/glutin                                                                                                                          0.25960s   70%  
  15:51  17.02.20  cargo apk run --example window --manifest-path android/Cargo.toml --target aarch64-linux-android
ERROR: unable to open 'android/target/debug/apk/examples/window-unaligned.apk' as Zip file for writing
   Compiling glutin v0.23.0-alpha1 (/home/gentz/Documents/gfx/eco3/glutin/android)
    Finished dev [unoptimized + debuginfo] target(s) in 10.09s
ERROR: failed opening/creating 'android/target/debug/apk/examples/window-unaligned.apk' as Zip file
Error: Command '/home/gentz/Android/Sdk/build-tools/29.0.3/aapt add android/target/debug/apk/examples/window-unaligned.apk lib/arm64-v8a/libwindow.so' had a non-zero exit code.

  ~/Do…ts/gfx/eco3/glutin                                                                                                                     1 ↵  10.47813s   70%  
  15:51  17.02.20  find target android/target | grep window-u
find: ‘target’: No such file or directory
```

dvc94ch: I'll look into it tomorrow, need to sleep...

katyo:
It seems build command is broken.

dvc94ch:
So this approach won't work, I opened https://github.com/rust-lang/cargo/issues/7899 but I'm not sure it'll be fixed anytime soon if at all. Suggestions for alternative approaches? :)

README.md
dvc94ch: Does it still do that after I reverted the `cargo-rustc` hack?

goddessfreya: Still broken when using `--manifest-path`:
```
  ~/Do…ts/gfx/eco3/glutin                                                                                                                                       70%  
  17:14  18.02.20  find target android/target | grep window-u | xargs rm

  ~/Do…ts/gfx/eco3/glutin                                                                                                                          0.07800s   70%  
  17:14  18.02.20  find target android/target | grep window-u           

  ~/Do…ts/gfx/eco3/glutin                                                                                                                    0|1 ↵  0.03062s   70%  
  17:14  18.02.20  cargo apk run --example window --manifest-path android/Cargo.toml --target aarch64-linux-android
ERROR: unable to open 'android/target/debug/apk/examples/window-unaligned.apk' as Zip file for writing
   Compiling glutin v0.23.0-alpha1 (/home/gentz/Documents/gfx/eco3/glutin/android)
    Finished dev [unoptimized + debuginfo] target(s) in 10.30s
ERROR: failed opening/creating 'android/target/debug/apk/examples/window-unaligned.apk' as Zip file
Error: Command '/home/gentz/Android/Sdk/build-tools/29.0.3/aapt add android/target/debug/apk/examples/window-unaligned.apk lib/arm64-v8a/libwindow.so' had a non-zero exit code.

  ~/Do…ts/gfx/eco3/glutin                                                                                                                     1 ↵  10.75456s   70%  
  17:14  18.02.20  find target android/target | grep window-u

  ~/Do…ts/gfx/eco3/glutin                                                                                                                    0|1 ↵  0.03277s   70%  
  17:15  18.02.20  cargo apk run --example window --manifest-path android/Cargo.toml --target aarch64-linux-androi 
Error: Target is not supported.

  ~/Do…ts/gfx/eco3/glutin                                                                                                                      1 ↵  0.11674s   70%  
  17:15  18.02.20  cargo apk run --example window --manifest-path android/Cargo.toml                              
ERROR: unable to open 'android/target/debug/apk/examples/window-unaligned.apk' as Zip file for writing
   Compiling glutin v0.23.0-alpha1 (/home/gentz/Documents/gfx/eco3/glutin/android)
    Finished dev [unoptimized + debuginfo] target(s) in 11.00s
ERROR: failed opening/creating 'android/target/debug/apk/examples/window-unaligned.apk' as Zip file
Error: Command '/home/gentz/Android/Sdk/build-tools/29.0.3/aapt add android/target/debug/apk/examples/window-unaligned.apk lib/arm64-v8a/libwindow.so' had a non-zero exit code.

  ~/Do…ts/gfx/eco3/glutin                                                                                                                     1 ↵  11.54048s   70%  
  17:15  18.02.20  find target android/target | grep window-u                                                     

  ~/Do…ts/gfx/eco3/glutin                                                                                                                    0|1 ↵  0.02819s   70%  
  17:15  18.02.20  cd android/                                                      

  ~/Do…ts/gfx/eco3/gl…in/android                                                                                                                   0.00127s   70%  
  17:15  18.02.20  cargo apk run --example window                                   
   Compiling glutin v0.23.0-alpha1 (/home/gentz/Documents/gfx/eco3/glutin/android)
    Finished dev [unoptimized + debuginfo] target(s) in 10.11s
 'lib/arm64-v8a/libwindow.so'...
Verifying alignment of /home/gentz/Documents/gfx/eco3/glutin/android/target/debug/apk/examples/window.apk (4)...
      49 AndroidManifest.xml (OK - compressed)
     970 lib/arm64-v8a/libwindow.so (OK - compressed)
Verification succesful
Generating 2,048 bit RSA key pair and self-signed certificate (SHA256withRSA) with a validity of 10,000 days
	for: CN=Android Debug, O=Android, C=US
[Storing /home/gentz/.android/debug.keystore]

Warning:
The JKS keystore uses a proprietary format. It is recommended to migrate to PKCS12 which is an industry standard format using "keytool -importkeystore -srckeystore /home/gentz/.android/debug.keystore -destkeystore /home/gentz/.android/debug.keystore -deststoretype pkcs12".
error: no devices/emulators found
Performing Push Install
adb: error: failed to get feature set: no devices/emulators found
Error: Command '/home/gentz/Android/Sdk/platform-tools/adb install -r /home/gentz/Documents/gfx/eco3/glutin/android/target/debug/apk/examples/window.apk' had a non-zero exit code.
```

dvc94ch: The problem is that the path returned from cargo-subcommand to the `Cargo.toml` is relative to the `current_dir`, but the `aapt` command is executed from the `target/{profile}/apk` dir. Should be fixed now

ndk-glue/src/lib.rs
```diff
@@ -14,6 +14,26 @@ use std::ptr::NonNull;
 use std::sync::{RwLock, RwLockReadGuard};
 use std::thread;
 
+#[macro_export]
+macro_rules! ndk_glue {
```
katyo: May be it will the better to implement _proc_macro_ instead.
So using attribute macros the usage may looks like the follow:
```rust
#[ndk::entry]
fn main() {
// ....
}
```

dvc94ch: that may be a good improvement, but I'm getting some PR fatigue, can we leave it for a follow up?

dvc94ch:
For some reason jni still doesn't work for me. @katyo can you provide an example that worked for you?

```rust
#![cfg(target_os = "android")]

use jni_android_sys::android::content::Context;

ndk_glue::ndk_glue!(main);

fn main() {
    let vm = unsafe { jni_glue::VM::from_jni_local(&*ndk_glue::native_activity().vm()) };

    let native_lib_dir = vm.with_env(|env| {
        let context = Context::new(env).unwrap();
        //let info = context.getApplicationInfo().unwrap().unwrap();
        //let lib_dir = info.nativeLibraryDir().unwrap();
        //lib_dir.to_string().unwrap()
        ""
    });

    println!("native lib dir is {:?}", native_lib_dir);
}
```
```sh
02-19 12:50:53.938 13658 13658 F DEBUG   : pid: 13609, tid: 13654, name: Thread-2  >>> rust.example.jni <<<
02-19 12:50:53.945 13658 13658 F DEBUG   :     #08 pc 000000000001d7e0  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (jni_glue::env::Env::new_object_a::h6744845ef91a6224+76)
02-19 12:50:53.945 13658 13658 F DEBUG   :     #09 pc 000000000001ddbc  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (jni_android_sys::android::content::Context::new::hcf0b513db4bff890+152)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #10 pc 0000000000005308  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (jni::main::_$u7b$$u7b$closure$u7d$$u7d$::hd7969440a25f58b6+20)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #11 pc 0000000000004b78  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (jni_glue::vm::VM::with_env::ha2bfdd8554ee29ca+416)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #12 pc 0000000000004d7c  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (jni::main::hfe40f5c643af1c31+44)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #13 pc 0000000000010878  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (ndk_glue::init::_$u7b$$u7b$closure$u7d$$u7d$::h1f9650c5afa39c72+152)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #14 pc 000000000000c2b4  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (std::sys_common::backtrace::__rust_begin_short_backtrace::hf5d359fb8c6f708f+16)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #15 pc 000000000000b894  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (std::thread::Builder::spawn_unchecked::_$u7b$$u7b$closure$u7d$$u7d$::_$u7b$$u7b$closure$u7d$$u7d$::hea75f786b66b8340+16)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #16 pc 000000000001305c  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (_$LT$std..panic..AssertUnwindSafe$LT$F$GT$$u20$as$u20$core..ops..function..FnOnce$LT$$LP$$RP$$GT$$GT$::call_once::hd818eb5bdbde1faa+16)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #17 pc 0000000000014598  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (std::panicking::try::do_call::hbc0d5782dfb4d564+44)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #18 pc 0000000000032990  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (__rust_maybe_catch_panic+32)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #19 pc 0000000000014440  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (std::panicking::try::hbfe660383477c598+80)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #20 pc 00000000000130e4  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (std::panic::catch_unwind::hca591d4539c5427a+16)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #21 pc 000000000000b6c8  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (std::thread::Builder::spawn_unchecked::_$u7b$$u7b$closure$u7d$$u7d$::hcc10361d60d81092+212)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #22 pc 0000000000008f20  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (core::ops::function::FnOnce::call_once$u7b$$u7b$vtable.shim$u7d$$u7d$::h45da1a1034e1bb1b+16)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #23 pc 000000000002b18c  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (_$LT$alloc..boxed..Box$LT$F$GT$$u20$as$u20$core..ops..function..FnOnce$LT$A$GT$$GT$::call_once::h72164f22cf22b003+72)
02-19 12:50:53.946 13658 13658 F DEBUG   :     #24 pc 00000000000324ac  /data/app/rust.example.jni-4w_2VFpG0ATyZAaj3WU-YA==/lib/arm64/libjni.so (std::sys::unix::thread::Thread::new::thread_start::h8fae6fef275a32d6+20)
```

ndk-glue/src/lib.rs
katyo: Of course, don't worry :)

katyo:
I using jni in my oboe-rs to deal with android APIs related to audio. See https://github.com/katyo/oboe-rs/tree/master/oboe/src/java_interface, in particular _utils.rs_.
Jni works good.
There is the [demo](https://github.com/katyo/oboe-rs/blob/master/oboe-demo/src/main.rs).

katyo:
> So this approach won't work, I opened [rust-lang/cargo#7899](https://github.com/rust-lang/cargo/issues/7899) but I'm not sure it'll be fixed anytime soon if at all. Suggestions for alternative approaches? :)
> 
> Tried to fix it in [rust-lang/cargo#7900](https://github.com/rust-lang/cargo/pull/7900), but don't understand the problem well enough.

If I understand right the problem in changing crate type to _cdylib_ for android builds only, i.e when cargo apk is used to build app.
Because the PR for cargo stalled for now (and it seems for long time), we forced try another workarounds.

katyo:
The [rust-android-gradle](https://github.com/mozilla/rust-android-gradle) plugin from mozilla team overrides cargo linker by wrapper script to add specific options to it when linking.
https://github.com/mozilla/rust-android-gradle/blob/master/plugin/src/main/kotlin/com/nishtahir/CargoBuildTask.kt#L180
https://github.com/mozilla/rust-android-gradle/blob/master/plugin/src/main/kotlin/com/nishtahir/CargoBuildTask.kt#L193

dvc94ch:
I think it still relies on cargo to create the cdylib? https://github.com/mozilla/rust-android-gradle/blob/master/samples/rust/Cargo.toml#L12 although I'm a little confused why the manifest contains dylib instead of cdylib, as I thought dylib is for macosx.

katyo:
Actually them wrapper for linker overrides cargo's behavior.
You free to set any _cargo-type_ this setting will be ignored because in current implementation wrapper overrides it.

dvc94ch:
The only interesting thing the linker wrapper seems to do is add this `"-Wl,-soname,lib${cargoExtension.libname!!}.so"`. I'll try adding those to RUSTFLAGS and see if I can get it to work

katyo:
For some reasons that cannot be done using RUSTFLAGS, otherwise no wrapper would be needed.

dvc94ch:
So what's the easiest way to test this?

katyo:
I think as experiment we may try add wrapper for linker in NDK.

dvc94ch:
@UndeadLeech let's start with this one! :)

Osspial:
@dvc94ch Hi! Is there anything in particular you'd like me to help out with? I'm not seeing anything in particular in the past few messages and this is a pretty long thread.

dvc94ch:
@osspial looks like this is ready to merge

Osspial:
Oh, neat. Will do.

@dvc94ch Do you want collaborator access on this repository? It looks like you've been doing good work here, and that would make it easier for you to continue improving things.

dvc94ch:
That would cool. Can you publish to crates.io?

Osspial:
I'll get that done as soon as I figure out how to build this properly on my Windows machine.

Osspial:
@dvc94ch I managed to run into a standard library bug that's preventing me from building this on Windows, so it'll be a bit before I can personally upload this to crates.io. Could you pull the latest changes I made to this repository, publish it yourself, then add `github:rust-windowing:publishers` to the owners list for each of the crates?

Osspial:
Alright, I managed to work around the bug in `cc-rs`. The crates have been published!

dvc94ch:
Great! Thanks. I'll backport the winit stuff, just haven't had time yet.

PR #18, opened 2020-05-13
ndk-build: Point CXX_<triple> to <triple>-clang++ from the ndk

MarijnS95:
Cross-compiling with build.rs using cc::Build and .use_cpp(true) (for
example) invokes <triple>-clang++ which is usually not on PATH or known
to point to the correct NDK. Set the CXX env var for the requested
triple to ensure the right C++ compiler is available and used.

---

Path logic remains in `fn clang`to do the same `exists()` check there. Let me know if this way of formatting and changing the name is the right way around.


dvc94ch:
I'll make a release tomorrow, I'm not at a computer right now

PR #20, opened 2020-05-14
ndk-build: Include libc++_shared in the APK

### LINKED ISSUE #19: Linking against C++ STL and using it at runtime
Native C++ code (in external dependencies) that utilize the STL have a hard time linking and running on-device. Dependencies usually set `cpp_link_stdlib` (if using the `cc` crate) to something sensible or leave it at the default `stdc++` which only works for GNU but not Clang, requiring `c++` instead. Only a single dependency needs to set it to `c++` or `c++_shared` for the right library to be linked in and succeed the build. Leaving `stdc++` in there is fine, this library [_is_ available on-device](https://developer.android.com/ndk/guides/cpp-support#system), providing only `new` and `delete`.
I assume the right course of action is to add an `android` case with `c++` to the [defaults in cc](https://docs.rs/cc/1.0.52/src/cc/lib.rs.html#2280)? (Or in the exceptional case, do that for all of clang? The documentation mentions this case yet does not account for it.)

As mentioned [here](https://developer.android.com/ndk/guides/cpp-support#libc) `libc++_shared` is not available on Android devices - it needs to be bundled with the APK. Builds link against `c++_static` by default and have no runtime dependencies, but aforementioned native projects would force linking against `c++`.

A simple solution adds the path to `libc++_shared.so` from `android_search_paths` to `artifacts` in readelf.rs, if it is found in the needed libs ([like this](https://github.com/MarijnS95/android-ndk-rs/commit/include-libc++_shared) - forgive the awful code and duplication from a Rust novice). A more sophisticated implementation would provide the option to choose between none, static or runtime STL (including the file for the latter) and clean up linker flags to accommodate this choice.

Any suggestions going forward? I'm willing to take this onwards but am not sure how to proceed.

MarijnS95:
Closes #19

libc++_shared is not available on-device, and has to be provided in the
APK when used.
See the NDK guide:
https://developer.android.com/ndk/guides/cpp-support#libc

### DIFF
--- a/ndk-build/src/readelf.rs
+++ b/ndk-build/src/readelf.rs
@@ -17,9 +17,9 @@ impl<'a> UnalignedApk<'a> {
         let min_sdk_version = self.config().manifest.min_sdk_version;
         let readelf_path = ndk.toolchain_bin("readelf", target)?;
 
-        let android_search_paths = vec![
-            ndk.sysroot_lib_dir(target)?,
-            ndk.sysroot_platform_lib_dir(target, min_sdk_version)?,
+        let android_search_paths = [
+            &*ndk.sysroot_lib_dir(target)?,
+            &*ndk.sysroot_platform_lib_dir(target, min_sdk_version)?,
         ];
 
         let mut provided = HashSet::new();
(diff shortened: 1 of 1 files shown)


dvc94ch:
Thanks!

PR #23, opened 2020-05-15
support <meta-data> tags in manifest

iamtommo:
use case: `<meta-data android:name="com.samsung.android.vr.application.mode" android:value="vr_only"/>`

naming of application_metadatas is rather unfortunate, suggestions?


dvc94ch:
Can you run rustfmt?

Issue #22, opened 2020-05-15
Adding support for more NDK libraries


JasperDeSutter:
I'm wondering what the best approach is to adding support to more Android NDK libraries.
Currently ndk-sys only links to 'libandroid', which is the only lib needed for basic windowing support.
Some other libraries that could be useful:
* libaaudio: can be used to add support to [cpal](https://github.com/RustAudio/cpal) for example to add android support to Rust game engines.
* libjnigraphics: writing into Android bitmaps can be useful for image processing libraries.
* libcamera2ndk: Android camera support would be very nice to have in the Rust ecosystem.
* libmediandk: needed for reading images from camera, media codecs could be useful too.

Could these be included in this repo, perhaps as different crates to still have the modularity? It would lead to this repo being something more in the line of winapi, and less just for native activity glue support.
The other option is maintaining these other libraries separately as an 'ndk-libs' library or individually as needed. This might lead to a very fragmented approach towards configuring API level (see below) and documentation.
What is your take on this @dvc94ch?

Other questions arise such as how to deal with Android API levels? This is needed because native API's only become available at certain android versions.
Cargo features map pretty well to this IMO. It would be nice that all these libs would use the same system for that.

Some ideal usage of these libs would look like this:

```toml
ndk = { version = "0.1.0", features = ["api-27", "camera", "media"] }
```
or
```toml
ndk = { version = "0.1.0", features = ["api-27"] }
ndk-camera = { version = "0.1.0", features = ["api-27"] }
ndk-media = { version = "0.1.0", features = ["api-27"] }
```

Maybe cargo-apk can also check these levels to match 'min_sdk_version' for example.

dvc94ch:
Maybe we should follow what jni-bindgen does, so that developers don't have to remember which library uses which feature names. [0]

If I understand correctly those libraries you mentioned are just thin wrappers around the jni api's. I don't understand enough about the jni, but it seems it leads to converting java::String to CString to String, so maybe the better approach would be to use the jni directly.

- [0] https://github.com/MaulingMonkey/jni-bindgen/tree/master/jni-android-sys#build-features

JasperDeSutter:
Using the same system makes sense, so then it just boils down to "api-level-X" cargo features.

These other libraries are not based on JNI though, except for 'jnigraphics' which just adds some extra functionality to the JNI exposed Bitmap type.
[...]

PR #24, opened 2020-05-15
Add mediandk image reader

JasperDeSutter:
Adds bindings to an extra Android NDK library for reading images from other API's like camera (#22).
This should already give a good idea of how it would look with more libraries like this added.

See `ndk/Cargo.toml` for the feature-gating approach. I tried adding both API levels and individual features this way. API levels would always depend on the lower API levels, and some features are also dependent on a specific API lever (hardware_buffer as an example).

I think regenerating the bindings with rust-bindgen on the latest NDK headersr turned `usize` into a generated `size_t`, hence some unrelated changes in `ndk/src/event.rs`.


ndk/Cargo.toml
```diff
@@ -13,6 +13,10 @@ repository = "https://github.com/rust-windowing/android-ndk-rs"
 
 [features]
 rustdoc = ["ffi/rustdoc", "jni", "jni-glue"]
```
dvc94ch, line 15: can you add the new features to rustdoc?

ndk/src/media/image_reader.rs
```diff
@@ -0,0 +1,309 @@
...
+
+    pub fn get_crop_rect(&self) -> Result<CropRect> {
+        construct(|res| unsafe { ffi::AImage_getCropRect(self.as_ptr(), res) })
+    }
+
+    pub fn get_width(&self) -> Result<i32> {
+        construct(|res| unsafe { ffi::AImage_getWidth(self.as_ptr(), res) })
+    }
+
+    pub fn get_height(&self) -> Result<i32> {
```
dvc94ch, line 272: it's uncommon to prefix getters with `get_`. on the other hand staying close to the original naming might also be good. I have mixed feelings on this.

PR #25, opened 2020-05-16
Add bitmap and hardware buffer

JasperDeSutter:
Continues the work for #22 

* adds Android Bitmap module. There are some additions to the NDK Bitmap in API-level-30, but I don't think it makes sense to add all of it before the full release, as things might still change.
* adds HardwareBuffer implementation. I'm not sure all the fence stuff is right, as I have no idea how to test it yet. I'll dive deeper when adding the camera API, that would also be the perfect time to add an example app using all of these modules.

### DIFF
--- a/ndk/src/hardware_buffer.rs
+++ b/ndk/src/hardware_buffer.rs
@@ -1,6 +1,12 @@
 #![cfg(feature = "hardware_buffer")]
-use std::ptr::NonNull;
+use jni_sys::{jobject, JNIEnv};
+use num_enum::{IntoPrimitive, TryFromPrimitive};
+use std::{
+    convert::TryInto, mem::MaybeUninit, ops::Deref, os::raw::c_void, os::unix::io::RawFd,
+    ptr::NonNull,
+};
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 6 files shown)


ndk/src/hardware_buffer.rs
```diff
@@ -73,12 +79,317 @@ impl HardwareBufferUsage {
...
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct HardwareBufferDesc {
+    width: u32,
+    height: u32,
+    layers: u32,
+    format: HardwareBufferFormat,
+    usage: HardwareBufferUsage,
+    stride: u32,
```
MarijnS95, line 348: @JasperDeSutter While writing/importing documentation for `hardware_buffer` and referring to one of these fields I found "by accident" that they are private. Is that intended?

JasperDeSutter: Heh, yeah that would make a lot more sense! I guess I didn't need it for my use case where the information is passed through jni, and simply forgot to add the pubs (no beer, I swear 😄)

MarijnS95: No worries, thanks for confirming!

PR #27, opened 2020-05-22
Add AAudio interface

JasperDeSutter:
Adds bindings for Android NDK's AAudio interface (part of #22).
Also fixes an issue with previous PR where `*mut JNIEnv` pointers were getting dereferenced which doesn't as I expected.

### DIFF
--- a/ndk-sys/Cargo.toml
+++ b/ndk-sys/Cargo.toml
@@ -14,6 +14,7 @@ repository = "https://github.com/rust-windowing/android-ndk-rs"
 [features]
 rustdoc = []
 test = []
+aaudio = []
 bitmap = []
 media = []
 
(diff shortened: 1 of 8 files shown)


dvc94ch:
We need tests and examples, so that we can run miri to detect ub one day. But for the time being they still don't support ffi calls.

JasperDeSutter:
@dvc94ch Makes sense, I'm wondering how the testing setup would look like. I suppose tests would need to run on an Android emulator to have access to the native binaries. Do you have anything in mind already?

dvc94ch:
Not really, I just know I had issues with segfaults when using jni-bindgen and I assume it's because our safe wrappers invoke ub somehow. There are some links in #28, but that's as far as my research went.

PR #37, opened 2020-06-04
Added `ndk_glue::main` attribute macro

katyo:
Because a more convenient way of metaprogramming in rust related to proc_macro, I tried to replace `ndk_glue` macro with attribute procedural macro (`main`) which applied directly to main function.

* Enabling backtraces using env-var isn't default behavior now and can be done using `backtrace` property.
* Added `logger` property which configures _android_logger_ before entering to
main function.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 members = [
     "ndk",
+    "ndk-macro",
     "ndk-build",
     "ndk-examples",
     "ndk-glue",
--- a/README.md
+++ b/README.md
@@ -13,11 +13,11 @@
 crate-type = ["lib", "cdylib"]
 ```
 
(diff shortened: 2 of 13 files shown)


katyo:
CI failed on nightly:
```
error: 'cargo-fmt' is not installed for the toolchain 'nightly-x86_64-unknown-linux-gnu'
```

dvc94ch:
It's a lot of code for a simple macro... But it's very clean and most of the code are tests, so I guess it's fine. Thanks!

katyo:
> It's a lot of code for a simple macro... But it's very clean and most of the code are tests, so I guess it's fine. Thanks!

Procedural macrosses usually requires many more code comparing to ordinary macro rules. But it's really powerful, clean to understand, testable and allows get hard things much simpler.

In our case it may looks overhead, but currently this is one only way to implement attribute macros in Rust.

Issue #36, opened 2020-06-04
Error: Android SDK has no platforms installed.


Araminos:
'cargo apt run' produces error. Please see the [issue](https://github.com/rust-windowing/android-ndk-rs/issues/35).

My related to android environmental variables:
```bash
# Deprecated (in Android Studio), use ANDROID_SDK_ROOT instead. 
export ANDROID_HOME="$HOME/bin/android-sdk"

# Installation directory of Android SDK package. Example: C:\AndroidSDK or ~/android-sdk/
export ANDROID_SDK_ROOT="$HOME/bin/android-sdk"

# Installation directory of Android NDK package. Example: C:\AndroidNDK or ~/android-ndk/
export ANDROID_NDK_ROOT="$HOME/bin/android-sdk/ndk"

# just because "cargo apk run" require
export NDK_HOME="$HOME/bin/android-sdk/ndk"
```

dvc94ch:
Do you want to submit a PR so that we can review it?

adals:
```
echo $ANDROID_SDK_ROOT                                        
/Users/xxx/Library/Android/sdk
echo $NDK_HOME        
/Users/xxx/Library/Android/sdk/ndk
 echo $ANDROID_NDK_ROOT
/Users/xxx/Library/Android/sdk/ndk
 echo $ANDROID_HOME                                        
/Users/xxx/Library/Android/sdk
 cargo apk run                                             
Error: Android SDK has no platforms installed.
```

what is the problem?

I have Android studio and run the emulator and everything is ok but cargo apk run will give an error

dvc94ch:
https://github.com/rust-windowing/android-ndk-rs/issues/26#issuecomment-633065487

adals:
@dvc94ch it works thanks

Araminos:
> Do you want to submit a PR so that we can review it?

I am not very experienced rust programmer. I am alien from java world. I can try if you would support me by answering some questions by email o chat channel. Is it possible?

Araminos:
I built cargo-apk from current branch and reinstalled SDK Platforms ans SDK Tools from Android Studio. After I started emulator manually "cargo-apk apk run" installed an application to emulator. I think the issue is resolved. I don't see "Hello world". Where is output of println?

dvc94ch:
Seems resolved

PR #40, opened 2020-06-10
Issue #36

Araminos:


Araminos:
> This is incorrect

I found platforms in this path. But after you reply I found platforms in ndk too. How to manage platforms in ndk. I can manage platforms in ndk by android studio.

dvc94ch:
Exactly, we're checking that there's an intersection. You changed it to be always true for all platforms in the sdk

PR #45, opened 2020-07-05
Specify assets directory via fully-qualified path

msiglreith:
The working directory is for the `aapt` packing command is set to the target directory while the assets directory path from the manifest is just forwarded. This results in the android tool to not find the directory.

Tested locally on Windows.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,12 +17,18 @@ jobs:
         ]
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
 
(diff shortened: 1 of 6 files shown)


msiglreith:
note: resources might be affected as well but didn't have use for them nor a test case

dvc94ch:
It seems like the res path should be handled the same way

dvc94ch:
Can you also fix ci? Someone has to do it, you can disable nightly

msiglreith:
All green now! ✔️

dvc94ch:
Great! Thanks

Issue #48, opened 2020-07-09
Cannot build example


stefan2904:
I am new to the Rust world, and I just tried to build the example.

Following the commands in the README:

```
cd ndk-examples
cargo apk build
```
(`cargo apk run` has the same problem.)

Unfortunately it does not work:

> % cargo apk build
>    Compiling proc-macro2 v1.0.18
>    Compiling unicode-xid v0.2.1
>    Compiling syn v1.0.33
>    Compiling serde v1.0.114
>    Compiling log v0.4.8
>    Compiling libc v0.2.72
>    Compiling cfg-if v0.1.10
>    Compiling ndk-sys v0.1.0 (/home/smore/projects/android-ndk-rs/ndk-sys)
>    Compiling jni-sys v0.3.0
>    Compiling lazy_static v1.4.0
>    Compiling android_log-sys v0.1.2
>    Compiling quote v1.0.7
>    Compiling toml v0.5.6
>    Compiling proc-macro-crate v0.1.4
>    Compiling thiserror-impl v1.0.20
>    Compiling derivative v2.1.1
>    Compiling num_enum_derive v0.4.3
>    Compiling ndk-macro v0.1.0 (/home/smore/projects/android-ndk-rs/ndk-macro)
>    Compiling thiserror v1.0.20
>    Compiling num_enum v0.4.3
>    Compiling ndk v0.1.0 (/home/smore/projects/android-ndk-rs/ndk)
>    Compiling ndk-glue v0.1.0 (/home/smore/projects/android-ndk-rs/ndk-glue)
>    Compiling ndk-examples v0.1.0 (/home/smore/projects/android-ndk-rs/ndk-examples)
>     Finished dev [unoptimized + debuginfo] target(s) in 14.35s
>
> Error: Path "/home/smore/projects/android-ndk-rs/target/aarch64-linux-android/debug/libndk_examples.so" doesn't exist.
> cargo apk build  54.83s user 2.89s system 396% cpu 14.554 total

Is there a step missing?

Thanks!

---

I am running `cargo 1.44.1 (88ba85757 2020-06-11)` on the latest Ubuntu.

adals:
+1 same here on MacOs

$ cargo -V
cargo 1.46.0-nightly (fede83ccf 2020-07-02)
$ rustc -V 
rustc 1.46.0-nightly (8ac1525e0 2020-07-07)

msiglreith:
`cargo apk build --example hello_world` should be working.
[...]

PR #49, opened 2020-07-11
Improve repository documentation

msiglreith:
Tried to improve the general documentation of the repository and crates to make it easier to discover features and getting started.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,23 @@
 # Rust on Android
 
- - Raw FFI bindings to the NDK ![ndk-sys-docs][ndk-sys-badge]
- - Safe abstraction of the bindings ![ndk-docs][ndk-badge]
- - Startup code ![ndk-glue-docs][ndk-glue-badge]
- - Everything for building apk's ![ndk-build-docs][ndk-build-badge]
- - Build tool ![cargo-apk-docs][cargo-apk-badge]
+[![Rust](https://github.com/rust-windowing/android-ndk-rs/workflows/Rust/badge.svg)](https://github.com/rust-windowing/android-ndk-rs/actions) ![MIT license](https://img.shields.io/badge/License-MIT-green.svg) ![APACHE2 license](https://img.shields.io/badge/License-APACHE2-green.svg)
+
+Libraries and tools for Rust programming on Android targets:
+
+Name | Description | Badges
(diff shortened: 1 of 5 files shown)


charleschege:
I have tried adding the `[[package.metadata.android.permission]] ` from this pull request and it is returning an error

```
AndroidManifest.xml:13: error: Error parsing XML: not well-formed (invalid token)

Error: Command '/usr/local/android/sdk/build-tools/28.0.3/aapt package -f -F /home/user/projects/orbandroid/target/debug/apk/orbandroid-unaligned.apk -M AndroidManifest.xml -I /usr/local/android/sdk/platforms/android-29/android.jar' had a non-zero exit code
```

dvc94ch:
Thanks @msiglreith

MarijnS95:
@msiglreith Was the `ndk-examples` removal (as previously discussed) removed from https://github.com/rust-windowing/android-ndk-rs/commit/0cc97ff2b67cb70860edd63bf5c859b33ffa3f4b in favour of submitting that in a separate PR?

https://github.com/rust-windowing/android-ndk-rs/issues/48 is not properly resolved yet without that. Would you mind if I re-apply and PR the change?

msiglreith:
@MarijnS95 Feel free to file a PR! I revert it as it didn't pan out as expected, but I don't recall the exact issues I had with removing `ndk-examples`. I think part of it was separating the manifest from the from cargo library (e.g for asset paths, crate-type, etc.)

MarijnS95:
@msiglreith That'd be for a more elaborate example I guess? In that case we're better off making `ndk-examples` _the_ example instead of the empty lib it is now. That saves polluting `ndk-glue` as [previously proposed](https://github.com/rust-windowing/android-ndk-rs/issues/48#issuecomment-656931548).

We should probably add a very simple looper to prevent the example from locking up and crashing, too.

dvc94ch:
I don't think this is necessary, just because people are too lazy to read the documentation or notice that the crate follows standard cargo conventions and the tool standard cargo cli options

MarijnS95:
@dvc94ch to be fair I have - in my limited Rust career - not seen a single example stuffed in an otherwise empty library project yet. Either the project _is_ the example, or the example is integrated with one of the top-level packages.

PR #55, opened 2020-08-01
Add native trace API

JasperDeSutter:
To test tracing functionality, see: https://developer.android.com/topic/performance/tracing/on-device


ndk/src/trace.rs
```diff
@@ -0,0 +1,94 @@
...
+#![cfg(feature = "trace")]
+use std::ffi::{CString, NulError};
+
+pub fn is_trace_enabled() -> bool {
+    unsafe { ffi::ATrace_isEnabled() }
+}
+
+#[derive(Debug)]
+pub struct Section {
+    _private: *mut (),
```
msiglreith, line 13: minor: a comment should be added why this field is required. I would also suggest to rather use `PhantomMarker` if it's soley to enforce !Send+!Sync

JasperDeSutter: Indeed, didn't have to use a pointer here. Thanks for the suggestion!

ndk/src/trace.rs
```diff
@@ -0,0 +1,92 @@
...
+    _pd: PhantomData<&'static ()>,
+}
+
+#[cfg(feature = "api-level-29")]
+impl AsyncSection {
+    pub fn new(name: &str, cookie: Cookie) -> Result<Self, NulError> {
+        let section_name = CString::new(name)?;
+        unsafe { ffi::ATrace_beginAsyncSection(section_name.as_ptr(), cookie.0) };
+
+        Ok(AsyncSection {
```
francesca64, line 57:
```suggestion
        Ok(Self {
```

ndk/src/trace.rs
```diff
@@ -0,0 +1,92 @@
...
+    // Section is !Sync and !Send
+    _pd: PhantomData<*mut ()>,
+}
+
+impl Section {
+    pub fn new(name: &str) -> Result<Self, NulError> {
+        let section_name = CString::new(name)?;
+        unsafe { ffi::ATrace_beginSection(section_name.as_ptr()) };
+
+        Ok(Section { _pd: PhantomData })
```
francesca64, line 23:
```suggestion
        Ok(Self { _pd: PhantomData })
```

PR #57, opened 2020-08-07
Drop failure for anyhow, derive Error impls using thiserror

### LINKED ISSUE #16: drop failure in favor of thiserror

JasperDeSutter:
fixes #16

before (cargo tree --invert failure):
```
failure v0.1.8
├── exitfailure v0.5.1
│   └── cargo-apk v0.5.1 (/home/jadsu/git/android-ndk-rs/cargo-apk)
└── which v3.1.1
    └── ndk-build v0.1.1 (/home/jadsu/git/android-ndk-rs/ndk-build)
        └── cargo-apk v0.5.1 (/home/jadsu/git/android-ndk-rs/cargo-apk)
```
        
after:
`error: package ID specification 'failure' matched no packages`

* Replaced some Error/Display impls with `thiserror`
* Fixed small clippy nits, remaining issues are now mainly missing native function safety docs.

### DIFF
--- a/ndk/src/aaudio.rs
+++ b/ndk/src/aaudio.rs
@@ -190,7 +190,7 @@ impl AAudioErrorResult {
 pub enum AAudioError {
     #[error("error AAudio result ({0:?})")]
     ErrorResult(AAudioErrorResult),
-    #[error("unknown AAudio result value ({0})")]
+    #[error("unknown AAudio error result ({0})")]
     UnknownResult(i32),
     #[error("unsupported AAudio result value received ({0})")]
     UnsupportedValue(i32),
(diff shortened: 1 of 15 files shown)


ndk/src/aaudio.rs
```diff
@@ -229,7 +229,7 @@ impl AAudioError {
 
 pub type Result<T, E = AAudioError> = std::result::Result<T, E>;
 
-fn construct<T>(with_ptr: impl FnOnce(*mut T) -> ffi::camera_status_t) -> Result<T> {
+fn construct<T>(with_ptr: impl FnOnce(*mut T) -> ffi::aaudio_result_t) -> Result<T> {
```
francesca64, line 232: What's this about?

JasperDeSutter: It doesn't do much since both are just typedefs for `c_int`/`i32`, but the `aaudio` one should be used in this case because it handles aaudio functions.

francesca64: Gotcha, thanks for catching this!

PR #63, opened 2020-08-17
Add screenOrientation configuration via metadata

genuinebyte:
Allows changing the orientation of the screen by setting `android:screenOrientation` in the manifest. It might be worth adding to the cargo-apk readme that the default value of "unspecified" doesn't do any rotation for you at all, but "sensorLandscape" and "sensorPortrait" allow rotation from their normal orientations and their reverse orientations.

### DIFF
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -44,6 +44,12 @@ fullscreen = false
 # Defaults to [3, 1]
 opengles_version = [3, 0]
 
+# Sets the applications screenOrientation.
+# See https://developer.android.com/guide/topics/manifest/activity-element
+# and look for `a`ndroid:screenOrientation` for possible values
+# Defaults to "unspecified"
+orientation = "sensorLandscape"
+
 # Adds a uses-feature element to the manifest
 # Supported keys: name, required
 # See https://developer.android.com/guide/topics/manifest/uses-feature-element
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -46,8 +46,9 @@ opengles_version = [3, 0]
 
 # Sets the applications screenOrientation.
 # See https://developer.android.com/guide/topics/manifest/activity-element
-# and look for `a`ndroid:screenOrientation` for possible values
-# Defaults to "unspecified"
+# and look for `android:screenOrientation` for possible values
+# Defaults to "unspecified" which makes the system pick an orientation and
+# doesn't give you help with rotation.
 orientation = "sensorLandscape"
 
 # Adds a uses-feature element to the manifest
--- a/ndk-build/src/manifest.rs
+++ b/ndk-build/src/manifest.rs
@@ -18,6 +18,7 @@ pub struct Manifest {
     pub intent_filters: Vec<IntentFilter>,
     pub icon: Option<String>,
     pub fullscreen: bool,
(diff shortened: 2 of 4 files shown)


ndk-build/src/manifest.rs
```diff
@@ -45,6 +46,12 @@ impl Manifest {
             ""
         };
 
+        let orientation = if let Some(orient) = self.orientation.as_ref() {
```
francesca64, line 49: I think this would look cuter:
```rust
let orientation = self.orientation.as_ref().unwrap_or("unspecified");
```

cargo-apk/README.md
```diff
@@ -44,6 +44,12 @@ fullscreen = false
 # Defaults to [3, 1]
 opengles_version = [3, 0]
 
+# Sets the applications screenOrientation.
+# See https://developer.android.com/guide/topics/manifest/activity-element
+# and look for `a`ndroid:screenOrientation` for possible values
```
francesca64, line 49:
```suggestion
# and look for `android:screenOrientation` for possible values
```

francesca64:
>  It might be worth adding to the cargo-apk readme that the default value of "unspecified" doesn't do any rotation for you at all

That'd definitely be great to document!

dvc94ch:
Don't you get a resized event when rotating?

genuinebyte:
It seems that no, you don't get a resized event. When I tested it there was one when the app first started, then none at all. I'm using forks of winit and glutin so it might be possible that the event gets lost somewhere, but I get one when it launches so probably not?

genuinebyte:
I got an email that CI failed and then I realized I probably should've tried to recompile after I made the change (stupid mistake). In my last commit I reverted how we get the orientation from the `Option<String>`. The details are in the message of the commit, but essentially the types wouldn't match. Maybe there's a way to do it that I'm unaware of, but right now the `if let` seems cleaner than making another variable.

dvc94ch:
Can this be made optional? The relevant part of the manifest is here:
```
android:configChanges="orientation|keyboardHidden|screenSize">
```
Which means that the config changes event will be emitted on orientation changes.

francesca64:
> Maybe there's a way to do it that I'm unaware of

Sorry about that! `self.orientation.as_deref().unwrap_or("unspecified")` would work, but I'm fine with just keeping `if let`.

genuinebyte:
> Can this be made optional? The relevant part of the manifest is here:
> 
> ```
> android:configChanges="orientation|keyboardHidden|screenSize">
> ```
> 
> Which means that the config changes event will be emitted on orientation changes.

Make this PR optional, or "orientation" in configChanges? Do you think there should there be a configurability to all the values in configChanges?

> > Maybe there's a way to do it that I'm unaware of
> 
> Sorry about that! `self.orientation.as_deref().unwrap_or("unspecified")` would work, but I'm fine with just keeping `if let`.

Hey, this works! I wasn't aware of `as_deref` so I read the docs and now I feel a little smarter, thanks! I'm going to revert to when I fist tried to use `unwrap_or` with `as_ref` and force push. I've heard force pushing is taboo but I'm fairly confident I can't break anything here and it'll make the git history cleaner.

francesca64:
Whoops, forgot to ever merge this!

PR #69, opened 2020-09-09
Allow change apk label with apk_label in metadata

genuinebyte:
Allows you to change the label of your APK using the metadata in your Cargo.toml. Hope this is useful!


msiglreith (approved):
thanks!
tested and works perfectly

PR #70, opened 2020-09-09
ci: only format on stable channel

msiglreith:
`rustfmt` is not included always in the nightly releases, resulting in spurious ci build fails


francesca64 (approved):
Thanks! This is great.

PR #14, opened 2020-04-22
Improvement pass on github actions

Osspial:
Pulls in some of the CI improvements from Winit. Adds auto-publish, and runs tests on all first-class 64-bit host platforms.

This is a WIP pull request, and I'll be updating it as I get responses from GitHub Actions.

PR #71, opened 2020-09-10
CI publishing workflow

msiglreith:
Replacement for https://github.com/rust-windowing/android-ndk-rs/pull/14
Automatically publish crates once the versions get bumped on the master

Reverts the old 0.1.1 version bumps which got yanked due to breaking compatibility
I'm not 100% this works atm, but I guess the secrets are setup already

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -8,6 +8,7 @@ on:
 jobs:
   Publish:
     runs-on: ubuntu-latest
+    continue-on-error: true
     strategy:
       max-parallel: 1 # ensure crate order
       fail-fast: false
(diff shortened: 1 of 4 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,34 @@
...
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+        target: ${{ matrix.rust-target }}
+        override: true
+    - name: Publish ${{ matrix.crate }}
+      uses: actions-rs/cargo@v1
+      with:
+        command: publish
+        args: --manifest-path ${{ matrix.crate }}/Cargo.toml --dry-run --target ${{ matrix.rust-target }} --token ${{ secrets.cratesio_token }}
```
francesca64, line 34: Won't `--dry-run` keep this from actually publishing?

Also, I've personally been using `continue-on-error: true` to keep CI from turning red if Cargo.toml is modified but the version isn't bumped... but that's not a perfect solution, since it could obscure publish failures.

francesca64 (commented):
Thanks so much! I've added the secret.

.github/workflows/publish.yml
msiglreith: oops! fixed

francesca64 (approved):
Thanks again!

PR #79, opened 2020-10-11
Fix res folder resolve

enfipy:
Currently, there's an error if you try to specify the `res` folder in the project that running not from root Cargo.toml


msiglreith:
Thanks, could you add an entry to the changelog?

enfipy:
Sure!

msiglreith (approved):
Great, thanks again!

PR #83, opened 2020-10-14
ci: cache ndk download

msiglreith:
curl download of the ndk seems to fail pretty frequently today at least causing the builds to fail most of the time.
This caches the ndk download which shoves off ~45sec from each build (which so far took 4-5min each)


francesca64 (approved):
Looks great!

PR #84, opened 2020-10-22
Added support for activity metadata

JoNil:


msiglreith (commented):
Thanks!
Could you add a changelog entry and a short example entry [here](https://github.com/rust-windowing/android-ndk-rs/tree/master/cargo-apk#manifest) please?

JoNil:
Done! I added an example for application metadata as well since i was using that too. The example is what I'm actually using it for. Maybe not the most widely applicable 😅😋

msiglreith (approved):
Great!

Issue #53, opened 2020-07-29
ndk_glue::main assumes ndk-glue is in Cargo.toml


francesca64:
My use case involves exporting ndk-glue from another crate, so `proc-macro-crate` being used to get the crate name makes this macro unusable for me.

It would be nice to have an option to manually specify the crate name, though honestly I'd be content just to have the `ndk_glue!` macro still be available as an alternative...

dvc94ch:
I like the ndk_glue! macro, it's simple and concise. This does seem like a case of over engineering to me, but since @katyo 's pr was clean and well tested, I didn't feel comfortable rejecting it.

dvc94ch:
@francesca64 can you come up with a solution that makes everyone happy?

francesca64:
@dvc94ch I believe so. I recall you [mentioned](https://github.com/rust-windowing/glutin/pull/1313#issuecomment-657211507) that you no longer depend on Rust's Android stack? I work for a company that intends to use this library in production, so I can definitely take a more active role here if you'd like.

@katyo can you make a stronger case for the merits of your proc macro? It replaces a ~20 line declarative macro with ~750 new lines of code, which is nothing to sneeze at. While it does look rather clean, and I commend you for adding extensive tests, having more code nonetheless represents a maintenance burden.

I love proc macros, but their complexity can be an obstacle for new contributors (or just for people reading the code), and `syn` is becoming increasingly unpopular with the "fewer dependencies for faster build times" crowd. However, I can't argue against the power proc macros offer, so I'm just asking you to make the case that said power is warranted. It looks like you've made logging and backtraces configurable, which I imagine would be difficult to implement as elegantly using declarative macros. But even ignoring that, are you confident that having an attribute is substantially more convenient for users?

dvc94ch:
@francesca64 that would be great. I'm working on completely different stuff at the moment so if you want to take over that's great.

katyo:
@francesca64 Attribute macros is a proper way to apply transformations to Rust code. Almost all crates use something like that (see `async_std::main`, `tokio::main`).
I planned to re-implement `ndk-macro` using __darling__. In spite that it little bit breaks current API (ex. macro will accepts `backtrace = "full"` instead of `backtrace(full)`) but source code will looks much shorter and simpler.

katyo:
I completely re-implemented `main` using __darling__ in pr #54.

katyo:
[...]

PR #88, opened 2020-11-03
ndk-macro: Reexport android_logger from ndk-glue and use it in macros

MarijnS95:
Hi all,

Upfront apologies for my bluntness in this "RFC": I'm usually rather disappointed after running into `undeclared type or module` errors when a macro generates code in my (programs) scope with a hard dependency on some crate or symbol there.

This PR is a _suggested_ "fix" for addressing the missing `android_logger` logger when enabling the `logging` feature, without importing that crate into the target package.  It is not how I envision a final solution so I'm opening this for feedback, read on for more questions.  Note that tests were intentionally not updated until a proper discussion how to best implement this.

I haven't really understood the purpose of providing the path to `android_logger` in the macro; it is already linked by `ndk-glue` and is best used from there.  If this PR finds its way to master one way or another, how about removing this "path override" feature, as there should now be one consistent place to reference `android_logger` from?  We might give `log` the same treatment, even though this is _generally_ available in the same scope to print log messages.

`ndk-glue` doesn't seem to use `android_logger` at all, does it make more sense to move this dependency into `ndk-macro` so that it can be used on its own? There's no `$crate::android_logger` in proc-macro to make this any easier though.


msiglreith:
Relates to https://github.com/rust-windowing/android-ndk-rs/issues/53
On the topic I'm not sure yet how to approach this, I see that's not optimal but haven't looked into these parts very much so far. Maybe we can also minimize the feature set of the macro.

francesca64:
Thanks for calling this out and making a PR! We always want to improve user experience. I'm definitely in favor of macros being self-contained; implicitly requiring the user to add some deps really sucks.

> how about removing this "path override" feature

If we're re-exporting `android_logger`/etc. ourselves, then I agree that we don't need path overrides for it. We'll still need one for `ndk-glue` though; #53 explains the general motive for that feature in depth. 

> `ndk-glue` doesn't seem to use `android_logger` 

The only reason `ndk-macro` is separate from `ndk-glue` is because of the requirements of proc macro crates. They're still conceptually merged though; end-users only directly depend on `ndk-glue`, since `ndk-glue` exports the contents of `ndk-macro`. Therefore, I think it's very reasonable for `ndk-glue` to contain dependencies for `ndk-macro`, since the separation between the two crates is just an implementation detail. I'd even argue that it's the job of the crate that re-exports a macro to also export its dependencies.

> does it make more sense to move this dependency into `ndk-macro` so that it can be used on its own?

I don't think that's possible? Proc macro crates are pretty limited in what they can export.

> There's no `$crate::android_logger` in proc-macro to make this any easier though.

We already have prior art in `ndk-macro` for working around this (also discussed in #53), as un-fun as it is.

MarijnS95:
> > how about removing this "path override" feature
> 
> If we're re-exporting `android_logger`/etc. ourselves, then I agree that we don't need path overrides for it. We'll still need one for `ndk-glue` though; #53 explains the general motive for that feature in depth.

I'll remove the path to `android_logger` as it is not necessary, but what do you think about `log`? It's probably safest to re-export a known-compatible version and remove the path too. Crate users can then leverage `ndk_glue::log` or add the crate to `Cargo.toml`.
(A specific example: what if a user wants to configure logging level without printing any logs themselves, like when attempting to trace something in winit which utilizes `trace!` a bunch?)

> > `ndk-glue` doesn't seem to use `android_logger`
> 
> The only reason `ndk-macro` is separate from `ndk-glue` is because of the requirements of proc macro crates. They're still conceptually merged though; end-users only directly depend on `ndk-glue`, since `ndk-glue` exports the contents of `ndk-macro`. Therefore, I think it's very reasonable for `ndk-glue` to contain dependencies for `ndk-macro`, since the separation between the two crates is just an implementation detail. I'd even argue that it's the job of the crate that re-exports a macro to also export its dependencies.

Right, they are already supposed to be used together if the rest of the code is believed.

> > does it make more sense to move this dependency into `ndk-macro` so that it can be used on its own?
> 
> I don't think that's possible? Proc macro crates are pretty limited in what they can export.

Indeed, I wasn't aware of that. Since you've just made the case for the two crates being heavily tied together there is no reason to move this - it would in fact be impossible which is why this structure is in place already :blush: 

> > There's no `$crate::android_logger` in proc-macro to make this any easier though.
> 
> We already have prior art in `ndk-macro` for working around this (also discussed in #53), as un-fun as it is.

I see, that was clarified there as well. `crate_path` is already used for `ndk-glue` so that's a breeze.

MarijnS95:
Pretty glad all these nice tests are now actually ran by the CI following #89, makes me feel a lot safer even though I made sure to run most of them locally before pushing :)

francesca64:
> what do you think about `log`? It's probably safest to re-export a known-compatible version and remove the path too. Crate users can then leverage `ndk_glue::log` or add the crate to Cargo.toml.

I think that seems reasonable. Is compatibility much of a problem here? `log` seems super stable.

> (A specific example: what if a user wants to configure logging level without printing any logs themselves, like when attempting to trace something in winit which utilizes `trace!` a bunch?)

That wouldn't actually involve using the `log` crate, right? Either way, I'm glad to hear winit produces a lot of logging! Maybe [this issue](https://github.com/rust-windowing/winit/issues/535) should be marked as resolved...

MarijnS95:
> > what do you think about `log`? It's probably safest to re-export a known-compatible version and remove the path too. Crate users can then leverage `ndk_glue::log` or add the crate to Cargo.toml.
> 
> I think that seems reasonable. Is compatibility much of a problem here? `log` seems super stable.

Agreed there should be no problem in android-ndk using a "random" `log` version from the surrounding program/scope, leaving usage ergonomics as the sole argument.

> > (A specific example: what if a user wants to configure logging level without printing any logs themselves [...])
> 
> That wouldn't actually involve using the `log` crate, right?

It doesn't involve the log crate if one merely wants to configure the logger with `logger(level = "trace", tag = "my-tag")`, suffering the same problem as `android_logger` outlined above. Pushed the change necessary to pull `log` from `ndk_glue` as well.

> Either way, I'm glad to hear winit produces a lot of logging! Maybe [this issue](https://github.com/rust-windowing/winit/issues/535) should be marked as resolved...

Unfortunately nope. Turns out that's merely a `trace!` call in `ndk_glue::wake`, printed for every event that arrives.

MarijnS95:
@francesca64 @msiglreith Is anything left to be discussed here? If I remember correctly (it's been a while) this should be ready for review and merging actually :)

ndk-examples/Cargo.toml
```diff
@@ -5,8 +5,11 @@ authors = ["David Craven <david@craven.ch>"]
 edition = "2018"
 publish = false
 
+[dependencies]
+log = "0.4.8"
```
MarijnS95, line 9: I'll move this under android-only dependencies again.

francesca64:
> if I remember correctly (it's been a while) this should be ready for review and merging actually :)

That's my recollection too!

MarijnS95:
> That's my recollection too!

Cool! Are you in a position to give this a quick read-through and review?

msiglreith (approved):

MarijnS95:
Documented the changes in `CHANGELOG.md`, I think we're almost good to go :tada:!

francesca64 (approved):
Thanks so much!

MarijnS95:
Thanks for the review all!

PR #89, opened 2020-11-03
ci: Run all tests in isolation

MarijnS95:
When using && to cd back out of a subdirectory the error code from cargo test is implicitly consumed instead of being caught by set -e.  While this does return a non-zero exit code it does not prevent the rest of the script from running if more lines follow.

The next lines run and, if successful, do not fail the step because GH actions base the result of a multiline shell script solely on the last command [1].

You can try this for yourself by throwing the following three steps in GH actions and observing the result [2]:

    false
    echo "This is not printed"

    false && echo "This is not printed"

    false && echo "This is not printed"
    echo "Oh noes, his is printed, and the script exits successfully"

In addition structuring the tests this way allows working-directory to be used, and shows every test individually in a PR to more quickly identify where things are going wrong.

[1]: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference
[2]: https://github.com/MarijnS95/gh-actions-test/actions/runs/343934038

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -59,11 +59,22 @@ jobs:
         target: ${{ matrix.rust-target }}
         override: true
 
-    - name: Run tests
-      run: |
-        cd ndk-sys && cargo test --features test && cd ..
-        cd ndk && cargo test --features rustdoc --doc && cd ..
+    - name: Test ndk-sys
+      working-directory: ndk-sys
+      run:
+        cargo test --features test
+
+    - name: Test ndk
+      working-directory: ndk
+      run:
+        cargo test --features rustdoc --doc
+
+    - name: Test ndk-build
+      run:
         cargo test -p ndk-build
+
+    - name: Test cargo-apk
+      run:
         cargo test -p cargo-apk
 
     - name: Install cargo-apk
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -60,26 +60,24 @@ jobs:
         override: true
 
     - name: Test ndk-sys
-      working-directory: ndk-sys
       run:
-        cargo test --features test
+        cargo test -p ndk-sys --all-features
 
     - name: Test ndk
-      working-directory: ndk
       run:
-        cargo test --features test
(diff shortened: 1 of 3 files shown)


MarijnS95:
```
     Running /home/runner/work/android-ndk-rs/android-ndk-rs/target/debug/deps/ndk-130ea9646c5271e4

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

   Doc-tests ndk

running 4 tests
test src/asset.rs - asset::Asset (line 144) ... ok
test src/asset.rs - asset::AssetDir (line 59) ... ok
test src/native_activity.rs - native_activity::NativeActivity::vm (line 99) ... ok
test src/native_activity.rs - native_activity::NativeActivity::vm (line 89) ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

:tada:

Looks like all tests succeed now after introducing a `test` feature to `ndk` as well. Note that I've removed `--doc` which is implied by `cargo test` (but not by `--all-targets`!), see https://doc.rust-lang.org/cargo/commands/cargo-test.html#target-selection.

msiglreith (approved):
Awesome, thanks!

MarijnS95:
@msiglreith I _just_ pushed more changes to push more features :)

msiglreith:
@MarijnS95 yeah github yelled at me when I tried to merge 😃
let me know when it's fine to merge

MarijnS95:
@msiglreith Pushed again and added `ndk-glue`, if only for `check`ing. Let's see if the tests still pass :sweat_smile:

MarijnS95:
@msiglreith All tests succeeded and this is ready from my perspective. This is ready for review and merging if you don't see anything missing.

Someone should perhaps take a look at the deprecation warning in `BitmapFormat` though, originating from the `TryFromPrimitive`/`IntoPrimitive` macros using the deprecated enum variant `RGBA_4444`. Normally GH picks these up and prints them in the diff (warnings on unchanged lines/files) but it seems to have stopped doing so.

PR #92, opened 2020-11-06
ci: Remove obsolete NDK/Java setup and run once on Windows

MarijnS95:
GitHub already includes the Android SDK/NDK and Java in their image, no need to download and install it again slowing down the build process.

Also run at least one build job on Windows to make sure it remains sane. It didn't seem necessary to run both nightly and stable for every target though.

Note that this is in part to bring to attention an issue with spaces in CC/LINKER paths:

```
error: linking with `C:\Program Files (x86)\Android\android-sdk\ndk-bundle\toolchains\llvm\prebuilt\windows-x86_64\bin\aarch64-linux-android30-clang.cmd` failed: exit code: 1
  |
  = note: "C:\\Program Files (x86)\\Android\\android-sdk\\ndk-bundle\\toolchains\\llvm\\prebuilt\\windows-x86_64\\bin\\aarch64-linux-android30-clang.cmd" "-Wl,--as-needed" "-Wl,-z,noexecstack" "-Wl,--allow-multiple-definition" "-Wl,--eh-frame-hdr" "-L" "C:\\Rust\\.rustup\\toolchains\\stable-x86_64-pc-windows-msvc\\lib\\rustlib\\aarch64-linux-android\\lib" [..]
  = note: 'C:\Program' is not recognized as an internal or external command,
          operable program or batch file.
```

It looks like the code using `CARGO_TARGET_aarch64_linux_android_LINKER` (though the same happens for `CC_/CCX_<triple>`) properly understands that that's one string (seeing the quotes in the log) yet somehow invokes it in the wrong way. Anyone have an idea where to look and resolve this?

(I'm not a fan of the `if: runner.os != 'Windows'` around most tests but we can't really run "Android-native" unix code on Windows)

EDIT: On the note of CC, how about including an [example/test](https://github.com/MarijnS95/android-ndk-rs/commit/compile-c) that validates native compilation?

### DIFF
--- a/ndk-build/src/ndk.rs
+++ b/ndk-build/src/ndk.rs
@@ -17,7 +17,7 @@ impl Ndk {
             let mut sdk_path = std::env::var("ANDROID_HOME").ok();
             if sdk_path.is_some() {
                 println!(
-                    "Warning: You use environment variable ANDROID_HOME that is deprecated.\
+                    "Warning: You use environment variable ANDROID_HOME that is deprecated. \
                  Please, remove it and use ANDROID_SDK_ROOT instead. Now ANDROID_HOME is used"
                 );
             }
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -33,25 +33,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
-    - name: Cache NDK
-      id: cache-ndk
-      uses: actions/cache@v2
-      with:
-        path: android-ndk-r20
-        key: android-ndk-r20-${{ runner.os }}
-
-    - name: Download NDK
-      if: steps.cache-ndk.outputs.cache-hit != 'true'
-      run: |
-        curl -LO https://dl.google.com/android/repository/android-ndk-r20-linux-x86_64.zip
-        unzip android-ndk-r20-linux-x86_64.zip -d $GITHUB_WORKSPACE
-
-    - name: Setup java
-      uses: actions/setup-java@v1
-      with:
-        java-package: jre # Runtime new enough for android toolchain
-        java-version: '13'
-
     - name: Installing Rust ${{ matrix.rust-channel }} w/ ${{ matrix.rust-target }}
       uses: actions-rs/toolchain@v1
       with:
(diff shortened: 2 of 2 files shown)


MarijnS95:
Turns out the problem is in the NDK's compiler wrapper scripts, on both Windows and Linux: neither `$0` nor the result of `dirname $0` is quoted making it unable to deal with spaces. See for example `$ANDROID_NDK_ROOT/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android30-clang`:

```bash
#!/bin/bash
if [ "$1" != "-cc1" ]; then
    `dirname $0`/clang --target=aarch64-linux-android30 "$@"
else
    # Target is already an argument.
    `dirname $0`/clang "$@"
fi
```

The NDK is apparently [not built](https://stackoverflow.com/q/37052934/2844473) to support this case anyway, and while it appears github is [aware of this issue](https://github.com/actions/virtual-environments/issues/1074) they [bailed out](https://github.com/actions/virtual-environments/issues/1122#issuecomment-659309445) of a [potential fix](https://github.com/actions/virtual-environments/pull/1077) at the last moment. Perhaps we should seek to revive that PR or ping the GH environment maintainers about it again, but for now the workaround is to create a symlink ourselves.

ndk-build/src/ndk.rs
```diff
@@ -17,7 +17,7 @@ impl Ndk {
             let mut sdk_path = std::env::var("ANDROID_HOME").ok();
             if sdk_path.is_some() {
                 println!(
-                    "Warning: You use environment variable ANDROID_HOME that is deprecated.\
+                    "Warning: You use environment variable ANDROID_HOME that is deprecated. \
                  Please, remove it and use ANDROID_SDK_ROOT instead. Now ANDROID_HOME is used"
```
MarijnS95, line 21: I'm not a fan of this warning. Any objections to flipping it around and using non-deprecated `ANDROID_SDK_ROOT` (just like `ANDROID_NDK_ROOT` below) before resorting to legacy `ANDROID_HOME`?

We might still print a warning if legacy variables are set, and mention they are not used.

Besides, turns out `ANDROID_NDK_HOME` and `ANDROID_NDK_PATH` are [a thing](https://github.com/actions/virtual-environments/blob/fa9109cc62be11105688bc7aa2305864a9dc762c/images/win/scripts/Installers/Update-AndroidSDK.ps1#L82-L83) as well, though I've never seen `NDK_HOME`.

msiglreith: I think it's derived from here https://developer.android.com/studio/command-line/variables#envar

msiglreith (approved):
Thanks! Also for going down the rabbit-hole of windows cross-compile support 😃

ndk-build/src/ndk.rs
MarijnS95: Right, despite its deprecation it still has precedence over `ANDROID_SDK_ROOT`. If we want to adhere to these rules perhaps we should implement number 3 as well:

> If `ANDROID_HOME` is defined but does not exist or does not contain a valid SDK installation, the value in `ANDROID_SDK_ROOT` is used instead.

msiglreith: Even though I'm also more in favor of dropping support for `ANDROID_HOME` or reordering priorities with `ANDROID_SDK_ROOT` being higher on the list

PR #94, opened 2020-11-11
ndk-build: Support NDK r22

MarijnS95:
`$ndk/platforms` is deprecated and finally removed in r22; android-ndk-rs was never using this so it's safe to remove.

If NDK platform compatibility checking is still desired, we should check for `$ndk/toolchains/${toolchain:-llvm}/prebuilt/${host:-linux-x86_64}/sysroot/usr/lib/$triple/$api/` (Linux ndk, might differ on other platforms) instead.

At the same time perform some minor code cleanup. How are the maintainers feeling about enabling `clippy` in the CI and fixing few remaining issues?

### DIFF
--- a/ndk-build/src/dylibs.rs
+++ b/ndk-build/src/dylibs.rs
@@ -22,11 +22,11 @@ pub fn get_libs_search_paths(
             };
             for line in BufReader::new(File::open(output_file)?).lines() {
                 let line = line?;
-                if line.starts_with("cargo:rustc-link-search=") {
-                    let mut pie = line.split("=");
-                    let (kind, path) = match (pie.next(), pie.next(), pie.next()) {
-                        (Some(_), Some(kind), Some(path)) => (kind, path),
-                        (Some(_), Some(path), None) => ("all", path),
+                if let Some(link_search) = line.strip_prefix("cargo:rustc-link-search=") {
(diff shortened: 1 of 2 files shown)


msiglreith:
> At the same time perform some minor code cleanup. How are the maintainers feeling about enabling clippy in the CI and fixing few remaining issues?

I'm personally fine with it, as I'm using a copy of embarks github action 😄

msiglreith (approved):

MarijnS95:
>> How are the maintainers feeling about enabling clippy in the CI and fixing few remaining issues?
>
> I'm personally fine with it, as I'm using a copy of embarks github action

@msiglreith You mean their `cargo-deny` action? That serves a different purpose than a simple `clippy` job that performs code linting.

msiglreith:
I was referring to the linting parts: https://github.com/EmbarkStudios/opensource-template/blob/main/.github/workflows/rust-ci.yml#L8-L35

MarijnS95:
@msiglreith Right, I expected that after posting. That's the one I'm suggesting to add, but `actions-rs/toolchain` can install components nowadays with `components: rustfmt, clippy`, superseding `rustup component add`. Perhaps worth submitting a PR for that over at Embark as well :smiley:

PR #106, opened 2021-01-04
Manifest (de)serialization

maxded:
This PR contains major changes to the `Manifest` struct in `ndk-build`, changing it to a proper (de)serialization struct. The `Cargo.toml`'s `[package.metadata.android]` is now directly deserialized into `Manifest` and can be serialized into an `AndroidManifest.xml` using [quick-xml](https://github.com/tafia/quick-xml).

The `AndroidManifest` structure matches, [android's manifest structure](https://developer.android.com/guide/topics/manifest/manifest-element) and it's components. Note that not all components have been implemented yet, just the ones that were already implemented.

One downside of my current implementation is the verbosity in the `.toml` file. Having to write `[[package.metadata.android.application.activity.intent_filter]]` is a bit much. If needed this could be improved by implementing `Deserialize` manually (I think, never implemented `Deserialize` myself yet).

All default values remain the same so for a user of `cargo-apk`, the only thing that changes is the syntax in their `Cargo.toml`, the generated `AndroidManifest` is exactly the same.

Hope this is a welcome change :)


msiglreith:
@MarijnS95 would you be fine with the proposed changes as well?

MarijnS95:
@msiglreith Yeah I'm totally content with the proposed refactor, which I suggested to do after wading through the "pipeline" (and stringformatted xml :see_no_evil:). This is still stuck in my review queue, will try to prioritize it first thing tomorrow morning :)

msiglreith (commented):
In general I'm fine with the new manifest structure, thanks for that!
I would prefer to keep some of the defaults tho to better match the old manifest. In particular, it should be possible to run most applications using bare minimum changes to ones cargo toml. As example, the current `ndk-examples` won't execute on my phone due to missing MAIN action intent.

maxded:
All the default values from the previous implementation should be present in my implementation except for 
- activity: label
- activity: launchMode
- activity: screenOrientation

Android provides default values for these components, which are the exact same as the previous default values so I left them out,  should we be explicit here?

msiglreith:
I would go forward with merging it tomorrow if there are no reasons against doing so

MarijnS95:
Sorry for holding this up so long. I'll go through the changes _now_ and let you know.

cargo-apk/src/apk.rs
```diff
@@ -47,49 +47,64 @@ impl<'a> ApkBuilder<'a> {
...
+        // Set default Android manifest values
+        let mut manifest = self.manifest.android_manifest.clone();
+        manifest.package = package_name;
+        manifest.version_name = Some(self.manifest.version.clone());
+        manifest.version_code = Some(VersionCode::from_semver(&self.manifest.version)?.to_code(1));
+
+        manifest.sdk.target_sdk_version = manifest
+            .sdk
+            .target_sdk_version
+            .or(Some(self.ndk.default_platform()));
```
MarijnS95, line 60: Could possibly use `get_or_insert(_with)` here (and below), even though you don't need to _get_ the item (dunno if std already has a function for that usecase):
```suggestion
        manifest
            .sdk
            .target_sdk_version
            .get_or_insert(self.ndk.default_platform());
```

cargo-apk/src/apk.rs
```diff
@@ -47,49 +47,64 @@ impl<'a> ApkBuilder<'a> {
...
-            .as_deref()
-            .unwrap_or(artifact.name())
-            .to_string();
 
-        let config = Config {
+        // Set default Android manifest values
+        let mut manifest = self.manifest.android_manifest.clone();
+        manifest.package = package_name;
+        manifest.version_name = Some(self.manifest.version.clone());
+        manifest.version_code = Some(VersionCode::from_semver(&self.manifest.version)?.to_code(1));
```
MarijnS95, line 55: Not sure if I like these being unconditionally overriden. Don't think we can/should fail during deserialization, but her should be good to let the user know that their value has no effect:

```suggestion
        assert!(manifest.package.is_empty());
        manifest.package = package_name;
        manifest
            .version_name
            .replace(self.manifest.version.clone())
            .expect("version_name should not be set in TOML");
        manifest
            .version_code
            .replace(VersionCode::from_semver(&self.manifest.version)?.to_code(1))
            .expect("version_code should not be set in TOML");
```

cargo-apk/src/apk.rs
```diff
@@ -47,49 +47,64 @@ impl<'a> ApkBuilder<'a> {
...
-            .manifest
-            .metadata
-            .apk_label
-            .as_deref()
-            .unwrap_or(artifact.name())
-            .to_string();
 
-        let config = Config {
+        // Set default Android manifest values
+        let mut manifest = self.manifest.android_manifest.clone();
```
MarijnS95, line 52: Isn't it cleaner to fill this in directly after loading the file in `fn from_subcommand`?

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,345 @@
...
+    #[serde(rename(serialize = "android:configChanges"))]
+    #[serde(default = "default_config_changes")]
+    pub config_changes: Option<String>,
+    #[serde(rename(serialize = "android:label"))]
+    pub label: Option<String>,
+    #[serde(rename(serialize = "android:launchMode"))]
+    pub launch_mode: Option<String>,
+    #[serde(rename(serialize = "android:name"))]
+    #[serde(default = "default_activity_name")]
+    pub name: Option<String>,
```
MarijnS95, line 129: Isn't this also always required to be `android.app.NativeActivity` for native activities?

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,345 @@
...
+/// See https://developer.android.com/guide/topics/manifest/application-element
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct Application {
+    #[serde(rename(serialize = "android:debuggable"))]
+    pub debuggable: Option<bool>,
+    #[serde(rename(serialize = "android:theme"))]
+    pub theme: Option<String>,
+    #[serde(rename(serialize = "android:hasCode"))]
+    #[serde(default = "default_has_code")]
+    pub has_code: Option<bool>,
```
MarijnS95, line 90:
```suggestion
    #[serde(default)]
    pub has_code: bool,
```
That defaults to `false` as well, right?

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,345 @@
...
+                #[serde(rename = "android:name")]
+                pub name: String,
+            }
+            let mut c = Vec::new();
+            for category in categories {
+                c.push(Category {
+                    name: category.clone(),
+                });
+            }
+            serializer.serialize_some(&c)
```
MarijnS95, line 239: When `Option<Vec<>>` is removed you can also use `serialize_seq` here.

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,345 @@
...
 }
 
-#[derive(Debug)]
-pub struct Permission {
-    pub name: String,
-    pub max_sdk_version: Option<u32>,
+/// See https://developer.android.com/guide/topics/manifest/intent-filter-element
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub struct IntentFilter {
+    /// Serialize as struct for proper xml formatting
```
MarijnS95, line 188:
```suggestion
    /// Serialize wrapped in `<action android:name="..." />`
```
Or something like that, making it clear what "proper xml" is supposed to mean (text nodes in xml are valid, just not what Android expects). Same below.

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,345 @@
...
     }
+}
 
+impl AndroidManifest {
     pub fn write_to(&self, dir: &Path) -> Result<(), NdkError> {
         let mut file = File::create(dir.join("AndroidManifest.xml"))?;
-        writeln!(file, "{}", self.to_string())?;
+        let serialized = quick_xml::se::to_string(&self).unwrap();
+        writeln!(file, r#"<?xml version="1.0" encoding="utf-8"?>"#)?;
+        writeln!(file, "{}", serialized)?;
```
MarijnS95, line 76: You shouldn't need to worry about the XML declaration and can instead serialize to the writer directly:
```suggestion
        quick_xml::se::to_writer(file, &self).unwrap();
```

Should we add the error type to `NdkError` so that it can be `?`'d instead of unwrapping?

cargo-apk/README.md
```diff
@@ -26,81 +26,128 @@ cargo install --path .
...
 
-# Adds a uses-feature element to the manifest
-# Supported keys: name, required
 # See https://developer.android.com/guide/topics/manifest/uses-feature-element
+# 
+# Note: there can be multiple .feature entries.
 [[package.metadata.android.feature]]
-name = "android.hardware.camera"
+name = "android.hardware.vulkan.level"
+required = true
```
MarijnS95, line 54: Like the old readme this should have `version = "1"` instead of `required = true`, it's a `level` after all?

MarijnS95 (changes requested):
Looking really good thus far! This review contains a couple questions and suggestions to simplify the code. Things to pay attention to:

- Android requires certain fields to have a specific value. Some are filled with defaults upon deserialization, others are _unconditionally_ overwritten during build. That should be streamlined and made more explicit to the user;
- Tests: While the readme is nice I'd suggest moving it over to an example app that is _actively parsed and tested by the CI_. That way the documentation can hardly ever get out of sync with the actual parsing logic/structure in Rust.

ndk-build/src/manifest.rs
maxded: The default value is currently always set to `android.app.NativeActivity`. I wasn't sure if in the future support for multiple activities would be implemented so I left it as an `Option` in this PR. Could change that of course to always default to `android.app.NativeActivity`

cargo-apk/README.md
maxded: It's a bool based on https://developer.android.com/guide/topics/manifest/uses-feature-element. Should that be changed?

MarijnS95: Seems like it is a bit of both: https://developer.android.com/guide/topics/manifest/uses-feature-element#vulkan-hw-features

Not sure what this is supposed to be, whether it makes sense to infer `required=true`.

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,345 @@
...
+    pub name: Option<String>,
+    #[serde(rename(serialize = "android:screenOrientation"))]
+    pub orientation: Option<String>,
+
+    #[serde(rename(serialize = "meta-data", deserialize = "metadata"))]
+    pub meta_datas: Option<Vec<MetaData>>,
+    /// If no `MAIN` action exists in any intent filter, a default `MAIN` filter is serialized.
+    #[serde(serialize_with = "serialize_intents")]
+    #[serde(rename(serialize = "intent-filter", deserialize = "intent_filter"))]
+    pub intent_filters: Option<Vec<IntentFilter>>,
```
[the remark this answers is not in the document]
MarijnS95: There doesn't seem to be a difference between `Option` and default-empty `vec`, at not one that concerns us, so that should be more ergonomic to write :)

cargo-apk/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Breaking, uses `ndk-build`'s new (de)serialized `Manifest` struct to properly serialize a toml's `[package.metadata.android]` to an `AndroidManifest.xml`. The `[package.metadata.android]` now closely resembles the structure of [`an android manifest file`](https://developer.android.com/guide/topics/manifest/manifest-element). See [README](README.md) for an example of the new `[package.metadata.android]` structure and all manifest attributes that are currently supported.
```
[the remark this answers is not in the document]
maxded: Pushed a new version containing these changes!

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,345 @@
...
-    pub application_metadatas: Vec<ApplicationMetadata>,
-    pub activity_metadatas: Vec<ActivityMetadata>,
-}
+/// See https://developer.android.com/guide/topics/manifest/manifest-element
+#[serde(rename = "manifest")]
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct AndroidManifest {
+    #[serde(rename(serialize = "xmlns:android"))]
+    #[serde(default = "default_namespace")]
+    pub android: String,
```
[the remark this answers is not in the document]
maxded: Pushed a new version containing these changes!

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,345 @@
...
-        };
-        let (major, minor) = self.opengles_version;
-        let opengles_version = format!("0x{:04}{:04}", major, minor);
+    #[serde(rename(serialize = "uses-sdk"))]
+    #[serde(default)]
+    pub sdk: Sdk,
+    /// If no `opengles_version` exists in any feature, a default one of 3.1 is serialized.
+    #[serde(serialize_with = "serialize_features")]
+    #[serde(rename(serialize = "uses-feature", deserialize = "feature"))]
+    pub features: Option<Vec<Feature>>,
```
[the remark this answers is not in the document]
maxded: Removed the original requirement in my latest push if needed we could always add it back in later..

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,345 @@
...
+    #[serde(rename(serialize = "uses-permission", deserialize = "permission"))]
```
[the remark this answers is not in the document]
maxded: Renamed `features` to `uses_feature` and removed the deserialize renaming

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,345 @@
...
-            "".to_string()
-        };
-        let (major, minor) = self.opengles_version;
-        let opengles_version = format!("0x{:04}{:04}", major, minor);
+    #[serde(rename(serialize = "uses-sdk"))]
+    #[serde(default)]
+    pub sdk: Sdk,
+    /// If no `opengles_version` exists in any feature, a default one of 3.1 is serialized.
+    #[serde(serialize_with = "serialize_features")]
+    #[serde(rename(serialize = "uses-feature", deserialize = "feature"))]
```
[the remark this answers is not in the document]
maxded: Ditto to the above comment

ndk-build/src/manifest.rs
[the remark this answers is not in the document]
maxded: Changed all occurrences of  `Option<Vec<T>>` to just `Vec<T>` :)

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,345 @@
...
-        )
+fn serialize_features<S>(features: &Option<Vec<Feature>>, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: Serializer,
+{
+    // Check if features contains an `opengles_version` if not, add a default feature.
+    match features {
+        Some(features) => {
+            if features.iter().all(|f| f.opengles_version.is_none()) {
+                let mut f = features.clone();
```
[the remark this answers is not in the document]
maxded: Thanks for the tip, done!

cargo-apk/src/apk.rs
maxded: Changed it so `cargo-apk` panics when the user specifies these attributes themselves with the suggested error message.

ndk-build/src/manifest.rs
maxded: Done!

ndk-build/src/manifest.rs
```diff
@@ -216,30 +185,30 @@ where
...
+        #[serde(rename = "android:name")]
+        pub name: String,
+    }
+
+    let mut c = Vec::new();
+    for category in categories {
+        c.push(Category {
+            name: category.clone(),
+        });
+    }
```
MarijnS95, line 203: The idea here was that you could just `serialize_element` them directly instead of putting it in a temporary Vec.

ndk-build/src/manifest.rs
```diff
@@ -277,6 +246,11 @@ pub struct Feature {
     pub name: Option<String>,
     #[serde(rename(serialize = "android:required"))]
     pub required: Option<bool>,
+    /// When `name` is FEATURE_VULKAN_HARDWARE_COMPUTE, see: https://developer.android.com/reference/android/content/pm/PackageManager#FEATURE_VULKAN_HARDWARE_COMPUTE
+    /// When `name` is FEATURE_VULKAN_HARDWARE_LEVEL, see: https://developer.android.com/reference/android/content/pm/PackageManager#FEATURE_VULKAN_HARDWARE_LEVEL
+    /// When `name` is FEATURE_VULKAN_HARDWARE_VERSION, see: https://developer.android.com/reference/android/content/pm/PackageManager#FEATURE_VULKAN_HARDWARE_VERSION
```
MarijnS95, line 249: Iirc you're supposed to place links `<https://developer.android.com/reference/android/content/pm/PackageManager#FEATURE_VULKAN_HARDWARE_COMPUTE>` but please double-check the documentation.

It looks like we only test/check the docs now, perhaps need a `cargo test --doc` in a separate PR (perhaps prior to this PR).

MarijnS95 (commented):
That cleans up really nicely! Some nits on the last commit, will check the remaining comments after.

cargo-apk/src/apk.rs
MarijnS95: Oh right, this should have been an `expect_none` which is not stabilized yet :facepalm:

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,313 @@
...
 }
 
-#[derive(Debug)]
-pub struct ActivityMetadata {
-    pub name: String,
-    pub value: String,
+/// See https://developer.android.com/guide/topics/manifest/intent-filter-element
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub struct IntentFilter {
+    /// Serialize rapped in `<action android:name="..." />`
```
MarijnS95, line 152:
```suggestion
    /// Serialize strings wrapped in `<action android:name="..." />`
```

ndk-build/src/manifest.rs
MarijnS95: Rapped :joy:

ndk-build/src/manifest.rs
MarijnS95: You should perhaps write out the _string values_ for these constants, so `"android.hardware.vulkan.level"` and the like. Perhaps also explain in short what the value means, before referring to the Android docs (those links like to break every year or so).

cargo-apk/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- **Breaking:**, uses `ndk-build`'s new (de)serialized `Manifest` struct to properly serialize a toml's `[package.metadata.android]` to an `AndroidManifest.xml`. The `[package.metadata.android]` now closely resembles the structure of [an android manifest file](https://developer.android.com/guide/topics/manifest/manifest-element). See [README](README.md) for an example of the new `[package.metadata.android]` structure and all manifest attributes that are currently supported.
```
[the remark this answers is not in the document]
maxded: Whoops looked over that one.. fixed!

maxded:
@MarijnS95 think we are good to go now!

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,322 @@
...
-        )
+    // Check if `intent_filters` contains a `MAIN` action. If not, add a default filter.
+    if intent_filters
+        .iter()
+        .all(|i| i.actions.iter().all(|f| f != "android.intent.action.MAIN"))
+    {
+        seq.serialize_element(&IntentFilter {
+            actions: vec!["android.intent.action.MAIN".to_string()],
+            categories: vec!["android.intent.category.LAUNCHER".to_string()],
+            data: Vec::new(),
```
MarijnS95, line 143: Idr if `vec![]` or `Default::default` is preferred here. Either way, make it consistent with all the other `Vec<>` initializations in this file. Perhaps `..Default::default()` makes more sense here, though being explicit about the last member is probably the better way to go.

MarijnS95 (changes requested):
@maxded Thanks! We're almost there, just a couple more documentation changes. Be sure to check the unresolved comments above, we should either fix/discuss those here or open an issue to follow up on them.

maxded:
Resolved all the comments above! 

> Be sure to check the unresolved comments above, we should either fix/discuss those here or open an issue to follow up on them.

What exactly did you mean here? Running `cargo test --doc` in CI or?

ndk-build/src/manifest.rs
```diff
@@ -232,30 +232,23 @@ pub struct MetaData {
...
-    /// optional hardware features the app requires.
-    ///
-    /// More detail: <https://developer.android.com/reference/android/content/pm/PackageManager#FEATURE_VULKAN_HARDWARE_LEVEL>
-    ///
-    /// - name "android.hardware.vulkan.compute", `version` indicates the minimum version
-    /// of Vulkan API support the app requires.
-    ///
-    /// More detail: <https://developer.android.com/reference/android/content/pm/PackageManager#FEATURE_VULKAN_HARDWARE_VERSION>
+    /// - `name="android.hardware.vulkan.version"`: The minimum version of the Vulkan API required. See the [Android documentation](https://developer.android.com/reference/android/content/pm/PackageManager#FEATURE_VULKAN_HARDWARE_VERSION)
+    ///    for available levels and the respective Vulkan features required/provided.
```
MarijnS95, line 251: Perhaps worth mentioning that this is a bitmask with the same encoding as Vulkan version numbers, representing the value of `VkPhysicalDeviceProperties::apiVersion` (and has nothing to do with Vulkan features in the sense as the other two features).

MarijnS95:
> What exactly did you mean here? Running `cargo test --doc` in CI or?

The question where to perform manifest default overrides: https://github.com/rust-windowing/android-ndk-rs/pull/106#discussion_r588294744

Testing all docs besides (target-specific) ndk docs was a separate bulletpoint in the "PR review description" apparently; regardless we should look into this.

maxded:
Moved the non-artifact dependent default overrides to the `from_subcommand` function as, like you mentioned, that makes more sense indeed.

MarijnS95 (approved):
> Moved the non-artifact dependent default overrides to the `from_subcommand` function as, like you mentioned, that makes more sense indeed.

Great, thanks! Upon second thought it might have been better situated in `fn build()` (including the `parse_from_toml` call, `cmd` is available in `ApkBuilder` anyway) since no other call needs it, even though they all call `fn build()` now anyway. Bit unfortunate that we still need the clone and `mut` to set one or two extra fields basted on the `Artifact` to build.

Anyway, I guess this looks good now! Haven't given the final code a full read-through but the last diff addresses the remaining issue. Thanks for pulling it all the way through :partying_face: !

@msiglreith Do you want to give this a quick look and make sure I didn't miss anything? Then how do we want to release this, just bump the version and throw it out in the wild, hoping all projects read the changelog and update their toml?

ndk-build/src/manifest.rs
```diff
@@ -1,258 +1,315 @@
...
-    pub activity_metadatas: Vec<ActivityMetadata>,
-}
+/// Android [manifest element](https://developer.android.com/guide/topics/manifest/manifest-element), containing an [`Application`] element.
+#[serde(rename = "manifest")]
+#[derive(Clone, Debug, Deserialize, Serialize)]
+pub struct AndroidManifest {
+    #[serde(rename(serialize = "xmlns:android"))]
+    #[serde(default = "default_namespace")]
+    ns_android: String,
+    #[serde(rename(serialize = "package"))]
```
msiglreith, line 12: can be removed

ndk-build/CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 # Unreleased
 
+- Breaking, refactored `Manifest` into a proper (de)serialization struct. `Manifest` now closely matches [`an android manifest file`](https://developer.android.com/guide/topics/manifest/manifest-element).
```
msiglreith, line 3:
```suggestion
- **Breaking:** refactored `Manifest` into a proper (de)serialization struct. `Manifest` now closely matches [`an android manifest file`](https://developer.android.com/guide/topics/manifest/manifest-element).
```

ndk-build/CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
...
+- Breaking, removed `Config` in favor of using the new `Manifest` struct directly. Instead of using `Config::from_config` to create a `Manifest`, now you instantiate `Manifest` directly using, almost all, the same values.
```
msiglreith, line 4:
```suggestion
- **Breaking:** removed `Config` in favor of using the new `Manifest` struct directly. Instead of using `Config::from_config` to create a `Manifest`, now you instantiate `Manifest` directly using, almost all, the same values.
```

msiglreith (approved):
Really nice and thanks for the review! A few minor nitpicks from me (: 

> Then how do we want to release this, just bump the version and throw it out in the wild, hoping all projects read the changelog and update their toml?

Mostly, yeah 😄 With the current CHANGELOG and example manifest it should be pretty clear on how to transition imo.

maxded:
Done!

MarijnS95:
> Fix cargo doc

Not specifically the docs, but the `rustdoc` run disallows any warnings in documented code (`` /// ```rust `` stuff) and apparently also any code outside it! The test run (https://github.com/rust-windowing/android-ndk-rs/runs/2168935502) finds these issues as well but doesn't exit with a failure code. We'll need to add some explicit `check`/`clippy` runs to the CI in the future...

ndk-build/src/manifest.rs
MarijnS95: So we pushed a similar change in our own codebase where we were writing 60+MB JSON files and guess what, everything slowed down to a halt!

The serializer performs a write for every little thing that it emits, simply because that is how Serde is structured. Since I'm not a huge fan of serializing anything to a string followed by dumping that to a file the best of both worlds is probably a `BufWriter`, that abstracts away the buffering logic you'd get with `to_string` but with a sensible size, performing infrequent writes:

```rust
let file = File::create(dir.join("AndroidManifest.xml"))?;
let w = std::io::BufWriter::new(file);
quick_xml::se::to_writer(w, &self)?;
```

MarijnS95: > completely looked over it

NP: It was my suggestion and I didn't realize this either :). Rusts file writer is just a plain writer that does exactly what it says on the tin - execute the `write()` syscall every time it is called without any hidden batching. Same for Serde, minimizing its own runtime overhead and putting the burden of caching on the writer-backend - which would be suboptimal if writing directly to memory anyway. That is exactly what happens when serializing to a string, that's backed by a `Vec<u8>`! Pretty neat and explicit if you ask me.

Even if these files are small, that count still adds up if a write is invoked for every primitive (`<`/`>`, indentation and spaces, key and value strings, `=`, etc).

cargo-apk/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- **Breaking:** uses `ndk-build`'s new (de)serialized `Manifest` struct to properly serialize a toml's `[package.metadata.android]` to an `AndroidManifest.xml`. The `[package.metadata.android]` now closely resembles the structure of [an android manifest file](https://developer.android.com/guide/topics/manifest/manifest-element). See [README](README.md) for an example of the new `[package.metadata.android]` structure and all manifest attributes that are currently supported.
```
MarijnS95, line 3: See *the* README for...?

MarijnS95 (approved):
This all looks okay to me. I guess we're just too afraid to press the merge button on larger changesets and do a release, but we should push through anyhow. Let's make this happen and give it some wider attention :tada:

PR #108, opened 2021-01-08
ndk: Add missing API levels, use intra-links in AAudio docs

MarijnS95:
Following the removal of `set_callback` in #107 this re-introduces the "proper" doc (according to NDK docs) to `AudioStreamDataCallback`. At the same time the rest of the docs have been upgraded to use intra links (it is hard to resist: start cleaning up one line, but the rest can't be left behind; that'd be inconsistent).

Some names were broken or renamed in Rust, again signifying the relevance of proper doc links. Not only is it useful for readers to click through to a reference, it also ensures that reference is valid at build time making sure the documentation is useful at all times (yeah - you always end up grepping for that one function in the docs, frustrated that it doesn't seem to exist anywhere).

More importantly this adds missing features for API level 25, 27 and 28. 28 is used by `AAudio` and had me wonder why certain functions didn't show up in the docs even with `all`... :grimacing: 

Final note: Links are a bit inconsistent; some use the name of an enum variant (ie. ``[`Variant`](EnumType::Variant)``), others use the full path including type (ie. ``[`EnumType::Variant`]``). That should probably be cleaned up.

### DIFF
--- a/ndk/src/aaudio.rs
+++ b/ndk/src/aaudio.rs
@@ -92,7 +92,7 @@ pub enum AAudioFormat {
     /// Values outside that range may be clipped.
     ///
     /// See also 'floatData' at
-    /// https://developer.android.com/reference/android/media/AudioTrack#write(float[],%20int,%20int,%20int)
+    /// <https://developer.android.com/reference/android/media/AudioTrack#write(float[],%20int,%20int,%20int)>
     PCM_Float = ffi::AAUDIO_FORMAT_PCM_FLOAT,
     /// This format uses the i16 data type.
     /// The maximum range of the data is -32768 to 32767.
--- a/ndk/src/aaudio.rs
+++ b/ndk/src/aaudio.rs
@@ -27,8 +27,8 @@ use thiserror::Error;
 pub enum AAudioAllowedCapturePolicy {
     /// Indicates that the audio may be captured by any app.
     ///
-    /// For privacy, the following usages can not be recorded: `VoiceCommunication*`,
-    /// `Notification*`, `Assistance*` and `Assistant`.
+    /// For privacy, the following [usages](AAudioUsage) can not be recorded:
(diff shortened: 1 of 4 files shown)


MarijnS95:
>  it also ensures that reference is valid at build time

This is not the case currently. `cargo check` doesn't do it, and `-D warnings` cannot be passed to `cargo doc`. Any suggestions?

msiglreith:
This seems to be https://github.com/rust-lang/rust/issues/79792
I guess we just keep it as is for now assuming this will be fixed in near future

MarijnS95:
@msiglreith Thanks! I wasn't aware of `RUSTDOCFLAGS`, but `-Dwarnings` catch `broken_intra_doc_links` just fine. We can leave this enabled and have more warnings (none of which are there currently :grimacing:) be turned to errors over time.

Should we add more documentation checking to the CI? I guess `ndk` and `ndk-glue` are the most interesting for crate users, the latter which seems to contain no documentation whatsoever, not even on `pub` functions and types.

ndk/src/aaudio.rs
```diff
@@ -92,7 +92,7 @@ pub enum AAudioFormat {
     /// Values outside that range may be clipped.
     ///
     /// See also 'floatData' at
-    /// https://developer.android.com/reference/android/media/AudioTrack#write(float[],%20int,%20int,%20int)
+    /// <https://developer.android.com/reference/android/media/AudioTrack#write(float[],%20int,%20int,%20int)>
```
MarijnS95, line 95: Note that this link is broken; rustdoc rightfully converts `[]` to `%5B%5D`, but android.com does not resolve that to the right anchor.

MarijnS95:
Some flukes in the CI, GH actions' virtual environments are still at 1.48 and `actions-rs/toolchain` updates it to 1.49 on every job.

msiglreith (approved):
Thanks again!
CI support is nice and intra-doc links are really nice feature

PR #109, opened 2021-01-09
ndk/aaudio: Fix link to AudioTrack#write

MarijnS95:
Fixes https://github.com/rust-windowing/android-ndk-rs/pull/108#discussion_r554393678

Autolinks escape `[]` resulting in `AudioTrack#write(float%5B%5D,%20int,%20int,%20int)` which is not resolved to the right anchor on android.com. Using `<a>` however does not apply escaping rules, allowing the link to work properly.


MarijnS95:
Ended up fixing it after all, sorry for lots-of-PRs spam :grimacing:

msiglreith (approved):

PR #110, opened 2021-01-12
Update android_logger

bjorn3:
android_logger 0.8.6 depends on android_log-sys 0.1.2. As Bevy depends on both ndk-glue and android_log-sys 0.2, this will cause android_log-sys to be compiled twice. android_logger 0.9.2 however depends on android_log-sys 0.2, avoiding the duplicate dependency.

### DIFF
--- a/ndk-glue/Cargo.toml
+++ b/ndk-glue/Cargo.toml
@@ -18,7 +18,7 @@ ndk-macro = { path = "../ndk-macro", version = "0.2.0" }
 lazy_static = "1.4.0"
 libc = "0.2.66"
 log = "0.4.8"
-android_logger = { version = "0.9.2", optional = true }
+android_logger = { version = "0.9", optional = true }
 
 [features]
 default = []
--- a/ndk-glue/Cargo.toml
+++ b/ndk-glue/Cargo.toml
@@ -18,7 +18,7 @@ ndk-macro = { path = "../ndk-macro", version = "0.2.0" }
 lazy_static = "1.4.0"
 libc = "0.2.66"
 log = "0.4.8"
-android_logger = { version = "0.8.6", optional = true }
+android_logger = { version = "0.9.2", optional = true }
 
 [features]
 default = []


bjorn3:
The CI failure is an unrelated deprecation notice.

MarijnS95:
The CI failure is obscured by `cargo check` output, but that's not the cause. (We should disallow warnings at some point, _and_ make it the last statement in a `run:` block so that it can actually stop the CI; otherwise its error code is simply ignored)

This happens right below:

```
warning: 1 warning emitted

    Finished dev [unoptimized + debuginfo] target(s) in 38.29s
Error: Didn't find Cargo.toml.
Error: Process completed with exit code 1.
```

But it also happens on master since https://github.com/rust-windowing/android-ndk-rs/commit/cb91bb2ebf04b3d0a09b75689842ad3279278f12 - a harmless documentation change. Might be another effect of my recent changes to `cargo-apk`/`cargo-subcommand` only affecting Windows. I'll get it checked. CC @dvc94ch 
(We really need CI there as well...)

ndk-glue/Cargo.toml
```diff
@@ -18,7 +18,7 @@ ndk-macro = { path = "../ndk-macro", version = "0.2.0" }
 lazy_static = "1.4.0"
 libc = "0.2.66"
 log = "0.4.8"
-android_logger = { version = "0.8.6", optional = true }
+android_logger = { version = "0.9.2", optional = true }
```
MarijnS95, line 21: Is there anything particular needed from `.2` that is not available in `.0`? Otherwise I prefer to write just `"0.9"` here though the rest of the file doesn't adhere to that either. What do the other maintainers think?

bjorn3: Switched to `0.9`.

msiglreith (approved):
Thanks, ignorning the spurious CI check for now

MarijnS95:
For completeness the CI check (and in general broken `cargo-apk` on Windows :see_no_evil:) should be fixed in https://github.com/dvc94ch/cargo-subcommand/pull/6.

PR #112, opened 2021-01-13
ndk,ndk-glue: Do not pass Looper ident as userdata pointer

MarijnS95:
The value of `0` or `1` is not a valid pointer to a userdata structure.  Not that it has to be, but with the nasty consequence that winit Android code was accidentally reinterpreting this pointer as integer to extract the identifier from, instead of using the appropriate `ident` field directly. After that is addressed in winit we can set the pointer back to 0 cleanly (and prepare for a potential pointer to some user data that may be passed in the future).

Also add some constants to document these "special" identifiers.

---

Maintainers, this is a breaking change in `ndk-glue` that we want to bump our minor version for, to v0.3. Anyone upgrading now will always see `null()` in the `data` field, wrecking `winit` unless https://github.com/rust-windowing/winit/pull/1826 is in, which will only land in 0.25.

After releasing this we should do a small patch to `winit` to use the new v0.3 crate together with the new constants for clarity.

CC @nwessing: the extra bit of documentation in here might help your adventure into using the glue crate directly!

### DIFF
--- a/ndk-glue/CHANGELOG.md
+++ b/ndk-glue/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Unreleased
 
+- **Breaking** Looper `ident` not passed in `data` pointer anymore.
+  If you are relying on `Poll::Event::data` to tell event fd and
+  input queue apart, please use `Poll::Event::ident` and the new
+  constants introduced in `ndk-glue`!
+
 # 0.2.1 (2020-10-15)
 
 - Fix documentation build on docs.rs
--- a/ndk-glue/CHANGELOG.md
+++ b/ndk-glue/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+# 0.3.0 (2021-01-30)
+
 - **Breaking** Looper `ident` not passed in `data` pointer anymore.
   If you are relying on `Poll::Event::data` to tell event fd and
   input queue apart, please use `Poll::Event::ident` and the new
--- a/ndk-glue/src/lib.rs
+++ b/ndk-glue/src/lib.rs
@@ -16,6 +16,18 @@ use std::thread;
 
 pub use ndk_macro::main;
 
+/// `ndk-glue` macros register the reading end of an event pipe with the
+/// main [`ThreadLooper`] under this `ident`.
+/// When returned from [`ThreadLooper::poll_*`](ThreadLooper::poll_once)
+/// an event can be retrieved from [`poll_events()`].
+pub const NDK_GLUE_LOOPER_EVENT_PIPE_IDENT: i32 = 0;
+
+/// The [`InputQueue`] received from Android is registered with the main
+/// [`ThreadLooper`] under this `ident`.
+/// When returned from [`ThreadLooper::poll_*`](ThreadLooper::poll_once)
+/// an event can be retrieved from [`input_queue()`].
+pub const NDK_GLUE_LOOPER_INPUT_QUEUE_IDENT: i32 = 1;
+
 pub fn android_log(level: Level, tag: &CStr, msg: &CStr) {
     let prio = match level {
         Level::Error => ndk_sys::android_LogPriority_ANDROID_LOG_ERROR,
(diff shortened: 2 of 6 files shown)


ndk-glue/src/lib.rs
```diff
@@ -16,6 +16,18 @@ use std::thread;
 
 pub use ndk_macro::main;
 
+/// `ndk-glue` macros register the reading end of an event pipe with the
+/// main [`ThreadLooper`] under this ident `ident`.
+/// When returned from [`ThreadLooper::poll_*`](ThreadLooper::poll_once)
+/// an event can be retrieved from [`poll_events()`].
```
MarijnS95, line 22: I really wanted to have newlines here, but that makes the second sentence invisible in the documentation unless explicitly clicking on the constant :disappointed:

ndk/src/input_queue.rs
```diff
@@ -71,14 +71,14 @@ impl InputQueue {
         }
     }
 
-    pub fn attach_looper(&self, looper: &ForeignLooper, id: u32) {
+    pub fn attach_looper(&self, looper: &ForeignLooper, id: i32) {
```
MarijnS95, line 74: We should probably add userdata as argument to this function already?

dvc94ch (approved):

ndk/src/input_queue.rs
dvc94ch: you really can't pass user data without making attach_looper unsafe I don't think.

```rust
pub fn attach_looper<'a, T>(&'a self, looper: &ForeignLooper, id: i32, &'a T)
```

looking at the code the lifetime of looper isn't correct, so maybe it should be marked unsafe anyway

MarijnS95: Hmm yeah it doesn't look like the lifetime generic of `RwLockReadGuard` is propagated in `Deref` so we can't really enforce it to be `'static` either. Might hold off adding this until someone needs it and feels like implementing it :)

ndk-glue/src/lib.rs
```diff
@@ -16,6 +16,18 @@ use std::thread;
 
 pub use ndk_macro::main;
 
+/// `ndk-glue` macros register the reading end of an event pipe with the
+/// main [`ThreadLooper`] under this ident `ident`.
```
msiglreith, line 20: minor: should this `ident` be here twice?

ndk-glue/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Unreleased
 
+- Do not pass Looper `ident` as userdata pointer
```
msiglreith, line 3: I think keeping the default formatting here is better
```
- **Breaking:** Looper `ident` not passed as userdata pointer anymore. If you are ...
```

msiglreith (approved):

ndk-glue/src/lib.rs
MarijnS95: Whoops, that's sloppy of me. Thanks for noticing!

ndk-glue/CHANGELOG.md
MarijnS95: Yeah, better to not use the commit title here; but stil keeping the second paragraph. I prefer to use `in` instead of `as` to indicate the variable/member it's provided "into", though I better change userdata to `data` then to refer to the field.

MarijnS95:
@msiglreith I'm thinking about merging this and releasing a minor version bump this weekend, followed by an MR in winit to use the new version and constants. Any other breaking changes we might want to get in?

(Manifest changes - which I also intend to _finally_ look at - are only build-related fortunately)

msiglreith:
Sounds good to me. I think there is nothing else in the pipeline right now.

MarijnS95:
@msiglreith Shall we bump the version at the same time, or in one PR? Also, do we want to keep `ndk-sys` versioning in sync with `ndk`, even though no changes have been made?

msiglreith:
@MarijnS95 Doing the version bump in this PR is fine.

> Also, do we want to keep ndk-sys versioning in sync with ndk, even though no changes have been made?

I would prefer to keep it at the current version in this case.

MarijnS95:
@msiglreith neat, so the `0.2.1` version match was just a coincidence, nothing intentional.

Let's merge this then :)

PR #119, opened 2021-01-29
Bump num-enums 

maxded:
Small bump getting `num-enums` to the latest version

### DIFF
--- a/ndk/Cargo.toml
+++ b/ndk/Cargo.toml
@@ -32,7 +32,7 @@ api-level-30 = ["api-level-29"]
 test = ["ffi/test", "jni", "jni-glue", "all"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


ndk/Cargo.toml
```diff
@@ -32,7 +32,7 @@ api-level-30 = ["api-level-29"]
 test = ["ffi/test", "jni", "jni-glue", "all"]
 
 [dependencies]
-num_enum = "0.4.2"
+num_enum = "0.5.1"
```
MarijnS95, line 35: Feel free to drop the leading `.1`, for the same reason [outlined here](https://github.com/rust-windowing/android-ndk-rs/pull/110#discussion_r555750738).

MarijnS95 (approved):
Thanks!

PR #120, opened 2021-01-31
Update deps.

dvc94ch:


MarijnS95:
@dvc94ch Any particular reason to re-add minor and patch versions? I asked to omit these in two recent PRs because we don't depend on a specific minor/patch version and semver allows bumping everything but the most-significant version field _anyway_.

Assuming you used a tool like "crates" in VSCode to bump these all at once, that will now unnecessarily say the specified version is out-of-date really quick for crates that see fast releases on their patch version.

Shall I send a PR to simplify these again?

dvc94ch:
I just run `cargo upgrade --workspace`, I don't see much benefit in simplifying the version numbers. Unless you specify `=a.b.c` cargo will still use the latest patch version and if vscode reports differently, then vscode is broken. But if you prefer...

MarijnS95:
I think you missed this option then:

```
--skip-compatible     Only update a dependency if the new version is semver incompatible
```

That is - IMO - the only reason for making updates in `Cargo.toml`: if it brings us a new version that would not be allowed otherwise.

It's not VSCode itself btw, just an extension :)

PR #129, opened 2021-03-15
ci: Check all crate documentation in addition to target-specific ndk

MarijnS95:
Following https://github.com/rust-windowing/android-ndk-rs/pull/106#pullrequestreview-565655645

Only `ndk` docs are built for all `android` targets but the rest of the documentation remains untested. Especially now that we're expanding to provide more complete docs for ndk-build and cargo-apk should everything be tested, to make sure that things like intra-doc links remain in working order.

## TODO

The `ndk` docs build successfully here thanks to the `test` feature (enabled by `--all-features`); does it still make a lot of sense to build-test the documentation for every target individually?

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -73,6 +73,13 @@ jobs:
       run:
         cargo test -p cargo-apk --all-features
 
+    - if: runner.os != 'Windows'
+      name: Document all crates
+      env:
+        RUSTDOCFLAGS: -Dwarnings
+      run:
+        cargo doc --all --all-features
+
     - name: Install cargo-apk
       run:
         cargo install --path cargo-apk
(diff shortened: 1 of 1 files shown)


msiglreith:
> does it still make a lot of sense to build-test the documentation for every target individually?

I don't expect any target specific doc parts tbh. Doing it once for one would be fine as well imo.

MarijnS95:
@msiglreith I guess it's safe to remove then, and have the documentation depend on the `test` flag (and iirc a `cfg(doc)` feature is on its way, though it was disabled again on a repo I was contributing to due to it not being propagated to all crates properly).

MarijnS95:
@msiglreith Removed it :)

Btw we have a lot of build steps now that do not use `--target` but are repeatedly ran for every `rust-target` in the matrix, wasting a ton of time and resources. As soon as this is merged I want to separate those out into two jobs that can run in parallel: some that build/test the host tools and docs, while the other simply only installs cargo-apk (bit of overlap with the build in the first job) followed by just the target-specific `ndk` check and apk build.

msiglreith (approved):

PR #131, opened 2021-03-20
ndk-glue: Don't return from onCreate until LOOPER is created

### LINKED ISSUE #116: Sometimes running apps panicked on the Android Emulator
Sometimes running apps panicked on the Android Emulator with following error:

```bash
01-20 22:05:38.816  3402  3421 I RustStdoutStderr: thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/ndk-glue-0.2.1/src/lib.rs:255:34
01-20 22:05:38.816  3402  3421 I RustStdoutStderr: stack backtrace:
01-20 22:05:38.816  3402  3421 I RustStdoutStderr:    0:     0x7b6194f50110 - <unknown>
```

This error is constantly repeated in CI: https://github.com/gfx-rs/gfx/runs/1738010575?check_suite_focus=true

Sometimes I saw this error on my dev machine (MacOS host), but restart the Android Emulator fixed it. And I never saw this error on a real device.

MarijnS95:
Android only performs additional initialization like providing an input queue through `onInputQueueCreated` after the main initialization function (`ANativeActivity_onCreate`) returns.  In some cases it is possible that the main thread spawned here hasn't attached a looper to its thread and assigned it to the accompanying `LOOPER` static variable yet, resulting in a `None` unwrap when Android provides us with an input queue.

This race condition is simply solved by not returning from `fn init()` (`ANativeActivity_onCreate`) until the thread has finished its setup, through wrapping `LOOPER` in a `Mutex` and using a condition variable to signal when it is ready.

This condition is intentionally not exposed to `on_input_queue_created` may we ever have a cleanup procedure that sets `LOOPER` back to `None`; any call to `onInputQueueCreated` in that state would be an error and should not block indefinitely.  The `LOOPER` should simply be ready by the time `fn init()` returns.

Fixes #116
CC @Gordon-F

### DIFF
--- a/ndk-glue/src/lib.rs
+++ b/ndk-glue/src/lib.rs
@@ -11,7 +11,7 @@ use std::io::{BufRead, BufReader};
 use std::os::raw;
 use std::os::unix::prelude::*;
 use std::ptr::NonNull;
-use std::sync::{RwLock, RwLockReadGuard};
+use std::sync::{Arc, Condvar, Mutex, RwLock, RwLockReadGuard};
 use std::thread;
 
 pub use ndk_macro::main;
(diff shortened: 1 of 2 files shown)


Gordon-F:
@MarijnS95 Thank you! Tested this patch in my case. Works well.

msiglreith (approved):

PR #133, opened 2021-03-20
Test hello-world example with Android Emulator on CI

Gordon-F:


.github/workflows/emulator.yml
```diff
@@ -0,0 +1,26 @@
...
+      continue-on-error: true
+      steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: run tests
+        uses: reactivecircus/android-emulator-runner@v2
+        with:
+          api-level: 29
+          arch: x86_64
+          # the `googleapis` emulator target considerably slower on CI.
```
MarijnS95, line 17:
```suggestion
          # the `googleapis` emulator target is considerably slower on CI.
```

.github/workflows/android_test.sh
```diff
@@ -0,0 +1,26 @@
...
+else
+    exit 1
+fi
+
+if grep -e 'thread.*panicked at' ~/logcat.log;
+then
+    exit 1
+else
+    exit 0
+fi
```
MarijnS95, line 26: Should these `echo ::error::something` so that the failure reason can be read from a CI "annotation" instead of having to download+grep the logs?

MarijnS95 (approved):
Haven't reviewed in detail, but looks cool and simple enough!

.github/workflows/android_test.sh
MarijnS95: That could work as well, but having at least the `panicked at` line with the panic reason in an `::error::` creates a nice annotation in CI run "overview".

https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message

MarijnS95: Perhaps it's worth experimenting some with these, introducing an intentional panic and seeing what is printed in the final output.

You can possibly substitute this entire if block with:

```sh
! (grep -e 'thread.*panicked at' ~/logcat.log | sed 's/^/::error::/')
```

dvc94ch: Is it good enough as is to merge?

Gordon-F: @dvc94ch Sorry, completely forgot about this PR. I will fix this and will squash commits.

MarijnS95: @Gordon-F Yeah I guess that works well enough. Bonus would be to include the stacktrace, but that's more annoying to parse and it's probably more relevant to download and read the whole logs from artifacts in that case.

Note that the test currently isn't working, can you look into that? It looks like the app hasn't started at all.

Gordon-F: Used a wrong tag. Fixed now.

MarijnS95 (approved):
Awesome, the test works now! We should probably come up with a more sophisticated test-case at some point, but validating app startup and `"hello world"` being printed is an important first milestone. Thanks!

MarijnS95:
@Gordon-F Would you be able to help out? For quite some time now the CI seems to be unresponsive around `adb install`, timing out the job 3hrs after:

https://github.com/rust-windowing/android-ndk-rs/actions/runs/1452352305

MarijnS95:
I vaguely remember writing down some more info in https://github.com/rust-windowing/android-ndk-rs/pull/169, too.

MarijnS95:
@Gordon-F I bumped it from `macos-10.15` to `macos-latest` in https://github.com/rust-windowing/android-ndk-rs/commit/b665be0ae32ebbab82b5c3079ea8d3ed90ca3962, and vaguely remember it started happening exactly since that patch. However:

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#github-hosted-runners

> The macos-latest label currently uses the macOS 10.15 runner image.

So it should not have made a difference?

Gordon-F:
Also, the emulator starts 3 times for 3 different tasks, if I understood everything correctly. Is it necessary? It seems to me that it would be more optimal to start it only 1 time and run all tests on it.

MarijnS95:
The broken run also shows that it was on 10.15:

```
Current runner version: '2.284.0'
Operating System
  Mac OS X
  10.15.7
  19H1419
```

> Also, the emulator starts 3 times for 3 different tasks, if I understood everything correctly. Is it necessary? It seems to me that it would be more optimal to start it only 1 time and run all tests on it.

But we'd like to have a clean emulator for every artifact that we need to test. Could deinstall the application but that seems far from ideal. However, perhaps starting 3 of them instead of one is the problem "somehow" in scheduling?

Gordon-F:
> But we'd like to have a clean emulator for every artifact that we need to test.

[Android-emulator-runner docs](https://github.com/ReactiveCircus/android-emulator-runner#usage) suggest using a clean snapshot to speed up CI in this case. What do you think about it? It should speed up CI time and maybe help with the runner issue.

MarijnS95:
Yeah, we can try that :)

MarijnS95:
It's a bit strange because this shouldn't happen regardless (I kept the `set -x` from aforementioned PR #169, and it doesn't show anything out of the ordinary).

MarijnS95:
@Gordon-F It also seems to have been fixed by disabling animations, GPU etc from that page, see #198. We can add caching in a separate PR and see how well that speeds things up separate from fixing the regression lockup.

PR #134, opened 2021-03-21
Notify event pipe before releasing NativeActivity resources

### LINKED ISSUE #117: ndk-glue does not handle NativeWindow lifecycle correctly
As far as I understand, `ndk-glue`'s `NativeActivity` callbacks make it impossible or at least very hard to handle the `NativeWindow` lifecycle in a manner that conforms with the Android documentation. The chief offending callback is [`on_window_destroyed`](https://github.com/rust-windowing/android-ndk-rs/blob/13914ef4c9ff39af289e26bb2f5d1ff12f91ca18/ndk-glue/src/lib.rs#L242).

[`ANativeActivityCallbacks` documentation](https://developer.android.com/ndk/reference/struct/a-native-activity-callbacks#struct_a_native_activity_callbacks_1a6aaafbbfae4c0ac066b9d61ebb3ab97f) seems to demand that `NativeWindow` cleanup code be executed in a synchronized manner, i.e. before the callback returns. For example, one might wish to ensure OpenGL is done drawing and destroy the associated `EGLSurface`.

[NDK `android_native_app_glue`](https://cs.android.com/android/platform/superproject/+/master:prebuilts/ndk/r21/sources/android/native_app_glue/android_native_app_glue.c) appears to follow this guideline, cf. especially how lifecycle event handling is split into pre-user-callback and post-user-callback parts (lines 89 and 146 respectively).

EDIT: for added context, compare [example usage](https://github.com/android/ndk-samples/blob/master/native-activity/app/src/main/cpp/main.cpp) of NDK `android_native_app_glue` feature.

maciejgodek:
Currently, some `NativeActivity` callbacks (`on_native_window_destroyed`, `on_input_queue_destroyed`):
(1) acquire a write lock on a static handle to the resource that is about to be destroyed, (2) drop the handle, (3) dispatch an asynchronous `Event` to the user on a UNIX pipe notifying of the release, (4) immediately return from the callback.

This is incorrect according to [`ANativeActivityCallbacks docs`](https://developer.android.com/ndk/reference/struct/a-native-activity-callbacks#struct_a_native_activity_callbacks_1a6aaafbbfae4c0ac066b9d61ebb3ab97f). If the user is holding on to a read lock on the handle (as they should), there will be a deadlock. If the user does not have a read lock on the handle, there is a race condition, since Android may free the resource before the user cleans up all the objects derived from that resource (e.g. `EGLSurface` from `NativeWindow`).

In order to prevent races, `ndk_glue` should (1) notify users of upcoming resource release, so they can correctly clean up derived objects, and only then (2) acquire a write lock on the handle, (3) drop the handle, (4) return from the callback.

Closes #117

### DIFF
--- a/ndk-glue/src/lib.rs
+++ b/ndk-glue/src/lib.rs
@@ -260,8 +260,8 @@ unsafe extern "C" fn on_window_destroyed(
     activity: *mut ANativeActivity,
     _window: *mut ANativeWindow,
 ) {
-    *NATIVE_WINDOW.write().unwrap() = None;
     wake(activity, Event::WindowDestroyed);
+    *NATIVE_WINDOW.write().unwrap() = None;
 }
 
 unsafe extern "C" fn on_input_queue_created(
@@ -279,10 +279,12 @@ unsafe extern "C" fn on_input_queue_destroyed(
     activity: *mut ANativeActivity,
     queue: *mut AInputQueue,
 ) {
+    wake(activity, Event::InputQueueDestroyed);
+    let mut input_queue_guard = INPUT_QUEUE.write().unwrap();
+    assert_eq!(input_queue_guard.as_ref().unwrap().ptr().as_ptr(), queue);
     let input_queue = InputQueue::from_ptr(NonNull::new(queue).unwrap());
     input_queue.detach_looper();
-    *INPUT_QUEUE.write().unwrap() = None;
-    wake(activity, Event::InputQueueDestroyed);
+    *input_queue_guard = None;
 }
 
 unsafe extern "C" fn on_content_rect_changed(activity: *mut ANativeActivity, rect: *const ARect) {


ndk-glue/src/lib.rs
```diff
@@ -279,10 +279,10 @@ unsafe extern "C" fn on_input_queue_destroyed(
     activity: *mut ANativeActivity,
     queue: *mut AInputQueue,
 ) {
+    wake(activity, Event::InputQueueDestroyed);
     let input_queue = InputQueue::from_ptr(NonNull::new(queue).unwrap());
     input_queue.detach_looper();
     *INPUT_QUEUE.write().unwrap() = None;
```
MarijnS95, line 285: For safety I wonder if we should assert here that `queue == INPUT_QUEUE.ptr()`. That also means the lock should be moved up so that the entire comparison + `detach_looper` + `= None` assignment happens while it is held:

```suggestion
    wake(activity, Event::InputQueueDestroyed);
    let mut write_guard = INPUT_QUEUE.write().unwrap();
    let input_queue = InputQueue::from_ptr(NonNull::new(queue).unwrap());
    assert_eq!(queue, write_guard.as_ref().unwrap().ptr().as_ptr());
    input_queue.detach_looper();
    *write_guard = None;
```

MarijnS95:
To be ever so slightly more specific, could you mention in the _commit_ message that acquiring a writelock in step`(2)` assumes the user holds a readlock on the resource to be destroyed (`INPUT_QUEUE` or `NATIVE_WINDOW`) that it should only release _after_ cleaning up dependent resources like surfaces/swapchains and the like?

maciejgodek:
Ok, I think I fixed it. :) Apologies for the repeated force pushes, I can't focus today.

MarijnS95 (commented):
Commit description is completely on point, I couldn't have worded it better, thanks!

Assigning this for final review to @dvc94ch since he originally discussed this with you in #117.

MarijnS95 (approved):

dvc94ch (approved):
Thanks!

PR #137, opened 2021-03-26
ndk-build: Use LLVM binutils when GNU binutils are removed (r23+)

### LINKED ISSUE #135: Error: Path "android-sdk/ndk/23.0.7196353/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android-ar" doesn't exist.
Not seeing this file in ndk version 23, it is present in version 22.

MarijnS95:
Since [r21] LLVM binutils are included _for testing_;
Since [r22] GNU binutils are deprecated in favour of LLVM's;
Since [r23] GNU binutils have been removed.

When GNU binutils are not available (`<triple>-<bin_name>`) "fall back" to `llvm-<bin_name>` binaries from LLVM binutils that are now included with the NDK.

To maintain stability with the current ndk-build crate release, prefer GNU binutils for as long as it is provided by the NDK instead of trying to use `llvm-<bin_name>` from r21 onwards.

[r21]: https://github.com/android/ndk/wiki/Changelog-r21
[r22]: https://github.com/android/ndk/wiki/Changelog-r22
[r23]: https://github.com/android/ndk/wiki/Changelog-r23

Fixes #135

---

We could also opt for LLVM binutils in r22 (requires extra logic to figure out whether we're building on r21 or r22) where GNU binutils are technically deprecated, but since the current crate release already uses that successfully there seems no need to change it at risk of breaking projects that are somehow incompatible.

### DIFF
--- a/ndk-build/src/ndk.rs
+++ b/ndk-build/src/ndk.rs
@@ -196,22 +196,35 @@ impl Ndk {
         Ok((clang, clang_pp))
     }
 
-    pub fn toolchain_bin(&self, bin: &str, target: Target) -> Result<PathBuf, NdkError> {
+    pub fn toolchain_bin(&self, name: &str, target: Target) -> Result<PathBuf, NdkError> {
         #[cfg(target_os = "windows")]
         let ext = ".exe";
         #[cfg(not(target_os = "windows"))]
         let ext = "";
 
-        let bin = self.toolchain_dir()?.join("bin").join(format!(
-            "{}-{}{}",
-            target.ndk_triple(),
-            bin,
-            ext
-        ));
-        if !bin.exists() {
-            return Err(NdkError::PathNotFound(bin));
+        let toolchain_path = self.toolchain_dir()?.join("bin");
+
+        // Since r21 (https://github.com/android/ndk/wiki/Changelog-r21) LLVM binutils are included _for testing_;
+        // Since r22 (https://github.com/android/ndk/wiki/Changelog-r22) GNU binutils are deprecated in favour of LLVM's;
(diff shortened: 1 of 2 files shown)


ndk-build/src/ndk.rs
```diff
@@ -196,22 +196,29 @@ impl Ndk {
...
+        // as long as it is provided by the NDK instead of trying to use llvm-* from r21 onwards.
+        let gnu_bin = toolchain_bin.join(format!("{}-{}{}", target.ndk_triple(), name, ext));
+        if gnu_bin.exists() {
+            Ok(gnu_bin)
+        } else {
+            let llvm_bin = toolchain_bin.join(format!("llvm-{}{}", name, ext));
+            llvm_bin
+                .exists()
+                .then(|| llvm_bin)
+                .ok_or_else(|| NdkError::PathNotFound(gnu_bin))
```
MarijnS95, line 220: Not sure if we should print the GNU or LLVM bin in the error. Theoretically LLVM is only going to become relevant in the future, but for now we might want to have both... "Could not find <gnu binutils> tool nor <llvm binutils> tool".

msiglreith: A dedicated error case would be nice imo

MarijnS95: I totally agree. How's this:

```console
me:android-ndk-rs/ $ cargo apk run -p ndk-examples
Error: GNU toolchain binary `aarch64-linux-android-ar` nor LLVM toolchain binary `llvm-ar` found in `"/home/marijn/Android/Sdk/ndk/23.0.7196353/toolchains/llvm/prebuilt/linux-x86_64/bin"`
```

msiglreith (approved):

PR #141, opened 2021-04-16
Added support for adding extra files to the APK.

agrande:
This replaces #138 as a more generic solution.

### DIFF
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -88,6 +88,11 @@ res = "path/to/res_folder"
 # If not specified, assets will not be included in the APK
 assets = "path/to/assets_folder"
 
+# Path to a folder containing a hierarchy of extra files to be added at the root of the APK.
+# All files will be added to the APK recursively.
+# If not specified, no extra files will be added to the APK
+extra_content = "path/to/extra_content_folder"
+
 # Adds application metadata to the manifest
 # Note that there can be several application_metadatas entries
 # this will add: <meta-data android:name="com.samsung.android.vr.application.mode" android:value="vr_only"/>
--- a/ndk-build/src/apk.rs
+++ b/ndk-build/src/apk.rs
@@ -3,8 +3,11 @@ use crate::error::NdkError;
 use crate::manifest::Manifest;
 use crate::ndk::{Key, Ndk};
 use crate::target::Target;
-use std::path::{Path, PathBuf};
 use std::process::Command;
+use std::{
+    fs,
+    path::{Path, PathBuf},
+};
 
 pub struct ApkConfig {
     pub ndk: Ndk,
@@ -12,6 +15,7 @@ pub struct ApkConfig {
     pub assets: Option<PathBuf>,
     pub res: Option<PathBuf>,
     pub manifest: Manifest,
+    pub extra_content: Option<PathBuf>,
 }
 
 impl ApkConfig {
@@ -77,6 +81,7 @@ impl ApkConfig {
             assets: config.assets,
             res: config.res,
             manifest,
+            extra_content: config.extra_content,
         }
     }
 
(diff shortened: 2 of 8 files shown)


cargo-apk/README.md
```diff
@@ -88,6 +88,11 @@ res = "path/to/res_folder"
 # If not specified, assets will not be included in the APK
 assets = "path/to/assets_folder"
 
+# Path to a folder containing a hierarchy of extra files to be added at the root of the APK.
+# All files will be added to the APK recursively.
+# If not specified, no extra files will be added to the APK
+extra_content = "path/to/extra_content_folder"
```
dvc94ch, line 94: I guess assets/resources are intended to be relative paths included in version control, while in your use case it's probably an os specific path. Also this doesn't help select the right library for the target architecture. Are there additional use cases for extra-content?

agrande: Are you suggesting that I write the example as an absolute path?
In my particular use-case I do have the shared library file under my repository directory (although not included in version control), so I'm using a relative path.

Regarding selection of the right library for the target architecture, if the APK contains multiple architecture, I think what happens is that `dlopen()` already appropriately uses the correct set of paths (including the proper `lib` subfolders in the APK) depending on which architecture is used when running the app.

As for additional use cases, honestly, I can't think of any off the top of my head.

dvc94ch: I don't think we produce fat apk's because that can lead to very large apks. We could add the feature.

MarijnS95: @dvc94ch Apparently `cargo-apk` already produces fat apks when more target architectures are specified in `build_targets`. I don't know if separate APKs can be distributed anyway unless we dive into the new "app bundles" system.

Regarding assets/resources we have to be _very careful_ that those still travel the normal `aapt package` `-A`/`-S` as they need additional processing and are not "zipped" into the target APK verbatim.

ndk-build/src/apk.rs
```diff
@@ -155,6 +160,41 @@ impl<'a> UnalignedApk<'a> {
...
+        if !aapt.status()?.success() {
+            return Err(NdkError::CmdFailed(aapt));
+        }
+        Ok(())
+    }
+
+    fn add_entry(&self, base_path: &Path, path: &Path) -> Result<(), NdkError> {
+        let full_path = base_path.join(path);
+        if full_path.is_dir() {
+            for entry in fs::read_dir(full_path)? {
```
MarijnS95, line 180: Can't we pass a directory to `aapt add` so that it adds them recursively?

If not you're better off building an array of filenames and adding them to the archive in one go, spawning an `aapt add` process per file can potentially become very costly and wasteful.

ndk-build/src/apk.rs
```diff
@@ -155,6 +160,41 @@ impl<'a> UnalignedApk<'a> {
...
     }
 
+    pub fn add_file(&self, base_path: &Path, path: &Path) -> Result<(), NdkError> {
+        let out = self.0.build_dir.join(path);
+        let out_dir = out.parent().unwrap();
+        std::fs::create_dir_all(&out_dir)?;
+        std::fs::copy(base_path.join(path), out)?;
+
+        let mut aapt = self.0.build_tool(bin!("aapt"))?;
+        aapt.arg("add").arg(self.0.unaligned_apk()).arg(path);
```
MarijnS95, line 170: Should we make sure that `aapt` runs relative to the build directory here? There are environment variables dealing with moved `target/` directories and I don't exactly remember how they are handled here, if we keep running in the right PWD.

If we do that this might as well run `aapt` with `base_path` as `PWD`.

ndk-build/src/apk.rs
```diff
@@ -155,6 +160,41 @@ impl<'a> UnalignedApk<'a> {
...
+        let full_path = base_path.join(path);
+        if full_path.is_dir() {
+            for entry in fs::read_dir(full_path)? {
+                let entry = entry?;
+                let tgt_path = path.join(entry.file_name());
+                self.add_entry(base_path, &tgt_path)?;
+            }
+        } else if full_path.is_file() {
+            self.add_file(base_path, path)?;
+        }
```
MarijnS95, line 187: May need an error if it's not a directory nor a file (ie. a special file like a character device) just in case?

ndk-build/src/apk.rs
```diff
@@ -155,6 +160,41 @@ impl<'a> UnalignedApk<'a> {
...
+        Ok(())
+    }
+
+    pub fn add_content_dir_recursively(&self, content_dir: &Path) -> Result<(), NdkError> {
+        if !content_dir.is_dir() {
+            return Err(NdkError::PathInvalid(content_dir.to_path_buf()));
```
MarijnS95, line 193: I'd say something along the lines of "Is not a directory". That way the path can either not exist at all, or be a file. Invalid is a tad too generic IMO.

MarijnS95 (commented):
How does this deal with symlinks?

MarijnS95:
One thing to note while seeing #142 pop up just a couple hours ago: a generic codepath might not be desired for libraries. `add_lib_recursively` takes care of using `readelf` to find and embed extra dynamic library dependencies, which you would now have to explicitly specify/copy yourself. For example where your Rust library does not need `libc++_shared.so` from the NDK but one of the copied libs does.

agrande:
#142 looks good to me, I'll test it against my use-case, but it's probably a better solutions for various reasons:
- automatic dependencies scan, as you mentioned
- there may not be that many additional use-cases for adding arbitrary files, and providing the feature may push users to do things wrongly (regarding resources and assets)
- it has a smaller code footprint

MarijnS95:
@agrande @kanerogers then I prefer #138 which was here first and takes care of the `target` appropriately by only including the target ABIs (can still be multiple!) that are actually defined in `Cargo.toml`.

agrande:
Ok, then, I'll re-open and modify #138 to use `add_lib_recursively` instead, and specify the directory through config.

kanerogers:
> @agrande @kanerogers then I prefer #138 which was here first and takes care of the `target` appropriately by only including the target ABIs (can still be multiple!) that are actually defined in `Cargo.toml`.

Ah, that's a very good point. I hadn't thought of that!

PR #142, opened 2021-04-20
Allow adding additional libraries to APK

kanerogers:
This change adds the additional `libs` parameter to `package.metadata.android`.

I added this so specifically address edge cases where my final library did not link against a particular object, but it was still required at run-time - eg. vulkan validation layers.

I tried to go with the "most obvious" approach, but if you have something else in mind, please just let me know. I tested this with my Android device and it seemed to work quite well.

I've added some simple documentation to `README.md` but if you need anything further, I'd be happy to add it.

### DIFF
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -101,6 +101,10 @@ value = "vr_only"
 [[package.metadata.android.activity_metadatas]]
 name = "com.oculus.vr.focusaware"
 value = "true"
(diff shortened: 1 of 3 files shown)


kanerogers:
Oh, comedy. I didn't see https://github.com/rust-windowing/android-ndk-rs/pull/141 ! Never mind. :)

PR #146, opened 2021-05-03
Add support for android string resource id in application label. (#145)

### LINKED ISSUE #145: allow use of android string resource id in application label
Cargo.toml:
```
[package.metadata.android]
resources = "res"

[package.metadata.android.application]
label = "@string/app_name"
```

res/values/strings.xml:
```
<?xml version="1.0" encoding="utf-8"?>
<resources>
    <string name="app_name">Localized App Label</string>
</resources>
```

`cargo apk build` failed with message:
```
ERROR: unable to open '/home/user/apkproject/target/debug/apk/@string/app_name-unaligned.apk' as Zip file for writing
```

xzenodart:
Fixes #145

Application label was used for apk file naming. Unfortunately, it lacks use of string resource id in application label. Application name (i.e. `applicaiton:name` tag in `AndroidManifest.xml`) is more appropriate for apk naming.

### DIFF
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,4 +1,4 @@
-# 0.6.0 (2021-05-06)
+# Unreleased (2021-05-06)
 
 - Added support for android string resource id in application label.
 - Added `apk_name` field to android metadata for APK file naming (fall back to artifact name).
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,7 +1,7 @@
-# Unreleased (2021-05-06)
+# Unreleased
 
-- Added support for android string resource id in application label.
-- Added `apk_name` field to android metadata for APK file naming (fall back to artifact name).
+- Added `apk_name` field to android metadata for APK file naming (defaults to Rust library name if unspecified).
+  The application label is now no longer used for this purpose, and can contain a string resource ID from now on.
 
 # 0.6.0 (2021-04-20)
 
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.6.0 (2021-05-06)
+
+- Added support for android string resource id in application label.
+- Added `apk_name` field to android metadata for APK file naming (fall back to artifact name).
+
 # 0.6.0 (2021-04-20)
 
 - **Breaking:** uses `ndk-build`'s new (de)serialized `Manifest` struct to properly serialize a toml's `[package.metadata.android]` to an `AndroidManifest.xml`. The `[package.metadata.android]` now closely resembles the structure of [an android manifest file](https://developer.android.com/guide/topics/manifest/manifest-element). See [README](README.md) for an example of the new `[package.metadata.android]` structure and all manifest attributes that are currently supported.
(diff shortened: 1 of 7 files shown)


ndk-build/src/manifest.rs
```diff
@@ -70,6 +70,9 @@ pub struct Application {
     #[serde(rename(serialize = "android:label"))]
     #[serde(default)]
     pub label: String,
+    #[serde(rename(serialize = "android:name"))]
+    #[serde(default)]
+    pub name: String,
```
MarijnS95, line 75: https://developer.android.com/guide/topics/manifest/application-element

> `android:name`
>
> The fully qualified name of an `Application` subclass implemented for the application. When the application process is started, this class is instantiated before any of the application's components.
>
> The subclass is optional; most applications won't need one. In the absence of a subclass, Android uses an instance of the base Application class.

I don't think that's better. IMO we should add a separate `apk_name` field to:

https://github.com/rust-windowing/android-ndk-rs/blob/7936944edc699d3e7f380cfa87515f8899ce7027/cargo-apk/src/manifest.rs#L7-L13

And fall back to the artifact name as usual.

MarijnS95 (commented):
Besides the unnecessary indentation changes as pointed out by the CI (run `cargo fmt --all` locally :slightly_smiling_face:) this seems okay to me. Don't forget to document this new property in the `cargo-apk` readme, and add a changelog entry to both `cargo-apk` and `ndk-build`.

cargo-apk/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 0.6.0 (2021-05-06)
+
+- Added support for android string resource id in application label.
```
MarijnS95, line 3: That's an implicit result of not using the application name as apk name anymore? Perhaps mention that more explicitly within the next bullet point instead.

cargo-apk/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 0.6.0 (2021-05-06)
```
MarijnS95, line 1: Looks like I've botched the pattern in the previous release. This should be `# Unreleased`.

At some point we'll have to investigate `release.toml` and the automated deploy process to take take care of updating this: https://github.com/sunng87/cargo-release/blob/master/docs/faq.md#maintaining-changelog

ndk-build/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 0.2.0 (2021-05-06)
```
MarijnS95, line 1:
```suggestion
# Unreleased
```

`0.2.0` has already been released.

MarijnS95 (commented):
Seems okay to me besides the changelog.

Would like someone else to take a look, too.

msiglreith (approved):
fine for me as well once the changelog is fixed

MarijnS95:
@xzenodart I've pushed the changelog the way I intended it to be written in the review, let me know if you like it this way.

In particular the change from application label to a separate field is a single change (though with multiple effects) and should only take up a single bullet-point. The release-date hasn't been set in stone yet either, but if you need this we can probably make a new one son.

MarijnS95 (approved):

xzenodart:
> if you need this we can probably publish a patch release soon

Thanks, it will be great!

MarijnS95:
@xzenodart Your change was published in https://github.com/rust-windowing/android-ndk-rs/releases/tag/cargo-apk-0.7.0 :)

PR #147, opened 2021-05-10
Allow using cargo-apk for ci builds.

### LINKED ISSUE #139: allow running cargo apk without building an apk
this is useful for ci. while `cargo build --target {android-target}` finds the linker these days, if you have c dependencies cargo apk is still convenient.

dvc94ch:
Closes #139


MarijnS95:
@dvc94ch What's this supposed to fix? A simple `cargo build` (as per #139) shouldn't invoke `cargo-apk build` at all?

This change however seems to indicate that not library should be added to the APK if the _library_ doesn't exist (but the APK will, as `create_apk` was called). Maybe I do not understand what you're doing in #139?

MarijnS95:
Check. So you're still building an APK and utilizing `cargo-apk` to set up the C compiler, just not zipping a shared library into the APK.

MarijnS95:
Can you clarify that in #139, this issue/title, and the comment in the code?

dvc94ch:
well in the case of a library like serde the apk will be empty. we could add a --no-apk flag, but that didn't seem very pretty either. I guess people should setup an emulator in ci and check that the apk startsup if they're building an app.

cargo-apk/src/apk.rs
```diff
@@ -167,7 +167,11 @@ impl<'a> ApkBuilder<'a> {
                 .map(|path| path.as_path())
                 .collect::<Vec<_>>();
 
-            apk.add_lib_recursively(&artifact, *target, libs_search_paths.as_slice())?;
+            // allow using cargo-apk for building libraries in ci that don't produce a cdylib
```
MarijnS95, line 170: You can probably leave out the `in ci` part or put it in parentheses - I take it this is useful on a host system too.

MarijnS95:
> we could add a --no-apk flag

Why not `cargo-apk check`? That should invoke `build.rs` at least and build native code with a C(++) compiler, no?

MarijnS95:
@dvc94ch If I create a new project but forgot to set the target to `cdylib`, I think this change silently gives me an empty APK back, right?

If that's true we do need a separate command for checking, that should be a hard error.

cargo-apk/src/main.rs
```diff
@@ -9,6 +9,7 @@ fn main() -> anyhow::Result<()> {
     let builder = ApkBuilder::from_subcommand(&cmd)?;
 
     match cmd.cmd() {
+        "check" => builder.check()?,
         "build" => {
             for artifact in cmd.artifacts() {
```
MarijnS95, line 14: Seeing that `build` handles artifacts, should `check` also do something like that so that `-p` and `--example` work, or is that implicitly working already when `self.cmd.args()` is passed to `cargo check`?

MarijnS95 (commented):
Yeah that's much better!

Perhaps you want to add this to the changelog too, and maybe document that it is useful to compile-test crates with C(++) dependencies using the NDK without requiring the final crate to be a `cdylib` that's turned into an APK (your usecase in #139)?

cargo-apk/src/apk.rs
```diff
@@ -86,6 +86,28 @@ impl<'a> ApkBuilder<'a> {
...
+    pub fn check(&self) -> Result<(), Error> {
+        for target in &self.build_targets {
+            let triple = target.rust_triple();
+            let target_sdk_version = self
+                .manifest
+                .android_manifest
+                .sdk
+                .target_sdk_version
+                .unwrap();
+            let mut cargo = cargo_apk(&self.ndk, *target, target_sdk_version)?;
```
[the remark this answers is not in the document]
dvc94ch: that would be a breaking change no? maybe leave it for another PR?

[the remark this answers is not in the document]
MarijnS95: I think the next version can be a patch release, only a field was added to the `ndk_build::ApkConfig` struct.

Regardless would be nice to fix the name and make it somewhat representative :D

cargo-apk/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+- Added `cargo apk check`. Useful for compile-testing crates that contain C/C++ dependencies or
+  target specific conditional compilation.
```
MarijnS95, line 4:
```suggestion
  target-specific conditional compilation.
```

MarijnS95 (approved):
@dvc94ch Yes please, thanks!

cargo-apk/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+- Added `cargo apk check`. Useful for compile-testing crates that contain C/C++ dependencies or
+  target-specific conditional compilation, but do not provide a cdylib target.
+
```
MarijnS95, line 5: Oh, in hindsight I don't think we ever had newlines between bulletpoints in the changelog. Feel free to merge this and make a new release after dropping this one :)

PR #152, opened 2021-07-05
release cargo-apk 0.8 and ndk-build 0.4

msiglreith:
Requested in https://github.com/rust-windowing/android-ndk-rs/pull/138#issuecomment-873394048

### DIFF
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Unreleased
+# 0.8.0 (2021-07-06)
 
 - Added `runtime_libs` path to android metadata for packaging extra dynamic libraries into the apk.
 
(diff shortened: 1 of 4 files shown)


MarijnS95:
@msiglreith It looks like this is missing the `version =` bumps themselves?

msiglreith:
oh..🤦
thanks!

cargo-apk/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# Unreleased
```
MarijnS95, line 1: Might want to keep this around for the next release... I accidentally removed it in one of the previous releases too and it botched some rebases.

MarijnS95 (approved):

MarijnS95:
Shouldn't this be a patch bump btw? We're still on `0.` and no breaking changes were made, last I remember.

msiglreith:
thanks for reviews!
yes probably a patch bump would have been enough as well

PR #157, opened 2021-07-22
ndk-build: Only use supported platforms as default target SDK

### LINKED ISSUE #155: ndk directory structure changed?
Started getting this ci failure. Looks like google released a new ndk with a changed directory structure. Or github-actions changed something.

Error: Path `"/usr/local/lib/android/sdk/ndk-bundle/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android31-clang"` doesn't exist.

MarijnS95:
Use `build/core/platforms.mk` to filter out installed platforms (`$SDK/platforms/android-<version>`) that are not actually supported by the selected NDK.  This list of platforms is currently only used to determine a sensible (highest) default target SDK.

Fixes #155

### DIFF
--- a/ndk-build/src/ndk.rs
+++ b/ndk-build/src/ndk.rs
@@ -1,5 +1,6 @@
 use crate::error::NdkError;
 use crate::target::Target;
+use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
@@ -53,6 +54,19 @@ impl Ndk {
             .max()
             .ok_or(NdkError::BuildToolsNotFound)?;
 
+        let ndk_platforms = std::fs::read_to_string(ndk_path.join("build/core/platforms.mk"))?;
+        let ndk_platforms = ndk_platforms
+            .split('\n')
+            .map(|s| s.split_once(" := ").unwrap())
+            .collect::<HashMap<_, _>>();
+
+        let min_platform_level = ndk_platforms["NDK_MIN_PLATFORM_LEVEL"]
+            .parse::<u32>()
+            .unwrap();
+        let max_platform_level = ndk_platforms["NDK_MAX_PLATFORM_LEVEL"]
+            .parse::<u32>()
+            .unwrap();
+
         let platforms_dir = sdk_path.join("platforms");
         let platforms: Vec<u32> = std::fs::read_dir(&platforms_dir)
             .or(Err(NdkError::PathNotFound(platforms_dir)))?
@@ -63,6 +77,7 @@ impl Ndk {
                 name.strip_prefix("android-")
                     .and_then(|api| api.parse::<u32>().ok())
             })
+            .filter(|level| (min_platform_level..=max_platform_level).contains(level))
             .collect();
 
         if platforms.is_empty() {


ndk-build/src/ndk.rs
```diff
@@ -113,8 +114,16 @@ impl Ndk {
...
     pub fn default_platform(&self) -> u32 {
-        self.platforms().iter().max().cloned().unwrap()
+        self.platforms()
+            .iter()
+            .cloned()
+            .rev()
+            // Detect NDK support by existence of the clang compiler, with `android<sdk version>` in the name
+            // TODO: Don't assume some architecture, despite the NDK supporting them all at once
+            .find(|&p| self.clang(Target::Arm64V8a, p).is_ok())
+            .unwrap()
```
MarijnS95, line 126: In hindsight I think this should be moved back to `fn from_env()` to filter everything going into `Self::platforms`, but we can't call `Self::clang()` there yet.

dvc94ch: Not really that familiar with the code anymore. But looks like `from_env()` would be the place to put it.

MarijnS95: Yeah, I'm fairly certain that's the way forward. However, we still have the issue with targets. Maybe iterate them all, or just assume every NDK supports `Arm64V8a` for the time being and call it a day? Otherwise we might have to parse some config file to find out what android versions are supported - or base `fn default_platform()` on the selected target triple.

dvc94ch: how about parsing `build/core/platforms.mk`? I did some grepping and this seems to be the best way of figuring out the supported platforms.

```make
cat build/core/platforms.mk
NDK_MIN_PLATFORM_LEVEL := 16
NDK_MAX_PLATFORM_LEVEL := 30
NDK_PLATFORM_ALIAS_20 := android-19
NDK_PLATFORM_ALIAS_25 := android-24
NDK_PLATFORM_ALIAS_J := android-16
NDK_PLATFORM_ALIAS_J-MR1 := android-17
NDK_PLATFORM_ALIAS_J-MR2 := android-18
NDK_PLATFORM_ALIAS_K := android-19
NDK_PLATFORM_ALIAS_L := android-21
NDK_PLATFORM_ALIAS_L-MR1 := android-22
NDK_PLATFORM_ALIAS_M := android-23
NDK_PLATFORM_ALIAS_N := android-24
NDK_PLATFORM_ALIAS_N-MR1 := android-24
NDK_PLATFORM_ALIAS_O := android-26
NDK_PLATFORM_ALIAS_O-MR1 := android-27
NDK_PLATFORM_ALIAS_P := android-28
NDK_PLATFORM_ALIAS_Q := android-29
NDK_PLATFORM_ALIAS_R := android-30
```

MarijnS95: Not really fond of parsing makefiles, and we'd have to check how far back this method works if we wish to maintain backwards compatibility of some sort, too.

MarijnS95: @dvc94ch Done that now, works quite nicely for at least the [currently supported LTS and Rolling NDK releases](https://github.com/android/ndk/wiki#supported-downloads). Will follow up with more test results and a PR documenting supported NDKs shortly, but consider this PR done and ready for review :)

dvc94ch (approved):
nice, thanks!

MarijnS95:
@dvc94ch Thanks for review! I have a few more PRs to make (but am not close to my main PC), perhaps we can work through that tomorrow and publish some new releases with these :)

PR #158, opened 2021-07-23
looper: Model `ALOOPER_EVENT` with `bitflags`

MarijnS95:
The `event` field/parameter cannot have more states than these few carefully defined constants, and should be exposed as a proper restricted type since it's part of the public API.


dvc94ch (approved):
lgtm

MarijnS95:
@dvc94ch Thanks!

Issue #149, opened 2021-05-20
Android NDK r23 beta 3 and up do not include `libgcc` anymore


MarijnS95:
A heads-up to anyone upgrading _beyond_ NDK r23 beta 2:

```
error: linking with `~/Android/Sdk/ndk/23.0.7272597/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android30-clang` failed: exit status: 1
  |
  = note: "~/Android/Sdk/ndk/23.0.7272597/toolchains/llvm/prebuilt/linux-x86_64/bin/aarch64-linux-android30-clang" "-Wl,--allow-multiple-definition" "-Wl,--eh-frame-hdr" "-Wl,-znoexecstack" "-Wl,--as-needed" "-L" "~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/aarch64-linux-android/lib" ... "-Wl,--gc-sections" "-shared" "-Wl,-zrelro" "-Wl,-znow" "-nodefaultlibs" "-L" "android-ndk-rs/target/aarch64-linux-android/debug/deps" "-L" "android-ndk-rs/target/debug/deps" "-L" "~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/aarch64-linux-android/lib" "-Wl,-Bstatic" ... "-Wl,-Bdynamic" "-landroid" "-ldl" "-llog" "-lgcc" "-lc" "-lm"
  = note: ld: error: unable to find library -lgcc
          clang-12: error: linker command failed with exit code 1 (use -v to see invocation)


error: aborting due to previous error

error: could not compile `ndk-examples`
```

r23 beta 2 is still working fine as per #137.

GCC Is being completely removed from the NDK, but unfortunately something deep down in Rust depends on `gcc`. This also happens on a project without dependencies built with `cargo b --target aarch64-linux-android`. 

The possible cause for these is:
https://github.com/rust-lang/rust/blob/9a3214e9be41a5d50ae6ba9bb8422dcc2cb10473/library/std/src/sys/unix/mod.rs#L210-L214
I'll see whether Rust really needs to link against `libgcc` and if that can otherwise be replaced with `libclang_rt.builtins` as per https://github.com/android/ndk/wiki/Changelog-r23#changes / https://github.com/android/ndk/issues/1231, without breaking existing builds.

MarijnS95:
[...]

PR #159, opened 2021-07-24
README: Document tested and supported NDK versions

MarijnS95:
Not sure if we want this as high up on the `README` though, but it should be helpful for us to keep track and for crate users to know what to expect (ie. that r23 isn't going to work until a newer Rust release that doesn't depend on `libgcc` anymore).

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ r20 | 20.1.5948944 | _Deprecated_ | :heavy_check_mark:
 r21 | 21.4.7075529 | LTS | :heavy_check_mark:
 r22 | 22.1.7171670 | Rolling Release | :heavy_check_mark:
 r23 beta 1/2 |  | Beta | :heavy_check_mark:
-r23 beta 3 and beyond |  | Beta | :x: Breaking on https://github.com/rust-windowing/android-ndk-rs/issues/149 :x:
+r23 beta 3 and beyond |  | Beta | :x: Breaking on [#149](https://github.com/rust-windowing/android-ndk-rs/issues/149) :x:
 
 ## Hello world
 
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-# Rust on Android
+Rust on Android
+=
 
 [![Rust](https://github.com/rust-windowing/android-ndk-rs/workflows/Rust/badge.svg)](https://github.com/rust-windowing/android-ndk-rs/actions) ![MIT license](https://img.shields.io/badge/License-MIT-green.svg) ![APACHE2 license](https://img.shields.io/badge/License-APACHE2-green.svg)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,6 +15,22 @@ Name | Description | Badges
...
+
+Branch | Version | Status | Working
+-|-|:-:|:-:
+r18 | 18.1.5063045 | _Deprecated_ | :x:
+r19 | 19.2.5345600 | _Deprecated_ | :heavy_check_mark:
+r20 | 20.1.5948944 | _Deprecated_ | :heavy_check_mark:
+r21 | 21.4.7075529 | LTS | :heavy_check_mark:
+r22 | 22.1.7171670 | Rolling Release | :heavy_check_mark:
+r23 beta 1/2 |  | Beta | :heavy_check_mark:
+r23 beta 3 and beyond |  | Beta | :x: Breaking on #149 :x:
```
dvc94ch, line 32: the link to #149 doesn't work. curious, why doesn't r18 work? I guess we can remove the empty winit/glutin and flutter sections too.

MarijnS95: > the link to #149 doesn't work.

Ah that's a bummer, it's a hardlink now.

> curious, why doesn't r18 work? 

It doesn't have `<triple>-android<sdk>-clang(++)` binaries, just `clang`/`clang++`. I don't know if `clang` is supported in these NDKs at all or if that's just a host compiler and GCC is supposed to be used. Someone could test and implement fallbacks if they really want to use this ancient one. Either way this seems to be the most convenient way to display that `r19` is the first actually supported and working NDK.

> I guess we can remove the empty winit/glutin and flutter sections too.

They're gone now.

dvc94ch: > Someone could test and implement fallbacks if they really want to use this ancient one.

Nah, was just curious since you appeared to have gone through the trouble of actually testing it. I don't think anyone cares if r18 doesn't work. that's a release from 2018 right?

README.md
```diff
@@ -14,6 +15,22 @@ Name | Description | Badges
...
+
+Branch | Version | Status | Working
+-|-|:-:|:-:
+r18 | 18.1.5063045 | _Deprecated_ | :x:
+r19 | 19.2.5345600 | _Deprecated_ | :heavy_check_mark:
+r20 | 20.1.5948944 | _Deprecated_ | :heavy_check_mark:
+r21 | 21.4.7075529 | LTS | :heavy_check_mark:
+r22 | 22.1.7171670 | Rolling Release | :heavy_check_mark:
+r23 beta 1/2 |  | Beta | :heavy_check_mark:
+r23 beta 3 and beyond |  | Beta | :x: Breaking on https://github.com/rust-windowing/android-ndk-rs/issues/149 :x:
```
dvc94ch, line 32:
```suggestion
r23 beta 3 and beyond |  | Beta | :x: Breaking on [#149](https://github.com/rust-windowing/android-ndk-rs/issues/149) :x:
```

dvc94ch (approved):
Thanks for testing this!

README.md
MarijnS95: Wasn't too hard to select a bunch of NDKs in the SDK manager and run through them with `ANDROID_NDK_ROOT=<the version> cargo apk run <some example>` :)

https://developer.android.com/ndk/downloads/revision_history September 2018 indeed!

PR #161, opened 2021-07-26
ndk-example: Demonstrate Looper usage without winit

### LINKED ISSUE #111: ndk-glue: impossible to poll events in a non-blocking fashion
it's impossible to poll events in ndk-glue without blocking the thread. In the NDK samples I have seen ALooper is used to poll events with a timeout value that can be configured to block or not. Where the ndk-glue module will read the file descriptor directly with a blocking call. This makes it impossible to program games, or any application that wants to keep running without  having to block for events/input.

MarijnS95:
Fixes #111

Originally proposed [here], this adds a simple looper example demonstrating how to use Android's `Looper` through the NDK bindings in similar fashion to `winit`.  It uses `ndk_glue`'s event pipe, Android's `InputQueue` and a custom Unix pipe to transfer events to the Looper (thread).

[here]: https://github.com/rust-windowing/android-ndk-rs/issues/111#issuecomment-768118616

### DIFF
--- a/ndk-examples/examples/looper.rs
+++ b/ndk-examples/examples/looper.rs
@@ -8,6 +8,8 @@ use log::info;
 use ndk::event::{InputEvent, Keycode};
 use ndk::looper::{FdEvent, Poll, ThreadLooper};
 
+const U32_SIZE: usize = std::mem::size_of::<u32>();
+
 #[cfg_attr(
     target_os = "android",
     ndk_glue::main(backtrace = "on", logger(level = "debug"))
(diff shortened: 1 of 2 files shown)


ndk-examples/examples/looper.rs
```diff
@@ -0,0 +1,136 @@
...
+                        info!("Read custom event from pipe: {}", recv);
+                    }
+                    i => panic!("Unexpected event identifier {}", i),
+                }
+            }
+        }
+    }
+
+    // Stop the activity
+    ndk_glue::native_activity().finish()
```
MarijnS95, line 135: This is still pending #160; but we can get this PR in already and it will just have to be cleaned up there (and it should probably fail to compile no matter how we implement it).

msiglreith (approved):
that's a nice reference!

MarijnS95:
@msiglreith Thanks for reviewing! Wanted to hold off merging until finalizing all the changes to `add_fd_with_callback` so that it can be demonstrated/tested as well, that's all here now too.

msiglreith (approved):

PR #162, opened 2021-07-26
Simplify some code in `ndk-macro` and `ndk-build`

MarijnS95:


ndk-macro/src/helper.rs
```diff
@@ -49,16 +49,11 @@ fn crate_name(name: &str) -> Result<String> {
...
     // try to use overridden crate path
     overridden_path.clone().unwrap_or_else(|| {
-        // the binding to hold string from `crate_name` fn
-        let mut detected_name = None;
         Ident::new(
             // try to determine crate name from Cargo.toml
             crate_name(name)
                 .ok()
-                .map(|name| {
-                    detected_name = Some(name);
```
dvc94ch, line 59: what's this :laughing:

dvc94ch (approved):

ndk-macro/src/helper.rs
MarijnS95: :joy: I think `.map(|name| name.as_str())` was tried first which moves the `String` from the `Option<>` into the closure, resulting in obvious lifetime issues when calling `.as_str()`. The way to solve that is to use `.as_ref().map(|name| ...)`, and in this case an even simpler `.as_deref()` because `String`'s `Deref` is implemented to target `str` :grimacing: (since the `String` stays alive during this entire statement).

But you knew that already, just writing it here for others who might stumble across :grinning:

ndk-build/src/apk.rs
```diff
@@ -79,17 +79,13 @@ impl<'a> UnalignedApk<'a> {
...
+        std::fs::create_dir_all(out.parent().unwrap())?;
+        std::fs::copy(path, out)?;
 
         let mut aapt = self.0.build_tool(bin!("aapt"))?;
-        aapt.arg("add").arg(self.0.unaligned_apk()).arg(format!(
-            "lib/{}/{}",
-            abi,
-            file_name.to_str().unwrap()
-        ));
+        aapt.arg("add").arg(self.0.unaligned_apk()).arg(&lib_path);
```
msiglreith, line 88: Unfortunately, this results in issues when building on Windows..
Using a backslash path for the library results in a runtime error where it fails to load the .so file.
```
prev: "lib/arm64-v8a/libhello_world.so"
new: "lib\\arm64-v8a\\libhello_world.so"
```

This only applies to the `lib_path`, for the `unaligned_apk` path it doesn't matter

MarijnS95: @msiglreith Ouch, small cleanups always result in horrible fallout. We need to resurrect our Windows CI (if this was ever tested on Windows at all). We've seen this before and I'm thinking of fixing it just like https://github.com/dvc94ch/cargo-subcommand/pull/1, will also take a look at the CI. Thanks for mentioning!

MarijnS95: https://github.com/rust-windowing/android-ndk-rs/runs/3165210588 So the CI runs fine, but the backslashes are now part of a full filename instead of serving as path separators.

msiglreith: Hmm, this would need to run inside the emulator I guess. As the runtime crash is on the target device not during building.

Example:
```
AndroidRuntime: java.lang.RuntimeException: Unable to start activity ComponentInfo{rust.wgpu_android/android.app.NativeActivity}: java.lang.IllegalArgumentException: Unable to find native library wgpu_android using classloader: dalvik.system.PathClassLoader[DexPathList[[],nativeLibraryDirectories=[/data/app/rust.wgpu_android-pSe8ljSIP942rgeN5FSlQQ==/lib/arm64, /system/lib64, /system/product/lib64]]]
```

MarijnS95: Correct, I think that's a good idea. Can we make this use artifacts from `cargo apk build`? That finally gives us a nice dependency graph in the Actions run overview, though I don't know if those can be shared between yml "files", which seem to be rendered as separate graphs (apparently called "runs"). We can just combine the two jobs in one yml file/run.

msiglreith: Yeah, workflow sharing is not possible afaik - we would need to define it as action and then transfer the artifacts inside the same workflow, but I never tried that :D

PR #163, opened 2021-07-26
ndk-macro: Update proc-macro-crate to 1.0, update darling to 0.13

MarijnS95:
Depends on #162 

proc-macro-crate 1.0 released, and [num_enum] bumped to that recently in a patch release causing a duplicate dependency in `Cargo.lock`.  We can upgrade to it too.
    
[num_enum]: https://github.com/illicitonion/num_enum/commit/8385dc995b6a796b02e072f02e9019f15b334791

### DIFF
--- a/ndk-macro/src/helper.rs
+++ b/ndk-macro/src/helper.rs
@@ -1,5 +1,6 @@
 use core::ops::Deref;
 use proc_macro2::{Ident, Span};
+use proc_macro_crate::FoundCrate;
 use syn::{
     parse::{Parse, ParseStream, Result},
     Path, Token,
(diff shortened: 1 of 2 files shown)


ndk-macro/src/helper.rs
```diff
@@ -42,22 +43,21 @@ impl Parse for AttributeArgs {
...
         Ident::new(
             // try to determine crate name from Cargo.toml
             crate_name(name)
                 .ok()
-                .map(|name| {
-                    detected_name = Some(name);
-                    detected_name.as_ref().unwrap().as_str()
+                .as_ref()
+                .map(|name| match name {
+                    FoundCrate::Itself => todo!(),
```
MarijnS95, line 59: This should probably be `"ndk_macro"`?

dvc94ch: looks like it

dvc94ch (approved):

ndk-macro/src/helper.rs
MarijnS95: Let's get that in then and merge this :)

PR #165, opened 2021-07-29
ndk-glue: Add missing entries to changelog and perform patch release 0.3.1

### LINKED ISSUE #164: New release with threading fixes?
Would it be possible to make a new release on crates.io (and then update winit)? The recent threading fixes in ndk_glue seem like very important improvements. If it would help to have a PR that bumps the versions then I can create one.

MarijnS95:
Fixes #164 

It seems these threading fixes went unnoticed for a while as they didn't have anything added to `CHANGELOG.md`. Still, they're both vital in preventing deadlocks and race-conditions.

I don't think any of these are breaking and warrant a minor release, so a patch should be fine to get it in everyones hands smoothly, before the next breaking changes (see open PRs) land.

### DIFF
--- a/ndk-glue/CHANGELOG.md
+++ b/ndk-glue/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Unreleased
 
+- Looper is now created before returning from `ANativeActivity_onCreate`, solving
+  race conditions in `onInputQueueCreated`.
+- Event pipe and looper are now notified of removal _before_ destroying `NativeWindow`
+  and `InputQueue`. This allows applications to unlock their read-locks of these instances
+  first (which they are supposed to hold on to during use) instead of deadlocking in
+  Android callbacks.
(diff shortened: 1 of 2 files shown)


dvc94ch (approved):

msiglreith (approved):

MarijnS95:
@msiglreith @dvc94ch I just realized I made a total rookie mistake:

> - Use new `FdEvents` `bitflags` for looper file descriptor events.

This patch definitely depends on an `ndk` update. Fortunately `cargo publish` totally caught it and failed to publish: https://github.com/rust-windowing/android-ndk-rs/runs/3194488659

I'll push a minor update for the `ndk` shortly. Should we give ndk-glue a minor bump as well if bumping one of these? Duplicate dependencies can be a hassle.

Additionally I'll see if we can make the CI fail properly instead of making this go unnoticed.

PR #166, opened 2021-07-30
Release ndk 0.4.0, ndk-glue 0.4.0, ndk-build 0.4.1

### LINKED ISSUE #164: New release with threading fixes?

MarijnS95:
Fixes #164

~ndk-glue 0.3.1 depends on a breaking change in the NDK, and can only be released if that's released under 0.4.0.~

---

~Draft: I'm not sure if this is the right thing to do, or if we should aim for ndk-glue `0.4.0`. ndk-glue exposes quite a few NDK symbols in its public API, and running `cargo update` in a crate that uses both (very likely) they'll end up with duplicate crates and incompatible types unless they manually update to ndk 0.4 (which is possibly, but unexpected sudden breakage because semver allows going from 0.3.0 to 0.3.1 without issues).~

---

ndk-glue depends on a breaking change in the NDK, and can only be released if that's released under 0.4.0.  This, in turn, causes ndk-glue to need a minor bump too despite no other breaking changes, as NDK symbols are used in its public API.

Note that ndk-glue 0.3.1 was never properly released as the ndk update requirement outlined above caused an adequate publishing failure.

### DIFF
--- a/ndk-build/CHANGELOG.md
+++ b/ndk-build/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+# 0.4.1 (2021-08-02)
+
+- Only the highest platform supported by the NDK is now selected as default platform.
+
 # 0.4.0 (2021-07-06)
 
 - Added `add_runtime_libs` function for including extra dynamic libraries in the APK.
--- a/ndk-build/Cargo.toml
+++ b/ndk-build/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ndk-build"
-version = "0.4.0"
+version = "0.4.1"
 authors = ["The Rust Windowing contributors"]
 edition = "2018"
 description = "Utilities for building Android binaries"
--- a/ndk-glue/Cargo.toml
+++ b/ndk-glue/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ndk-glue"
-version = "0.3.1"
+version = "0.4.0"
 authors = ["The Rust Windowing contributors"]
 edition = "2018"
 description = "Startup code for android binaries"
@@ -12,7 +12,7 @@ homepage = "https://github.com/rust-windowing/android-ndk-rs"
 repository = "https://github.com/rust-windowing/android-ndk-rs"
 
 [dependencies]
-ndk = { path = "../ndk", version = "0.3.0" }
+ndk = { path = "../ndk", version = "0.4.0" }
 ndk-sys = { path = "../ndk-sys", version = "0.2.1" }
 ndk-macro = { path = "../ndk-macro", version = "0.2.0" }
 lazy_static = "1.4.0"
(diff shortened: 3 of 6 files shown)


dvc94ch (approved):
Can we get a new cargo-apk version too? would love getting the ci fixed too while we're making releases

MarijnS95:
@dvc94ch it looks like the last change to the `cargo-apk` was its `0.8.0` release; anything you're missing in particular? What CI breakage are you referring to?

dvc94ch:
I guess ndk-build is enough if it's only a patch release. Looking for this commit:

https://github.com/rust-windowing/android-ndk-rs/commit/d21e51fb692cf778b921b747b2eaa85e5344eb71

MarijnS95:
@dvc94ch Right, yes, a patch release should be enough!

ndk-build/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+# 0.4.1 (2021-TODO)
```
MarijnS95, line 3: We'll have to fix these dates when it's time to merge this.

MarijnS95:
@dvc94ch Any comment on whether `ndk-glue` needs to perform a minor instead of patch release, when relying on a minor release of the `ndk`?

dvc94ch:
it's the safe thing to do. generally I release minor versions whenever I upgrade a dependency. there may be cases were it isn't strictly necessary. but in this case we are leaking types from `ndk` in `fn native_activity()` for example, so it is necessary.

MarijnS95:
@dvc94ch Yep, thanks for confirming. Should all be set now, would you mind reviewing once more before we merge this?

@iamralpht I'm not sure if we can fully fix #164 as `winit` takes quite some time to release new versions, unless we set up the proper infrastructure to perform non-breaking backport patch releases and actually get `ndk-glue 0.3.1` out the door without depending on `ndk 0.4.0`.

ndk-build/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ndk-build"
-version = "0.4.0"
+version = "0.4.1"
```
dvc94ch, line 3: checked, a patch is fine

ndk-glue/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ndk-glue"
-version = "0.3.1"
+version = "0.4.0"
```
dvc94ch, line 3: this is safe in any case and we already discussed it

ndk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ndk"
-version = "0.3.0"
+version = "0.4.0"
```
dvc94ch, line 3: this is clear from the changelog

dvc94ch (approved):
thanks!

MarijnS95:
@dvc94ch Thanks for checking! Let's get this merged now, I'll be able to monitor the crate during the day in case something happens after all.

PR #168, opened 2021-08-05
ndk-build: fix native activity library path on non-UNIX systems

msiglreith:
On Windows, using a backslash path for the library results in a runtime error where it fails to load the .so file.
```
prev: "lib/arm64-v8a/libhello_world.so"
new: "lib\\arm64-v8a\\libhello_world.so"
```

### DIFF
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fix NativeActivity library path on non-UNIX systems, which resulted in runtime crashes on startup.
+
 # 0.8.0 (2021-07-06)
 
 - Added `runtime_libs` path to android metadata for packaging extra dynamic libraries into the apk.
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- Fix NativeActivity library path on non-UNIX systems, which resulted in runtime crashes on startup.
+- Pass UNIX path separators to `aapt` on non-UNIX systems, ensuring the resulting separator is compatible with the target device instead of the host platform.
 
 # 0.8.0 (2021-07-06)
 
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Unreleased
 
-# 0.8.1 (2021-08-05)
+# 0.8.1 (2021-08-06)
 
-- Updated to use [ndk-build 0.4.2](../ndk-build/CHANGELOG.md#042-2021-08-05)
+- Updated to use [ndk-build 0.4.2](../ndk-build/CHANGELOG.md#042-2021-08-06)
 
 # 0.8.0 (2021-07-06)
 
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Unreleased
 
-- Pass UNIX path separators to `aapt` on non-UNIX systems, ensuring the resulting separator is compatible with the target device instead of the host platform.
+# 0.8.1 (2021-08-05)
+
+- Updated to use [ndk-build 0.4.2](../ndk-build/CHANGELOG.md#042-2021-08-05)
 
 # 0.8.0 (2021-07-06)
 
--- a/ndk-build/CHANGELOG.md
+++ b/ndk-build/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- Fix NativeActivity library path on non-UNIX systems, which resulted in runtime crashes on startup.
+- Pass UNIX path separators to `aapt` on non-UNIX systems, ensuring the resulting separator is compatible with the target device instead of the host platform.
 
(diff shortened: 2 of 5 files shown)


MarijnS95:
Heh, I was just writing a reply to https://github.com/rust-windowing/android-ndk-rs/pull/162#discussion_r683788005, as we've seen the inverse of this in https://github.com/dvc94ch/cargo-subcommand/pull/1. In short, I don't like `format!` for paths nor doing "the same" string join/format operations twice at risk of them getting out of sync one day. Perhaps we can just fix this by calling `.replace('\\', '/')`?

msiglreith:
Sure, I'm fine with that as well!

cargo-apk/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fix NativeActivity library path on non-UNIX systems, which resulted in runtime crashes on startup.
```
MarijnS95, line 3: This only applies to `cargo-apk`, right? I don't think we need to perform any minor/major dependency bump here?

ndk-build/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fix NativeActivity library path on non-UNIX systems, which resulted in runtime crashes on startup.
```
MarijnS95, line 3: Should probably make this a bit more clear than `resulted in runtime crashes on startup`. Something along the lines of: ``Pass UNIX path separators to `aapt` on non-UNIX systems, ensuring the resulting separator is treated as separator on Android instead of being part of the filename.``.

Not sure how to better word `being part of the filename` though, but there should be something more clear.

cargo-apk/CHANGELOG.md
msiglreith: Patch releases for `cargo-apk` and `ndk-build` would be good and `^` enforcing the `ndk-build` patch

MarijnS95: Feel free to do that :) - maybe mention just the version upgrade here and refer to the `ndk-build` changelog or adopt the line below?

Btw caret `^` semver matching is the default, meaning that least-significant (patch for 0.x.y, or minor for x.y.z) versions may always be bumped. But it's good to turn the patch into a `.1` to clarify that at least that version has to be used.

ndk-build/src/apk.rs
```diff
@@ -84,8 +84,14 @@ impl<'a> UnalignedApk<'a> {
         std::fs::create_dir_all(out.parent().unwrap())?;
         std::fs::copy(path, out)?;
 
+        // aapt requires UNIX convention for the library path.
+        // Otherwise, it results in a runtime error when loading the NativeActivity .so library.
```
MarijnS95, line 88: Can you also apply the changelog wording here?

MarijnS95 (approved):

ndk-build/src/apk.rs
```diff
@@ -84,8 +84,15 @@ impl<'a> UnalignedApk<'a> {
         std::fs::create_dir_all(out.parent().unwrap())?;
         std::fs::copy(path, out)?;
 
+        // Pass UNIX path separators to `aapt` on non-UNIX systems, ensuring the resulting separator
+        // is compatible with the target device instead of the host platform.
+        // Otherwise, it results in a runtime error when loading the NativeActivity `.so` library.
```
MarijnS95, line 89: I guess this is just a ``"failed to find/load `lib/<target arch>/lib<activityname>.so, -ENOENT`"`` kind of error?

msiglreith: yes, but not really clear imo why it fails:

AndroidRuntime: java.lang.RuntimeException: Unable to start activity ComponentInfo{rust.wgpu_android/android.app.NativeActivity}: java.lang.IllegalArgumentException: Unable to find native library wgpu_android using classloader: dalvik.system.PathClassLoader[DexPathList[[],nativeLibraryDirectories=[/data/app/rust.wgpu_android-pSe8ljSIP942rgeN5FSlQQ==/lib/arm64, /system/lib64, /system/product/lib64]]]

msiglreith:
Thanks for review!

ndk-build/src/apk.rs
MarijnS95: (That's exactly the kind of context we should paste in PR's / commit messages from the beginning :wink::wink:. Both for reviewer-understanding as well as users running into this issue to have a higher chance finding it with the search bar)

Seems pretty clear to me: try running `touch foo\\bar\\baz` on a UNIX system. This creates a file with that name, instead of the directory structure. Same happens here, see https://github.com/rust-windowing/android-ndk-rs/runs/3165210588:

```
Verifying alignment of D:\a\android-ndk-rs\android-ndk-rs\target\debug\apk\examples\hello_world.apk (4)...
      49 AndroidManifest.xml (OK - compressed)
     880 lib\arm64-v8a\libhello_world.so (OK - compressed)
```

`lib\arm64-v8a\libhello_world.so` is a filename in the root of the zip archive - it's not a directory structure. This doesn't get unpacked to `/data/app/rust.wgpu_android-pSe8ljSIP942rgeN5FSlQQ==/lib/arm64` so that the class loader finally fails to find the library.

PR #170, opened 2021-08-07
Add missing safety docs

MarijnS95:
Filing this as a separate PR in preparation for `clippy` linting in the CI, just so that these nontrivial documentation changes can be reviewed properly.

### DIFF
--- a/ndk/src/native_activity.rs
+++ b/ndk/src/native_activity.rs
@@ -24,6 +24,7 @@ unsafe impl Sync for NativeActivity {}
 impl NativeActivity {
     /// Create a `NativeActivity` from a pointer
     ///
+    /// # Safety
     /// By calling this function, you assert that it is a valid pointer to a native
     /// `ANativeActivity`.
     pub unsafe fn from_ptr(ptr: NonNull<ffi::ANativeActivity>) -> Self {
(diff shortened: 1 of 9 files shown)


ndk/src/native_activity.rs
```diff
@@ -120,6 +122,9 @@ impl NativeActivity {
     /// Path to the directory with the application's OBB files.
     ///
     /// Only available as of Honeycomb (Android 3.0+, API level 11+)
+    ///
+    /// # Safety
+    /// `obbPath` must be a pointer to a valid C-string.
```
msiglreith, line 127: I'm wondering it it would make sense to drop the `unsafe` decoration on this one, similar to `external_data_path`.
`obbPath` is an implementation details so not that helpful for the user imo.

MarijnS95: Yeah I agree, it's a bit weird to demand an application to guarantee safety of internal pointers. Perhaps it was done this way because of `Only available as of Honeycomb (Android 3.0+, API level 11+)`? If so, `# Safety` should simply be added above this line, relying on the user to guarantee API 11+ through other means instead.

msiglreith (approved):

PR #172, opened 2021-08-07
ndk/looper: Drop boxed callback after deregistering

MarijnS95:
The box created in `add_fd_with_callback` is turned into a raw pointer but its memory is never dropped when the callback requests for deregistration.

Note that it is still possible to deregister this callback using `ALooper_removeFd`, where the box cannot be accessed and would thus leak.  This is likely not a problem in practice because that function is not exposed in the safe API at the time of writing.
However, it is still possible to replace the callback by calling `ALooper_addFd` on the same `fd` again, leaking the boxes.

---

As an alternative we can write this in a (probably?) cleaner way by always converting back to a `Box` and instead preventing it from being dropped:

```rust
let mut cb = Box::<Box<dyn FnMut(RawFd) -> bool>>::from_raw(data as *mut _);
let keep_registered = cb(fd);
if keep_registered {
    forget(cb);
}
keep_registered as i32
```
Or, perhaps safer, wrap it in `ManuallyDrop` and explicitly drop it in `!keep_registered`.

### DIFF
--- a/ndk/src/looper.rs
+++ b/ndk/src/looper.rs
@@ -9,6 +9,7 @@
 
 use bitflags::bitflags;
 use std::convert::TryInto;
+use std::mem::ManuallyDrop;
 use std::os::raw::c_void;
 use std::os::unix::io::RawFd;
 use std::ptr;
@@ -273,8 +274,14 @@ impl ForeignLooper {
     ) -> Result<(), LooperError> {
         extern "C" fn cb_handler(fd: RawFd, _events: i32, data: *mut c_void) -> i32 {
             unsafe {
-                let cb: &mut Box<dyn FnMut(RawFd) -> bool> = &mut *(data as *mut _);
-                cb(fd) as i32
+                let mut cb = ManuallyDrop::new(Box::<Box<dyn FnMut(RawFd) -> bool>>::from_raw(
+                    data as *mut _,
+                ));
+                let keep_registered = cb(fd);
+                if !keep_registered {
+                    let _ = ManuallyDrop::into_inner(cb);
+                }
+                keep_registered as i32
             }
         }
         let data: *mut c_void = Box::into_raw(Box::new(callback)) as *mut _;
--- a/ndk/src/looper.rs
+++ b/ndk/src/looper.rs
@@ -265,33 +265,35 @@ impl ForeignLooper {
     ///
     /// See also [the NDK
     /// docs](https://developer.android.com/ndk/reference/group/looper.html#alooper_addfd).
-    pub fn add_fd_with_callback(
+    pub fn add_fd_with_callback<F: FnMut(RawFd) -> bool>(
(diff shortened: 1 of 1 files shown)


MarijnS95:
We only have this `Box<dyn Fn...>` pattern in two more places, namely `image_reader.rs` and `aaudio.rs`. Both do not suffer this problem as the surrounding Rust object holds on to the `Box`, and Android only supports one callback at a time for either object. In contrast, the looper supports one callback _per `fd`_ and would require more work to keep track of this mapping instead.

MarijnS95:
Additionally, since `add_fd_with_callback` does not store this callback anywhere it shouldn't need to be `dyn` but can instead use a generic parameter. `Box<dyn FnMut...>` is now not a fat pointer anymore, and we can get rid of the double boxes :)

```rust
pub unsafe fn add_fd_with_callback<F: FnMut(RawFd) -> bool>(
    &self,
    fd: RawFd,
    ident: i32,
    events: FdEvent,
    callback: Box<F>,
) -> Result<(), LooperError> {
    extern "C" fn cb_handler<F: FnMut(RawFd) -> bool>(
        fd: RawFd,
        _events: i32,
        data: *mut c_void,
    ) -> i32 {
        unsafe {
            let mut cb = ManuallyDrop::new(Box::<F>::from_raw(data as *mut _));
            let keep_registered = cb(fd);
            if !keep_registered {
                ManuallyDrop::into_inner(cb);
            }
            keep_registered as i32
        }
    }
    let data = Box::into_raw(callback) as *mut _;
    match ffi::ALooper_addFd(
        self.ptr.as_ptr(),
        fd,
        ident,
        events.bits() as i32,
        Some(cb_handler::<F>),
        data,
    ) {
        1 => Ok(()),
        -1 => Err(LooperError),
        _ => unreachable!(),
    }
}
```
Thoughts?

msiglreith (approved):

ndk/src/looper.rs
```diff
@@ -264,27 +265,35 @@ impl ForeignLooper {
...
     /// See also [the NDK
     /// docs](https://developer.android.com/ndk/reference/group/looper.html#alooper_addfd).
-    pub fn add_fd_with_callback(
+    pub fn add_fd_with_callback<F: FnMut(RawFd) -> bool>(
         &self,
         fd: RawFd,
         ident: i32,
         events: FdEvent,
-        callback: Box<dyn FnMut(RawFd) -> bool>,
+        callback: Box<F>,
```
MarijnS95, line 273: Whoops, I just realized that we can probably hide the `Box<F>` implementation detail (just accept `: F` here, and `Box::new()` that internally), and I should also add a changelog entry for this _breaking_ change. @msiglreith thoughts?

msiglreith: right, that's better!

PR #174, opened 2021-08-07
ndk-glue: Document lock behaviour in `Event` and ensure pointer equality

MarijnS95:
This completes #134 by applying the same pointer-equality check in `on_input_queue_destroyed` to `on_window_destroyed`, and documents when the user should hold on to or release any of the `RWLock`s to ensure proper lifetimes for dependencies like surfaces in graphics APIs.

### DIFF
--- a/ndk-glue/src/lib.rs
+++ b/ndk-glue/src/lib.rs
@@ -122,8 +122,16 @@ pub enum Event {
     WindowCreated,
     WindowResized,
     WindowRedrawNeeded,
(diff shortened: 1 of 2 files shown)


ndk-glue/src/lib.rs
```diff
@@ -122,8 +122,16 @@ pub enum Event {
     WindowCreated,
```
MarijnS95, line 122: Should we add some docs about acquiring and holding the lock here too? And on `InputQueueCreated`?

msiglreith (approved):

PR #169, opened 2021-08-06
Run build-artifacts from Ubuntu and Windows in the emulator

MarijnS95:
For https://github.com/rust-windowing/android-ndk-rs/pull/162#discussion_r683810422

~TODO: We're not testing `cargo apk run` functionality anymore, how do we want to work that into here? We can place the commands inside `rust.yml` in a third emulator run.~ This is now happening in a third `local` run.

Note that this PR is intentionally _not_ based on the merge of #168 to demonstrate it actually failing! This would have caught the bogus refactor in #162.

PR #175, opened 2021-08-07
cargo-apk: Fixed the library name in case of multiple build artifacts 

msiglreith:
When specifying multiple build targets (e.g --examples) only the first artifact was used to derive the library name.
In the `AndroidManifest.xml` the key `android:name="android.app.lib_name"` would always have the same value for each artifact. The resulting application wouldn't be able to correctly load the library on runtime.

This surfaced in the CI changes https://github.com/rust-windowing/android-ndk-rs/pull/169 which builds all examples with a single command. The uploaded apk artifacts may not run correctly as they internally refer to the jni-audio library.

### DIFF
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fixed the library name in case of multiple build artifacts in the Android manifest.
+
 # 0.8.1 (2021-08-06)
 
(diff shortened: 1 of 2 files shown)


MarijnS95 (approved):
That's a nice find! Already searching around the codebase for more hidden `[0]` gems, but it looks like we're good thus far.

Curious how you spotted that on the #169 CI, as the yaml is still broken. Will try to fix that ASAP, we really need to get some solid CI going :)

PR #176, opened 2021-08-08
ndk/looper: Accept unboxed closure in add_fd_with_callback

MarijnS95:
For https://github.com/rust-windowing/android-ndk-rs/pull/172#discussion_r684681058

b0f2148 ("ndk/looper: Drop boxed callback after deregistering (#172)") improved `add_fd_with_callback` to use a generic type instead of `dyn FnMut`, allowing us to get rid of a double `Box` that would otherwise be needed to hold the fat pointer with vtable.  Since the API is changing for the next `ndk` release anyway (`ident` has been removed as this has to use the `ALOOPER_POLL_CALLBACK` constant) we can also remove the `Box` implementation detail.  A new `Box` is still created just like the previous implementation though, since the lifetime of the closure is now unknown (at least as long as Android might be calling it).

### DIFF
--- a/ndk/src/looper.rs
+++ b/ndk/src/looper.rs
@@ -269,7 +269,7 @@ impl ForeignLooper {
         &self,
         fd: RawFd,
         events: FdEvent,
-        callback: Box<F>,
+        callback: F,
     ) -> Result<(), LooperError> {
         extern "C" fn cb_handler<F: FnMut(RawFd) -> bool>(
             fd: RawFd,
@@ -285,7 +285,7 @@ impl ForeignLooper {
                 keep_registered as i32
             }
         }
-        let data = Box::into_raw(callback) as *mut _;
+        let data = Box::into_raw(Box::new(callback)) as *mut _;
         match unsafe {
             ffi::ALooper_addFd(
                 self.ptr.as_ptr(),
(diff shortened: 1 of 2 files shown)


ndk/src/looper.rs
```diff
@@ -269,7 +269,7 @@ impl ForeignLooper {
         &self,
         fd: RawFd,
         events: FdEvent,
-        callback: Box<F>,
+        callback: F,
```
dvc94ch, line 272: Doesn't make much difference I guess. If F is already a `Box<F': FnMut(RawFd) -> bool>` it'll be boxed twice. LGTM

dvc94ch (approved):
Doesn't make much difference I guess? If

ndk/src/looper.rs
MarijnS95: Hmm, I wasn't even aware that if `F: FnMut(RawFd) -> bool` then `Box<F>: FnMut(RawFd) -> bool` too. That makes this a non-breaking change, while at the same time not requiring users to manually wrap the thing in `Box::new`. It's anyway curious that we didn't do this before, despite the requirement to `Box` the `dyn` twice (receiving a `Box::into_raw` pointer on a `dyn` object obviously returns a fat-pointer, which does not fit in a `*mut c_void`). We could have just done `Box::new(Box::new())` before... Oh well.

PR #178, opened 2021-08-12
ndk-sys: Regenerate against NDK r23

MarijnS95:
Android 12 reached platform stability today, including the first stable release of NDK r23.

---

Will add some of the new API surface to our safe wrappers in a separate PR. Perhaps it's a good idea to regenerate against identical (r22.1, beta 3) headers first but with newer bindgen, to make the diff more clear? Not that it matters for squash-merges unless we put that in a separate PR too, LMK.

At the same time we might want to start thinking about guarding symbols based on a user-selected version. We could for example have a feature version per NDK or per SDK/API version so that there is a clear up-front contract for the programmer that they can only call certain certain functions if at least some version feature is enabled; that's how the various gtk-rs bindings do it. They have the advantage of directly querying that exact version from pkg-build though, and running into linker errors immediately if things don't match.


msiglreith:
I'm generally fine with the PR as is.

> At the same time we might want to start thinking about guarding symbols based on a user-selected version. We could for example have a feature version per NDK or per SDK/API version so that there is a clear up-front contract for the programmer that they can only call certain certain functions if at least some version feature is enabled; that's how the various gtk-rs bindings do it. They have the advantage of directly querying that exact version from pkg-build though, and running into linker errors immediately if things don't match.

Is this feature guarding on a per iterm level or generating having different. seperately generated bindings per API version? Does Android NDK provide fine grained metadata to do this?

PR #187, opened 2021-11-08
Avoid AAPT errors stemming from attempts of adding the same library twice

### LINKED ISSUE #186: Multiple shared libs dependent on a common dependency lead to packaging errors
… by attempting to `aapt add <apk> <lib>` it multiple times.

Replication source repo: https://github.com/zer0def/qmetaobject-rs/tree/a9b871c4def8f37be74810608137331f64a75689/examples/graph

Output:
```
    Finished dev [unoptimized + debuginfo] target(s) in 6.89s
 'lib/armeabi-v7a/libgraph.so'...
 'lib/armeabi-v7a/libc++_shared.so'...
 'lib/armeabi-v7a/libQt5Widgets.so'...
 'lib/armeabi-v7a/libc++_shared.so'...
Unable to add 'lib/armeabi-v7a/libc++_shared.so' to '/home/zer0def/sync/work/qmetaobject-rs/target/debug/apk/graph-unaligned.apk': already exists in archive
Error: Command `/opt/android-sdk/build-tools/31/aapt add /home/zer0def/sync/work/qmetaobject-rs/target/debug/apk/graph-unaligned.apk lib/armeabi-v7a/libc++_shared.so` had a non-zero exit code.
```

zer0def:
Fixes #186.

### DIFF
--- a/ndk-build/src/readelf.rs
+++ b/ndk-build/src/readelf.rs
@@ -51,8 +51,9 @@ impl<'a> UnalignedApk<'a> {
                 };
 
                 if let Some(path) = find_library_path(search_paths, &need)? {
-                    provided.insert(path.file_name().unwrap().to_str().unwrap().to_string());
-                    artifacts.push(path);
+                    if provided.insert(path.file_name().unwrap().to_str().unwrap().to_string()) {
+                        artifacts.push(path);
+                    }
                 } else {
                     eprintln!("Shared library \"{}\" not found.", need);
                 }


MarijnS95:
We are also building the unaligned library with ndk-build, is there a particular reason why we can't look back at libs that have been added previously, instead needing to unzip it for every new library that is being added?

MarijnS95:
In other words, `UnalignedApk` should probably be transitioned into something that holds an intermediary list/set of libraries. Much like `fn align()` right now there should be another function that flattens and deduplicates this set and adds all libraries at once, preferrably by spawning `aapt` just once.

zer0def:
We could (and probably should, as per implications of the added comment) check whether the previously built library is the same one. Just not yet sure how one would go about it, as I'm uncertain whether just checking the checksum would be sufficient.

MarijnS95:
Instead of checksums I'd just look at the filenames. If someone tries to add the same file in the same location twice, that's fine. If those files come from different locations (ie. looking at `$(realpath)` - but the target location inside the apk remains the same) that should be disallowed - if this becomes an issue we can later on add support for hash comparison, ie. when two crates require and ship the same precompiled lib.

zer0def:
Hey @MarijnS95, don't mean to be a pest or anything, but I'm curious whether this may still be on your radar? I generally force-push changes, as I'm used to rebase-based development. It's been a while since I've made changes, but I *think* they still roughly adhere to your request.

dvc94ch:
@zer0def I'm sure @MarijnS95 just forgot about it, thanks for the ping. However I'm not convinced about this approach. Maybe the code calling `aapt.add_lib` should ensure it only gets called once instead of aapt.add_lib silently ignoring it.

zer0def:
That's a fair observation, which also reduces the complexity of introduced changes - already pushed out.

ndk-build/src/readelf.rs
```diff
@@ -51,8 +51,11 @@ impl<'a> UnalignedApk<'a> {
                 };
 
                 if let Some(path) = find_library_path(search_paths, &need)? {
-                    provided.insert(path.file_name().unwrap().to_str().unwrap().to_string());
-                    artifacts.push(path);
+                    let file_name = path.file_name().unwrap().to_str().unwrap().to_string();
```
dvc94ch, line 54: Think this can be simplified even more:
```
if provided.insert(...) {
    artifacts.push(...);
}
```

dvc94ch (approved):
Much more comfortable with this change

ndk-build/src/readelf.rs
zer0def: Ack, incorporated.

dvc94ch (approved):

dvc94ch:
Thanks @zer0def!

MarijnS95:
> Hey @MarijnS95, don't mean to be a pest or anything, but I'm curious whether this may still be on your radar?

> @zer0def I'm sure @MarijnS95 just forgot about it, thanks for the ping.

I've just been away for a couple weeks, with too little time to catch up to every notification inbetween. Though it seems the last activity before this ping was 8 Nov, guess it got lost somehow? Thanks anyway @dvc94ch for simplifying this in review :+1: 

> I generally force-push changes, as I'm used to rebase-based development. It's been a while since I've made changes, but I _think_ they still roughly adhere to your request.

Glad you adhere to clean development standards, thanks for not making us/me clean up "Apply feedback" noise and other garbage when setting the final message during squash-merge :heart::+1: 

This is the first time GH refuses to render `@zer0def force-pushed from <xxx> to <yyy>` though, making me unable to see older versions of this PR and understanding what my review comment was about :thinking:

ndk-build/src/readelf.rs
```diff
@@ -51,8 +51,9 @@ impl<'a> UnalignedApk<'a> {
                 };
 
                 if let Some(path) = find_library_path(search_paths, &need)? {
-                    provided.insert(path.file_name().unwrap().to_str().unwrap().to_string());
-                    artifacts.push(path);
+                    if provided.insert(path.file_name().unwrap().to_str().unwrap().to_string()) {
```
MarijnS95, line 54: I presume `.file_name()` gives back an equal value to `need`, which is also just a file name without path? Can we perhaps simplify this by not calling `find_library_path` if `need` is already in `provided`? OTOH I always appreciate the return value of `fn insert(...) -> bool` being used :+1:

PR #197, opened 2021-11-16
cargo-apk: Use `min_sdk_version` to select compiler target

### LINKED ISSUE #193: target_sdk_version seems to affect the NDK build version?
Targetting a higher SDK version should only change the uses-sdk tag in the manifest, but setting it to say 31 to gain access to the latest features accessible from Java APIs, this happens:

```
Error: Path `"C:\\Android\\sdk\\ndk\\21.3.6528147\\toolchains\\llvm\\prebuilt\\windows-x86_64\\bin\\aarch64-linux-android31-clang.cmd"` 
```

To be more clear, the below causes the above, but should actually work, even with an older NDK:

```
[package.metadata.android.sdk]
min_sdk_version = 29
target_sdk_version = 31
max_sdk_version = 31
```

MarijnS95:
Fixes #193, cc @hrydgard 

According to [1] minSdkVersion is used to determine the compiler target and ultimately limit what API is available at compile-time.  This is most likely because using/linking newer API will result in runtime linker errors on these older platforms.

[1]: https://developer.android.com/ndk/guides/sdk-versions#minsdkversion

---

I think we still have a couple things to do here:
- Now that `min_sdk_version` is mandatory (see the unwrap), perhaps we should default it to the _minimum_ SDK level available in the NDK (the `min()` of `default_platform()`)?;
  (obviously reword the defaults in the cargo-apk docs);
- Remove the "arbitrary" default of minimum version 23?  Does anyone recall why this was chosen?  NDK r24 beta still ships with support for SDK level 21, which would then be the lowest;
- Is `targetSdkVersion` still mandatory?  If so, does it still make sense to set it to the max available in the SDK?;
- when we have these defaults, should we remove the arbitrary overrides in `ndk-examples`?

### DIFF
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+- Use `min_sdk_version` to select compiler target instead of `target_sdk_version` ([#197](https://github.com/rust-windowing/android-ndk-rs/pull/197)).
+  See https://developer.android.com/ndk/guides/sdk-versions#minsdkversion for more details.
+
 # 0.8.2 (2021-11-22)
 
 - Fixed the library name in case of multiple build artifacts in the Android manifest.
--- a/cargo-apk/src/apk.rs
+++ b/cargo-apk/src/apk.rs
@@ -79,13 +79,7 @@ impl<'a> ApkBuilder<'a> {
     pub fn check(&self) -> Result<(), Error> {
         for target in &self.build_targets {
             let triple = target.rust_triple();
-            let target_sdk_version = self
-                .manifest
-                .android_manifest
-                .sdk
-                .target_sdk_version
-                .unwrap();
-            let mut cargo = cargo_ndk(&self.ndk, *target, target_sdk_version)?;
+            let mut cargo = cargo_ndk(&self.ndk, *target, self.min_sdk_version())?;
             cargo.arg("check");
             if self.cmd.target().is_none() {
                 cargo.arg("--target").arg(triple);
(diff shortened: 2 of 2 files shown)


hrydgard:
targetSdkVersion is very important to be able to set separately from the others, because Android uses it to decide behavior at times. For example, the new Scoped Storage restrictions have completely different behavior in Android 11 depending on if you're targeting sdk version <= 29 or >= 30. 

minSdkVersion just restricts which devices you can install on. I think just for ease of maintenance something like 23 makes perfect sense, older devices are not really relevant anymore.

MarijnS95:
> targetSdkVersion is very important to be able to set separately from the others

Yes, merely checking if it still makes sense to set a default here, and if so, what. Not planning to remove/disallow the ability to override it from metadata in any way :wink: 

> just for ease of maintenance

Here I have a hunch it's easier to use the minimum available in the NDK, as 23 will probably get removed from a future NDK at some point. On the other hand, this could cause minor variations in compilation depending on the environment. However, that's not an issue as long as it compiles fine (`minSdkVersion` should only have an effect on what's available during compile-time).
Not so much for `targetSdkVersion` which currently _is_ dependent on the environment (only the default, though) and has the ability to cause much more havoc _at runtime_.

In other words, I think it should be safe to:
- Default `minSdkVersion` to whatever is at least available in the NDK (perhaps with a lower bound of 23, to not break existing builds);
- Default `targetSdkVersion` to the least required by Google Play as of writing, instead of being dependent on the maximum platform shipped with the SDK combined with what's in `Android/Sdk/platforms/android-xx`.

MarijnS95:
@hrydgard This is ready now. I'll open a separate PR for `target_sdk_version`, which is a different issue that needs its own documentation. Thanks for bringing this all to my attention!

hrydgard:
Awesome, thanks for taking care of this!

msiglreith (approved):

PR #198, opened 2021-11-17
ci emulator: Disable animations and other UI/UX features

MarijnS95:
We only care about starting a very simple UI-less app for now, and this seems to fix the deadlock that occurs right about every CI emulator job just before `adb install` runs.  It hopefully also makes the runs a bit faster, but this wasn't measured/compared in detail due to already high runtime variance.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -115,7 +115,7 @@ jobs:
         arch: x86_64
         # the `googleapis` emulator target is considerably slower on CI.
         target: default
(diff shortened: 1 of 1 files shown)


MarijnS95:
Great, I just got alerted of another local run that actually did time out :(

https://github.com/MarijnS95/android-ndk-rs/actions/runs/1470956247

Hopefully the AVD cache helps with the starts, but it's really strange that this is happening in the first place.

PR #200, opened 2021-11-22
[CI] Enable Android AVD cache

Gordon-F:
Enable AVD cache as discussed in #133. I hope it will fix problems with CI. Also includes changes from #198.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -108,6 +108,27 @@ jobs:
       run: |
         cargo install --path cargo-apk
         rustup target add x86_64-linux-android
+    - name: AVD cache
+      uses: actions/cache@v2
+      id: avd-cache
+      with:
+        path: |
+          ~/.android/avd/*
+          ~/.android/adb*
+        # One cache for 3 apk files
+        key: avd-emul
+    - name: create AVD and generate snapshot for caching
+      if: steps.avd-cache.outputs.cache-hit != 'true'
+      uses: reactivecircus/android-emulator-runner@v2.19.1
+      with:
+        api-level: 29
+        arch: x86_64
+        profile: Nexus 6
+        target: default
+        force-avd-creation: true
+        emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+        disable-animations: false
+        script: echo "Generated AVD snapshot for caching."
     - name: Start hello_world example
       uses: reactivecircus/android-emulator-runner@v2
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -108,14 +108,37 @@ jobs:
...
         rustup target add x86_64-linux-android
+    - name: AVD cache
+      uses: actions/cache@v2
+      id: avd-cache
+      with:
+        path: |
+          ~/.android/avd/*
+          ~/.android/adb*
+        # One cache for 3 apk files
+        key: avd-emul
```
MarijnS95, line 119: No need to mention that this is for 3 apk files specifically, just that this key is not variable meaning it doesn't depend on the job type/target... Doesn't need a comment at all IMO.

Perhaps `avd-emulator` reads nicer though?

.github/workflows/rust.yml
```diff
@@ -108,14 +108,37 @@ jobs:
...
+    - name: create AVD and generate snapshot for caching
+      if: steps.avd-cache.outputs.cache-hit != 'true'
+      uses: reactivecircus/android-emulator-runner@v2.19.1
```
MarijnS95, line 122:
```suggestion
      uses: reactivecircus/android-emulator-runner@v2
      
```

.github/workflows/rust.yml
```diff
@@ -108,14 +108,37 @@ jobs:
...
+      with:
+        api-level: 29
+        arch: x86_64
+        profile: Nexus 6
+        target: default
+        force-avd-creation: true
+        emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+        disable-animations: false
```
MarijnS95, line 130: I saw this in the example, but I still don't understand why AVD creation should _NOT_ disable animations?

MarijnS95, line 130: Let's keep this as similar as possible to the actual use:
```suggestion
        api-level: 29
        arch: x86_64
        # the `googleapis` emulator target is considerably slower on CI.
        target: default
        profile: Nexus 6
        disable-animations: true # Still not sure
        emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
        force-avd-creation: true
```

MarijnS95:
> I had the same locally

With all these review comments I should have submitted my own approach instead :thinking: - but still thanks for doing the prep as well.

.github/workflows/rust.yml
Gordon-F: I don't know too :)

Gordon-F:
> With all these review comments I should have submitted my own approach instead 🤔

Ok, let's close this in favor of your PR.

PR #180, opened 2021-08-28
Wrap video coding APIs

zarik5:
Add support for `AMediaCodec` to ndk-sys. I tried to generate the bindings but I was not able to.

PR #201, opened 2021-11-22
ndk-sys: Add all missing android-related headers and regenerate against 23.1.7779620

MarijnS95:
Commented-out headers contain C++-like directives and will need extra treatment if we wish to support them.

---

Note that this brings `ndk-sys` buildtime from 0.7s to 1.15s on my machine, which is a hit I'm willing to take for others to use or start working on bindings with no safe (`ndk` crate) counterpart just yet, see #180.  Especially considering the much longer build times that has, this is almost negligible.

Perhaps we can leave out SLES and OpenMAX for now?  Without those two build times are only 0.79s.

---

Otherwise, be sure to validate that I did not skip any other interesting headers.  `vulkan/vulkan_android.h` has intentionally been skipped because those extension defines are also part of the spec `vk.xml` and covered in ie. [ash](https://github.com/MaikKlein/ash) already.


zarik5:
In the case build times are a problem, feature flags could be added.

zarik5:
> Perhaps we can leave out SLES and OpenMAX for now?

SLES is deprecated in favor of AAudio, and still SLES is supported in the oboe library (there is already support in the CPAL crate). Likewise, OpenMAXAL was superseded by MediaCodec in Android 4.1.

MarijnS95:
> > Perhaps we can leave out SLES and OpenMAX for now?
> 
> SLES  is deprecated in favor of AAudio, and still SLES is supported in the oboe library (there is already support in the CPAL crate). Likewise, OpenMAXAL was superseded by MediaCodec in Android 4.1.

I added these at last resort (may have noticed that they were not previously on the ndk-sys branch), without giving it much thought. Sounds like we should not preliminarily provide these, especially considering that someone probably writes/uses a safe wrapper and the Android-specific bindings here are just a thin layer on top. However, I remember SLES being more extensive than AAudio back in the day.

msiglreith (approved):

PR #202, opened 2021-11-22
ci emulator: Cache AVD emulator setup to speed up repeated jobs

MarijnS95:
Our CI currently builds and runs the emulator three times, once for every possible APK source.  It is possible to speed up this setup by only building the image once as per [1] and storing it using GitHub's cache system.  This approach was suggested in https://github.com/rust-windowing/android-ndk-rs/pull/133#issuecomment-967221312.

[1]: https://github.com/ReactiveCircus/android-emulator-runner#usage

CC @Gordon-F for the suggestion!

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -94,31 +94,66 @@ jobs:
     strategy:
       matrix:
         source_os: [ubuntu-latest, windows-latest, local]
+    env:
+      api-level: 29
+      emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+      arch: x86_64
+      # the `googleapis` emulator target is considerably slower on CI.
+      target: default
+      profile: Nexus 6
+
     steps:
     - name: Checkout
       uses: actions/checkout@v2
+
     - uses: actions/download-artifact@v2
       name: Download hello_world APK
       if: ${{ matrix.source_os != 'local' }}
       id: download
       with:
         name: hello_world_${{ matrix.source_os }}_x86_64-linux-android
+
     - name: Install `cargo-apk` and add `x86_64-linux-android` target
       if: ${{ matrix.source_os == 'local' }}
       run: |
         cargo install --path cargo-apk
         rustup target add x86_64-linux-android
+
+    - name: AVD cache
+      uses: actions/cache@v2
+      id: avd-cache
+      with:
+        path: |
+          ~/.android/avd/*
+          ~/.android/adb*
+        # Bump the trailing number when making changes to the emulator setup below
+        key: avd-${{ env.api-level }}-1
+
+    - name: create AVD and generate snapshot for caching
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -94,31 +94,63 @@ jobs:
...
       with:
-        api-level: 29
+        api-level: ${{ env.api-level }}
         arch: x86_64
         # the `googleapis` emulator target is considerably slower on CI.
         target: default
         profile: Nexus 6
         emulator-options: -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
         disable-animations: true
+        force-avd-creation: false
```
MarijnS95, line 138: @Gordon-F the example in their readme also sets `force-avd-creation` to `false`. That seems like a sensible default so perhaps we shouldn't even set this at all. I don't mind it not creating an `avd` when it already exists, though that is extremely unlikely/impossible unless GH actions environment starts to ship with one...

.github/workflows/rust.yml
```diff
@@ -94,31 +94,63 @@ jobs:
     strategy:
       matrix:
         source_os: [ubuntu-latest, windows-latest, local]
+    env:
+      api-level: 29
```
Gordon-F, line 98: I think it will be great to make `arch`, `profile` and `target` part of `env`.

Gordon-F (approved):

.github/workflows/rust.yml
MarijnS95: I didn't expect these to change as often as the `api-level` (which is also at least used in 3 places, the rest is only dereferenced twice), but it's indeed useful to deduplicate at least `emulator-options` which is a rather "complicated" string.

I really, really with GH starts supporting YAML anchors, it would have made this 100x better:
https://github.com/MarijnS95/android-ndk-rs/actions/runs/1494051637/workflow

Gordon-F: > I didn't expect these to change as often as the api-level

Maybe we should have `api-level` part of matrix? At least `min_sdk_version` and `target_sdk_version`.

MarijnS95: I guess now that we have this cache it's probably worth running it that many times?

Gordon-F: We can try :) But, probably, after we make sure that the CI is stable enough.

MarijnS95: However, we'll probably need to come up with a more interesting test/example too. All these versions do is limit API available at compile vs build time, and that's really only interesting if that API is actually used.

Otherwise the only thing we're checking is if the resulting APK is still compatible.

MarijnS95: Let's land such a change after the CI is stable like this (and it looks like it is already) - note that the examples default to `min_sdk_version=16`.

msiglreith (approved):
nice!

PR #203, opened 2021-11-22
ndk-build,cargo-apk: Default `target_sdk_version` to 30 or lower

MarijnS95:
Depends on #197

As discussed in [197] setting `target_sdk_version` to the "arbitrary" highest available SDK version is nonsense.  This target version (unlike `min_sdk_version` which defines the least set of symbols that should be available) has real impact on the runtime of an application, in particular the compatibility or stringency of rules Android applies to your application.  Certain APIs may not work at all or be heavily restricted on newer target versions because they are deemed too dangerous, and Android expects the user has tested their app against these limitations and is communicating this by setting `target_sdk_version` to that particular value.  Hence this shouldn't change purely based on the environment, even for the default.

To retain some backwards compatibility with previous `cargo-apk` we set this to level 30 which is the least [required by Google Play] today, and exactly what users will have been targeting using NDK r22 (assuming the SDK for this `platform` was installed as well) since SDK version 31 support with NDK r23 only [arrived just last week].

[197]: https://github.com/rust-windowing/android-ndk-rs/pull/197#issuecomment-970283527
[required by Google Play]: https://developer.android.com/distribute/best-practices/develop/target-sdk
[arrived just last week]: https://github.com/rust-windowing/android-ndk-rs/pull/189

Thanks @hrydgard for bringing this to my attention!

### DIFF
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 - Use `min_sdk_version` to select compiler target instead of `target_sdk_version` ([#197](https://github.com/rust-windowing/android-ndk-rs/pull/197)).
   See https://developer.android.com/ndk/guides/sdk-versions#minsdkversion for more details.
+- Default `target_sdk_version` to `30` or lower (instead of the highest supported SDK version by the detected NDK toolchain)
+  for more consistent interaction with Android backwards compatibility handling and its increasingly strict usage rules:
+  https://developer.android.com/distribute/best-practices/develop/target-sdk
 
 # 0.8.2 (2021-11-22)
 
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -49,10 +49,10 @@ runtime_libs = "path/to/libs_folder"
 
 # See https://developer.android.com/guide/topics/manifest/uses-sdk-element
 #
-# Defaults to a `min_sdk_version` of 23 and `target_sdk_version` is based on the ndk's default platform.
+# Defaults to a `min_sdk_version` of 23 and `target_sdk_version` of 30 (or lower if the detected NDK doesn't support this).
 [package.metadata.android.sdk]
-min_sdk_version = 16
-target_sdk_version = 29
+min_sdk_version = 23
+target_sdk_version = 30
 max_sdk_version = 29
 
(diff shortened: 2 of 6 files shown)


cargo-apk/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+- **Breaking**: Default `target_sdk_version` to `29` or lower for more consistent
+  interaction with Android backwards compatibility and tighter rules:
```
MarijnS95, line 4: Not a fan of this wording, will see if I can come up with something better after a night's
 sleep.

hrydgard:
I think you misread the page (which is because it's not properly updated), because it says:

> Starting in August 2021, new apps will need to:
> * Target API level 30 (Android 11) or above and adjust for behavioral changes; except Wear OS apps, which must continue to target API level 28 or higher.

So the correct level to use by default is 30, not 29, since we are now after August.

MarijnS95:
Thanks for pointing that out. Indeed, that page literally starts out with:

> New apps and app updates must target Android 10 (API level 29) or higher

Hence the confusion. Will update this.

msiglreith (approved):

cargo-apk/README.md
```diff
@@ -49,10 +49,10 @@ runtime_libs = "path/to/libs_folder"
 
 # See https://developer.android.com/guide/topics/manifest/uses-sdk-element
 #
-# Defaults to a `min_sdk_version` of 23 and `target_sdk_version` is based on the ndk's default platform.
+# Defaults to a `min_sdk_version` of 23 and `target_sdk_version` of 30 (or lower if the detected NDK doesn't support this).
 [package.metadata.android.sdk]
 min_sdk_version = 16
```
msiglreith, line 54: should maybe update this one as well, considering we internally limit it to >=23
```suggestion
min_sdk_version = 23
```

MarijnS95: Yeah we should update this as it represents the default.

Note that we're leaving it at 16 for the examples, for the time being.

MarijnS95: This was actually an oversight in #197, fwiw :)

cargo-apk/CHANGELOG.md
MarijnS95: Updated. @msiglreith what do you think?

PR #204, opened 2021-11-24
Add android:exported support in manifest

adrien-ben:
This attribute is now required when targeting android 12 : https://developer.android.com/about/versions/12/behavior-changes-12#exported

### DIFF
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -123,6 +123,11 @@ launch_mode = "singleTop"
 # Defaults to "unspecified".
 orientation = "landscape"
 
+# See https://developer.android.com/guide/topics/manifest/activity-element#exported
+#
+# Defaults to "unspecified".
+exported = "true"
+
 # See https://developer.android.com/guide/topics/manifest/meta-data-element
 #
 # Note: there can be several .meta_data entries.
--- a/ndk-build/src/manifest.rs
+++ b/ndk-build/src/manifest.rs
@@ -93,6 +93,8 @@ pub struct Activity {
     pub name: String,
     #[serde(rename(serialize = "android:screenOrientation"))]
     pub orientation: Option<String>,
+    #[serde(rename(serialize = "android:exported"))]
+    pub exported: Option<bool>,
 
     #[serde(rename(serialize = "meta-data"))]
     #[serde(default)]
@@ -112,6 +114,7 @@ impl Default for Activity {
             launch_mode: None,
             name: default_activity_name(),
             orientation: None,
+            exported: None,
             meta_data: Default::default(),
             intent_filter: Default::default(),
         }


cargo-apk/README.md
```diff
@@ -123,6 +123,11 @@ launch_mode = "singleTop"
 # Defaults to "unspecified".
 orientation = "landscape"
 
+# See https://developer.android.com/guide/topics/manifest/activity-element#exported
+#
+# Defaults to "unspecified".
```
MarijnS95, line 128: The page you link just above states that the default is considered `false` when it is unset/unspecified.

I'm not sure if we should have this as tri-state, @maxded what do you think? We can effectively set this (and the other booleans in the manifest to):
- Unset, nothing is emitted in the XML: results in whatever Android says/thinks is the default;
- `false`: explicitly set to false, we could let the serializer default to this if the docs say it's the default when unspecified;
- `true`: explicitly set to true, same story for the default as above.

ndk-build/src/manifest.rs
```diff
@@ -112,6 +114,7 @@ impl Default for Activity {
             launch_mode: None,
             name: default_activity_name(),
             orientation: None,
+            exported: None,
```
MarijnS95, line 117: We should default this to `true` for the activity(ies) with intent that advertise `LAUNCHER` - at least for the one that is added by default with:

https://github.com/rust-windowing/android-ndk-rs/blob/7f6ce6fe0fd33a43aa606a20915b86301a2e0f3b/ndk-build/src/manifest.rs#L102-L106

and

https://github.com/rust-windowing/android-ndk-rs/blob/7f6ce6fe0fd33a43aa606a20915b86301a2e0f3b/ndk-build/src/manifest.rs#L124-L147

MarijnS95 (changes requested):

cargo-apk/README.md
adrien-ben: I left it unspecified by default to avoid breaking anything relying on the actual default manifest state. 

I think it might be a good idea to let it unspecified by default and let android versions prior to 12 handle it just the way it did before (considered true when any `IntentFilter` is present) and force the user to make the choice for android 12.

ndk-build/src/manifest.rs
adrien-ben: Android prior to 12 will already consider it true if missing. They made the choice to make it safer by forcing the devs to explictly set it. I think we should not reintroduce the "automatically true when `IntentFilter` is present".

cargo-apk/README.md
maxded: If unset does not break the default manifest of android versions prior to 12, which I don't think it does, I _think_ unset is the best option here. That way a user is guaranteed that the android documentation is their source of truth and means it is less error-prone on our side.

MarijnS95: Fair enough, leaving it unset and keeping it out of the manifest by default (ie `None`) is our best option here.

> and force the user to make the choice for android 12.

I disagree as by far most android-ndk-rs use-cases are single-activity native applications that are supposed to be started externally, see the comment below. When ndk-build provides the `MAIN/LAUNCHER` intent (because the user hasn't set it) we should also make it `exported=true` for them.

In any case, this comment was initially placed to request a change on the `# Defaults to "unspecified".` - `screenOrientation` defaults to the _string_ "unspecified" - this boolean cannot default to a string. How about documenting that it is a boolean that is unset (or set to `true` if the default `MAIN` intent is provided), and also inheriting some of the AOSP docs explaining that this is mandatory Since Android 12?

ndk-build/src/manifest.rs
MarijnS95: We provide a default `android.app.NativeActivity` activity with the default `LAUNCHER` target when no other `MAIN` intent action is specified to allow users to easily build single-executable Android activities with Rust (only a single `android.app.NativeActivity` can exist anyway - and even our manifest representation only contains a singular `activity: Activity`) without having to copy this verbose config into `Cargo.toml` for every project.

For the same purpose, if the user left `exported` unset (`None`) in their manifest we should do our due-diligence to set it to `true` for the default activity so that their app functions consistently across Android 12 and <= 11.

MarijnS95 (changes requested):

cargo-apk/README.md
adrien-ben: > I disagree as by far most android-ndk-rs use-cases are single-activity native applications that are supposed to be started externally, see the comment below. When ndk-build provides the MAIN/LAUNCHER intent (because the user hasn't set it) we should also make it exported=true for them.

Ok, it makes sense, I'll change that.

> In any case, this comment was initially placed to request a change on the # Defaults to "unspecified". - screenOrientation defaults to the string "unspecified" - this boolean cannot default to a string. How about documenting that it is a boolean that is unset (or set to true if the default MAIN intent is provided), and also inheriting some of the AOSP docs explaining that this is mandatory Since Android 12?

My bad I read "unspecified" as "unset" I'll change that too. Yeah I agree on with adding the documentation explaining that it is now required.

adrien-ben:
The default MAIN activity is added at serialization. Would it make sense to do it in `cargo_apk::apk::ApkBuilder::from_subcommand` under `// Set default Android manifest values` and remove the custom serialization function?

Something like 

```rust
if manifest
    .android_manifest
    .application
    .activity
    .intent_filter
    .iter()
    .all(|i| i.actions.iter().all(|f| f != "android.intent.action.MAIN"))
{
    manifest
        .android_manifest
        .application
        .activity
        .intent_filter
        .push(IntentFilter {
            actions: vec!["android.intent.action.MAIN".to_string()],
            categories: vec!["android.intent.category.LAUNCHER".to_string()],
            data: vec![],
        });

    manifest
        .android_manifest
        .application
        .activity
        .exported
        .replace(true);
}
```

MarijnS95:
@adrien-ben It may be a bit confusing to have default initialization in multiple places but it seems like the manifest in `ndk-build` does everything it can by default, and `// Set default Android manifest values` fills it up with values that come from the surrounding environment (ie. after knowing what crate is built and what NDK toolchain is available).

I'd vouch for moving this initialization function to a constructor for `AndroidManifest` but that may make it harder to parse/create `cargo_apk::manifest::Manifest` with said default values. Not sure if we should move it into `cargo_apk::manifest::Manifest::from_toml` either.

MarijnS95:
Note also that the serialization functions don't fill the struct in the same way `impl Default` does - instead these are inserted in the output when the struct is serialized, which may make things more confusing.

If we end up moving things around for `exported=true` we should _probably_ also move all `serialize_with` functions to a single, generic place - that fill the struct and not the final serialization? I wouldn't be opposed to add an `fn make_manifest_minimally_usable(&mut AndroidManifest)` to `ndk_build::manifest` as long as it documents what it inserts depending on what other conditions. This way - if necessary at all - applications can have a clearer view of what was parsed from `toml` vs what we'll end up generating. `serde(default = "...")` impacts the deserialized result already though.

ssrlive:
Hi guys, please merge this PR and release a new version so that we can test the rust code on the API 32 system.

MarijnS95:
I've integrated all the requested changes in #242 after radio silence since November 2021 here.

adrien-ben:
Hi, sorry about radio silence, I've been busy with work lately. Thanks for reporting the change.

PR #207, opened 2021-12-06
ndk/native_window: Use `release`/`acquire` fns for `Drop` and `Clone`

MarijnS95:
Much like ALooper and AHardwareBuffer the reference counter of ANativeWindow must be properly incremented on `Clone`, in turn allowing us to also `_release` the window as soon as it is dropped.

### DIFF
--- a/ndk/src/native_window.rs
+++ b/ndk/src/native_window.rs
@@ -1,7 +1,7 @@
-//! Bindings for `ANativeWindow`
+//! Bindings for [`ffi::ANativeWindow`]
 use std::ptr::NonNull;
 
-#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[derive(Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct NativeWindow {
     ptr: NonNull<ffi::ANativeWindow>,
 }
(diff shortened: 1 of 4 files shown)


MarijnS95:
Note that I haven't tested this PR locally yet. Curious what will happen when we release the latest reference with `ANativeWindow_release` in `on_window_destroyed`, or if we give up all references that way before Android destroys our window.

Much like `ALooper`, should we also call `_acquire` in `fn ptr()`?

dvc94ch (approved):

MarijnS95:
> Much like `ALooper`, should we also call `_acquire` in `fn ptr()`?

Finally got around to testing this, and we must indeed increment the refcount if we ever wish to also decrement it. Turns out we can "keep using" the window after finishing the activity, even though 

@zarik5 This should also addess the `NativeWindow` leaking from `AMediaCodec_createInputSurface`. Note that you're taking this window _by value_ in `set_input_surface`/`set_output_surface` which will require the caller to clone the window while resulting in a `drop()` soon afterwards. I don't think that's intended?

ndk/src/native_window.rs
```diff
@@ -1,21 +1,47 @@
...
+    /// `ptr` must be a valid pointer to an Android [`ffi::ANativeWindow`].
     pub unsafe fn from_ptr(ptr: NonNull<ffi::ANativeWindow>) -> Self {
         Self { ptr }
     }
 
+    /// Acquires ownership of `ptr`
+    ///
+    /// # Safety
+    /// `ptr` must be a valid pointer to an Android [`ffi::ANativeWindow`].
+    pub unsafe fn clone_from_ptr(ptr: NonNull<ffi::ANativeWindow>) -> Self {
```
MarijnS95, line 40: Naming suggestions welcome here. There are some cases (`fn from_ptr()`) where Android gives us a pointer that we own, ie. we should call `ANativeWindow_release` when done (and not call `ANativeWindow_acquire` when constructing our owned wrapper). There are also cases - `on_window_created` + `on_window_destroyed` - where Android does its own refcounting/lifetime management. If we wish to participate we should _also_ increment the refcount before even thinking about releasing it again.

zarik5:
I actually have never checked about leaked NativeWindow, thanks for the work.

dvc94ch:
still looks good

PR #209, opened 2021-12-20
ndk-sys: Use `jni-sys` for low-level JNI bindings

MarijnS95:
As indicated by the massive (+2M) number of downloads of the `jni-sys` crate [1] and many crates depending on it including our own `ndk` wrapper crate, `jni-sys` is definitely the de facto standard for Java Native Interface bindings.

As such, to be more type-compatible with this crate directly (which is anyway also a `-sys` crate) it makes sense to directly use its types instead of having intermediary type representations emitted by `bindgen`.

Not only are these types (and the layout tests) generally unused (as the pointers are cast to `jni-sys` pointers), their compile-time overhead directly offsets the cost of bundling `jni-sys`.  Plain `ndk-sys` costs 1.20s (for `aarch64-linux-android`), 1.29s with `jni-sys` added, and 1.19s again after omitting all the now-unnecessary `bindgen` types.

Note that `bindgen` does not support type remapping: instead these should be ignored from generation (blocklisted) and the replacement types imported or provided in the same scope where the generated result is `include!`d.

[1]: https://crates.io/crates/jni-sys

CC @laptou


MarijnS95:
Dropping 17.8k lines spread out over 4 autogenerated files is quite a feat :)

msiglreith (approved):
nice!

PR #213, opened 2021-12-27
Make `HardwareBuffer::as_ptr()` public

zarik5:
~I suppose this was a mistake because there is no other way of accessing the underlying pointer.~ It probably isn't a mistake since other structs also have `as_ptr()` private. But `HardwareBuffer` should be spacial and have `as_ptr()` public, I need the pointer to import the memory into Vulkan.

### DIFF
--- a/ndk/src/hardware_buffer.rs
+++ b/ndk/src/hardware_buffer.rs
@@ -131,7 +131,7 @@ impl HardwareBuffer {
         Self { inner: ptr }
     }
 
-    fn as_ptr(&self) -> *mut ffi::AHardwareBuffer {
+    pub fn as_ptr(&self) -> *mut ffi::AHardwareBuffer {
         self.inner.as_ptr()
     }
 
--- a/ndk/src/hardware_buffer.rs
+++ b/ndk/src/hardware_buffer.rs
@@ -131,6 +131,11 @@ impl HardwareBuffer {
         Self { inner: ptr }
     }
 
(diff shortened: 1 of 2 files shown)


ndk/src/hardware_buffer.rs
```diff
@@ -131,7 +131,7 @@ impl HardwareBuffer {
         Self { inner: ptr }
     }
 
-    fn as_ptr(&self) -> *mut ffi::AHardwareBuffer {
+    pub fn as_ptr(&self) -> *mut ffi::AHardwareBuffer {
```
MarijnS95, line 134: @zarik5 Since this is a "special case" as you call it, perhaps add a documentation comment explaining how/why the user can use it.

```suggestion
    /// Returns the underlying [`ffi::AHardwareBuffer`] pointer
    ///
    /// The pointer can be used to import this hardware buffer into a Vulkan memory object using [`VK_ANDROID_external_memory_android_hardware_buffer`].
    ///
    /// [`VK_ANDROID_external_memory_android_hardware_buffer`]: https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_ANDROID_external_memory_android_hardware_buffer.html
    pub fn as_ptr(&self) -> *mut ffi::AHardwareBuffer {
```

zarik5: That's great, thanks for the suggestion

MarijnS95 (approved):

PR #216, opened 2022-01-14
Safe mediacodec bindings

zarik5:
Add `MediaCodec` and `MediaFormat` bindings with most of their methods. Crypto and DRM support is not implemented.

**Testing**
Only the decoder code path is tested. [sample](https://github.com/alvr-org/ALVR/blob/master/alvr/experiments/client/src/video_decoder/mediacodec.rs)

**Note:** This PR needs to be rebased after #213 is merged

### DIFF
--- a/ndk/src/media/media_codec.rs
+++ b/ndk/src/media/media_codec.rs
@@ -1,3 +1,5 @@
+#![warn(trivial_casts)]
+
 use super::{construct, construct_never_null, NdkMediaError, Result};
 use crate::native_window::NativeWindow;
 use std::{
@@ -445,11 +447,8 @@ impl InputBuffer<'_> {
     pub fn get_mut(&mut self) -> &mut [u8] {
         unsafe {
             let mut out_size = 0;
-            let buffer_ptr = ffi::AMediaCodec_getInputBuffer(
-                self.codec.as_ptr(),
-                self.index,
-                &mut out_size,
-            );
+            let buffer_ptr =
+                ffi::AMediaCodec_getInputBuffer(self.codec.as_ptr(), self.index, &mut out_size);
             slice::from_raw_parts_mut(buffer_ptr, out_size as usize)
         }
     }
@@ -466,12 +465,12 @@ impl OutputBuffer<'_> {
     pub fn get_buffer(&self) -> &[u8] {
         unsafe {
             let mut _out_size = 0;
-            let buffer_ptr = ffi::AMediaCodec_getOutputBuffer(
-                self.codec.as_ptr(),
-                self.index,
-                &mut _out_size,
-            );
-            slice::from_raw_parts(buffer_ptr.add(self.info.offset as usize), self.info.size as usize)
+            let buffer_ptr =
+                ffi::AMediaCodec_getOutputBuffer(self.codec.as_ptr(), self.index, &mut _out_size);
+            slice::from_raw_parts(
+                buffer_ptr.add(self.info.offset as usize),
+                self.info.size as usize,
+            )
         }
     }
 
--- a/ndk/src/media/media_codec.rs
+++ b/ndk/src/media/media_codec.rs
@@ -48,50 +48,47 @@ impl MediaFormat {
     pub fn get_i32(&self, key: &str) -> Option<i32> {
         let name = CString::new(key).unwrap();
         let mut out = 0;
-        unsafe { ffi::AMediaFormat_getInt32(self.as_ptr(), name.as_ptr(), &mut out as _) }
-            .then(|| out)
+        unsafe { ffi::AMediaFormat_getInt32(self.as_ptr(), name.as_ptr(), &mut out) }.then(|| out)
     }
 
     pub fn get_i64(&self, key: &str) -> Option<i64> {
         let name = CString::new(key).unwrap();
         let mut out = 0;
-        unsafe { ffi::AMediaFormat_getInt64(self.as_ptr(), name.as_ptr(), &mut out as _) }
-            .then(|| out)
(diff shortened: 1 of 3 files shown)


ndk/src/media/media_codec.rs
```diff
@@ -0,0 +1,501 @@
...
+        let name = CString::new(key).unwrap();
+        let mut out: ffi::size_t = 0;
+        unsafe { ffi::AMediaFormat_getSize(self.as_ptr(), name.as_ptr(), &mut out as _) }
+            .then(|| out as _)
+    }
+
+    pub fn get_buffer(&self, key: &str) -> Option<&[u8]> {
+        let name = CString::new(key).unwrap();
+        let mut out_buffer: *mut c_void = ptr::null_mut();
+        let mut out_size: ffi::size_t = 0;
```
MarijnS95, line 79: No need for the explicit typeS here.

ndk/src/media/media_codec.rs
```diff
@@ -0,0 +1,501 @@
...
+    pub fn get_f32(&self, key: &str) -> Option<f32> {
+        let name = CString::new(key).unwrap();
+        let mut out = 0.0;
+        unsafe { ffi::AMediaFormat_getFloat(self.as_ptr(), name.as_ptr(), &mut out as _) }
+            .then(|| out)
+    }
+
+    pub fn get_usize(&self, key: &str) -> Option<usize> {
+        let name = CString::new(key).unwrap();
+        let mut out: ffi::size_t = 0;
```
MarijnS95, line 71: No need for the explicit type here.

ndk/src/media/media_codec.rs
```diff
@@ -0,0 +1,501 @@
...
+    pub fn get_buffer(&self, key: &str) -> Option<&[u8]> {
+        let name = CString::new(key).unwrap();
+        let mut out_buffer: *mut c_void = ptr::null_mut();
+        let mut out_size: ffi::size_t = 0;
+        unsafe {
+            ffi::AMediaFormat_getBuffer(
+                self.as_ptr(),
+                name.as_ptr(),
+                &mut out_buffer as _,
+                &mut out_size as _,
```
MarijnS95, line 85: These will coerce, no need for `as _`.

MarijnS95 (changes requested):
For starters, there are a lot of unnecessary `as _` casts and typed initializations. Please either be explicit about a cast or remove it altogether.

Adding `#![warn(trivial_casts)]` at the top of this file (hopefully we can do that repo-wide at some point) will help you track these down.

MarijnS95:
Fwiw feel free to leave the warning at the top of the file for the meantime. I'll work out to combine that with #219.

MarijnS95:
> ```
> cargo clippy -p ndk --target armv7-linux-androideabi --features api-level-30 --features media
> cargo clippy -p ndk --target aarch64-linux-android --features api-level-30 --features media
> ```

You may also use `x86_64`/`i686` as those have a different pointer type for `c_char` (`u8`), afaik `armv7` also has `i8` like `aarch64`?

zarik5:
> You may also use x86_64/i686 as those have a different pointer type for c_char (u8), afaik armv7 also has i8 like aarch64?

Ok, tried and no extra warnings or errors.

zarik5:
There are still some warnings because of unused imports when using api-level-24, not only from my code

MarijnS95:
> There are still some warnings because of unused imports when using api-level-24, not only from my code

That is something we should perhaps build-test in the CI and fix? Mind opening a separate PR for that (or issue if nontrivial)?

zarik5:
Can this PR be merged? The warnings can be addressed in a second moment.

dvc94ch (approved):

dvc94ch:
LGTM, thanks @zarik5

MarijnS95:
I wasn't actually okay with this yet, but that's what I get for being on holiday. Doc comments with double instead of triple slashes, [`get_` prefixes](https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter), changelog in the past tense and no reference to this PR, inconsistent `use` statements that are unused when certain features are not enabled, `enum` that doesn't [derive common traits like `Eq`](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits) but instead has a strange `matches!` (why aren't the flags modeled directly with a `#[repr(u32)]` enum?), and that's just the stylistic nits.

dvc94ch:
Sometimes things are good enough. But agree that it would be better with the changes you mentioned.

MarijnS95:
If crates are going to be maintained (or at least PRs merged) with a "good is good enough" mentality I'll have to step down as a maintainer, as I can't be cleaning up behind peoples back to uphold an acceptable feeling of responsibility towards said project.

That is, unless there's clear agreement and documentation that a PR is dragging on too long (in terms of discussions, not silence), and is best merged in preparation for followup changes to land through separate PRs.

zarik5:
I'm sorry @MarijnS95, I realize that things were rushed. I tried picking up the style around but I did a poor job. The wrong doc comments were the most glaring mistake for me. I think it would really help a PR checklist (not necessarily in this repo, you could link to a good one in another repo or website).

MarijnS95:
I don't think a checklist for standard style errors really exists or make sense; it should all be caught by (locally runnable) CI tools like `clippy`. In the grand scheme of things I'd like to fail on many more default-allow (rustc or clippy) lint warnings, but it takes a ginormous effort to get this project to pass stricter lints initially :no_mouth: 

We could however use a checklist for obvious things like requiring a changelog entry if a PR modifies public-facing API/behaviour.

MarijnS95:
In any case, my post above links at least twice to https://rust-lang.github.io/api-guidelines/ - that's a good starting place if you're looking for more guidance.

PR #218, opened 2022-01-24
cargo-apk: Add `cargo apk version` command for querying release version

### LINKED ISSUE #217: `cargo-apk`: Parse and act on commandline arguments before finding and panicking on incomplete workspace
When running commands such as `cargo apk help` or the upcoming `cargo apk version`, which should run _anywhere_ and **don't depend on the workspace in `$PWD`**, `cargo-apk` still exits with a rather useless error message:

```
Error: Didn't find Cargo.toml
```

We should refactor `cargo-subcommand` to drop this workspace requirement for commands that don't need it. More details on fixing this in https://github.com/rust-windowing/android-ndk-rs/issues/182#issuecomment-926548594.

---

@dvc94ch Does this description supersede https://github.com/rust-windowing/android-ndk-rs/issues/182?

MarijnS95:
Print the package name and version from `Cargo.toml` that the crate/binary was built with.

Note that we still have to address [EDIT: don't use the word `fix` here to not trigger auto issue linking] #217 somehow to make this truly useful, without having to switch to a directory that contains a valid crate setup.


MarijnS95:
:( Our CI got broken by an untested crate release: https://github.com/harryfei/which-rs/pull/49

msiglreith (approved):

PR #219, opened 2022-01-24
ndk: Remove and warn on trivial casts

MarijnS95:
`trivial-casts` is allowed by default as it still has some odd cases where the lint triggers incorrectly or is difficult to fix.  For our intents and purposes however it works just fine and prevents against riddling the code with unnecessary - usually untyped - `as` casts.

https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#explanation-22

CC @zarik5, #216

### DIFF
--- a/ndk/src/configuration.rs
+++ b/ndk/src/configuration.rs
@@ -136,7 +136,7 @@ impl Configuration {
     pub fn country(&self) -> Option<String> {
         let mut chars = [0u8; 2];
         unsafe {
-            ffi::AConfiguration_getCountry(self.ptr.as_ptr(), chars.as_mut_ptr() as *mut _);
+            ffi::AConfiguration_getCountry(self.ptr.as_ptr(), chars.as_mut_ptr().cast());
         }
         if chars[0] == 0 {
             None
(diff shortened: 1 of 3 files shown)


ndk/src/configuration.rs
```diff
@@ -136,7 +136,7 @@ impl Configuration {
     pub fn country(&self) -> String {
         let mut result = "  ".to_owned();
         unsafe {
-            ffi::AConfiguration_getCountry(self.ptr.as_ptr(), result.as_mut_ptr() as *mut _);
+            ffi::AConfiguration_getCountry(self.ptr.as_ptr(), result.as_mut_ptr());
```
MarijnS95, line 139: Fwiw besides the build error on x86 platforms (not sure how to cleanly perform the cast without tripping `trivial_casts` besides just disabling it either locally or not pulling in this PR altogether), this can also return two null-characters (https://developer.android.com/ndk/reference/group/configuration#aconfiguration_getcountry) and should realistically return an `Option` too.

dvc94ch (approved):

PR #223, opened 2022-02-07
Add ndk-context.

### LINKED ISSUE #211: Runtime problems when including multiple versions of ndk-glue
Having multiple versions of ndk-glue in the dependency tree leads to crashes because of uninitialized `NATIVE_ACTIVITY`.

Example: Including ndk-glue 0.5. When entering `main()`, `NATIVE_ACTIVITY` 0.5 gets initialized. Including Oboe crate that uses ndk-glue 0.4.0. When enumerating audio devices, i get:
```
thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', /Users/ric/.cargo/registry/src/github.com-1ecc6299db9ec823/ndk-glue-0.4.0/src/lib.rs:59:39
```

I tried using `[patch.crates-io]` but that does not seem to work.

This bug might be related to using `resolver = "2"` as required by wgpu.

dvc94ch:
Closes #211


rklaehn (approved):

zarik5 (changes requested):

zarik5:
As this PR stands, winit will not be able to replace ndk-glue with ndk-context, but maybe this is intended.

dvc94ch:
> As this PR stands, winit will not be able to replace ndk-glue with ndk-context, but maybe this is intended.

This is indeed intended. winit (and presumably a game engine) is highly tied to the runtime and cargo-apk. I think this should be enough for most crates wanting to add some android support. If there are crates for which this isn't sufficient we can collect the exact reasons why not in an issue and then see if we have to take further steps.

dvc94ch:
Ping @MarijnS95

dvc94ch:
@msiglreith looks like @MarijnS95 is on vaccation or something. what are your thoughts on this PR?

msiglreith (approved):
API-wise it looks good to me. The crate should be listed in the README and probably also include a section regarding `ndk-context` vs `ndk-glue` somewhere

dvc94ch:
So wondering if we shoud remove the `fn native_activity()` from `ndk-glue`. `winit` is using it to access the `AssetManager`. I think there is probably a way to get an `AssetManager` using only `ndk-context`. Maybe marking it as deprecated would get the message accross to people to stop using it.

dvc94ch:
Yep, we should deprecate it [0] shows how to access it using the jni.

- [0] https://stackoverflow.com/questions/22436259/android-ndk-why-is-aassetmanager-open-returning-null/22436260#22436260

MarijnS95:
> looks like @MarijnS95 is on vac~c~ation or something

Not anymore :wink:. Taking a (still rather short) break from open source was well needed.

---

I thought the idea was to remove all `static`s from `ndk-glue` and shove them into `ndk-context` (with the premise that we never do any breaking changes with breaking updates). It is yet unclear to me if it's right to assume that all points of contention over these duplicate `static` variables originate from purely needing the `vm` and/or `activity`? Looks like "deprecating" `native_activity()` is a bit too harsh?

ndk-glue/src/lib.rs
```diff
@@ -55,6 +55,7 @@ lazy_static! {
 
 static mut NATIVE_ACTIVITY: Option<NativeActivity> = None;
 
+#[deprecated]
```
MarijnS95, line 58: Not sure if this is a good idea. If it is, at least set a sensible message here (`#[deprecated = "..."]`).

ndk-context/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+        self.context_jobject
+    }
+}
+
+/// Main entry point to this crate. Returns an [`AndroidContext`].
+pub fn android_context() -> AndroidContext {
+    unsafe { ANDROID_CONTEXT.expect("android context was initialized") }
+}
+
+/// Initializes the [`AndroidContext`]. [`AndroidContext`] is initialized by `ndk-glue` before
```
MarijnS95, line 75: `https://crates.io/crates/ndk-glue` just like we link to the `jni` crate?

ndk-context/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+/// main is called.
+///
+/// # Safety
+///
+/// The pointers need to be valid.
+pub unsafe fn initialize_android_context(java_vm: *mut c_void, context_jobject: *mut c_void) {
```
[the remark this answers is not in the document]
dvc94ch: well, it's unsafe to call after main was called (as it can spawn threads and cause data races etc) also it's unsafe to not be initialized before main. I think `#[doc(hidden)]` as originally proposed is probably the right thing to do.

ndk-glue/src/lib.rs
dvc94ch: will remove

dvc94ch:
Yes, think removing the static `NATIVE_ACTIVITY` is the right thing to do (and complementary to ndk-context).  Although not sure about removing `NATIVE_WINDOW`, `INPUT_QUEUE` and `CONTENT_RECT`.

So in general are you ok with the ndk-context approach to partly solving this problem?

dvc94ch:
So removed the NATIVE_ACTIVITY global as I think this is one we agree on and is worth doing.

MarijnS95:
https://github.com/MarijnS95/android-ndk-rs/compare/anativeactivity-instance @dvc94ch This is my draft idea for removing fallible `static` globals entirely, I can open a PR and we can iterate on it in parallel with your changes.

dvc94ch:
@MarijnS95 that looks good, I think it's an improvement for some use cases. But in general I still think we should try to discourage the use of ndk-glue unless it's absolutely necessary. If you open a PR we can merge it and I'll change jni-audio back to using ndk-context.

ndk-context/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+pub fn android_context() -> AndroidContext {
+    unsafe { ANDROID_CONTEXT.expect("android context was initialized") }
+}
+
+/// Initializes the [`AndroidContext`]. [`AndroidContext`] is initialized by [__ndk-glue__](https://crates.io/crates/ndk-glue)
+/// before main is called.
+///
+/// # Safety
+///
+/// The pointers must be valid and this function must be called exactly once before `main` is
```
MarijnS95, line 80: Fwiw you can decide to panic if `ANDROID_CONTEXT` was not `None` to uphold this guarantee. `ANDROID_CONTEXT.replace(AndroidContext { ... }).is_none()`.

MarijnS95:
> But in general I still think we should try to discourage the use of ndk-glue unless it's absolutely necessary.

I think both PRs are a great step in that direction :wink:. Mine still has some open questions though:
- what if a client application already uses the public `instance` field for its own cases?
- `static` is gone now, but what if two different `ndk-glue` versions with incompatible `GlueInstance` layout try to cast and use the same pointer?

Will need some more thoughts, I think I'll get this merged + released + backported ASAP, then PR the migration to most obvious crates and see where to move next.

ndk-context/src/lib.rs
```diff
@@ -0,0 +1,87 @@
+//! Provides a stable api to rust crates for interfacing with the android platform. It is
+//! initialized by the runtime, usually [__ndk-glue__](https://crates.io/crates/ndk-glue),
+//! but could also be initialized by java or kotlin code when embedding in an existing android
+//! project.
```
MarijnS95, line 4: Just in case, perhaps mention here that having this crate duplicated in the crate graph is dangerous (and we should do everything in our power to never make breaking changes again).

ndk-examples/examples/jni_audio.rs
```diff
@@ -7,9 +7,8 @@ const GET_DEVICES_OUTPUTS: jni::sys::jint = 2;
 
 fn enumerate_audio_devices() -> Result<(), Box<dyn std::error::Error>> {
     // Create a VM for executing Java calls
-    let native_activity = ndk_glue::native_activity();
-    let vm_ptr = native_activity.vm();
-    let vm = unsafe { jni::JavaVM::from_raw(vm_ptr) }?;
+    let ctx = ndk_context::android_context();
+    let vm = unsafe { jni::JavaVM::from_raw(ctx.vm() as *mut jni::sys::JavaVM) }?;
```
MarijnS95, line 11:
```suggestion
    let vm = unsafe { jni::JavaVM::from_raw(ctx.vm().cast()) }?;
```

Perhaps?

ndk-examples/examples/jni_audio.rs
```diff
@@ -18,7 +17,7 @@ fn enumerate_audio_devices() -> Result<(), Box<dyn std::error::Error>> {
 
     let audio_manager = env
         .call_method(
-            native_activity.activity(),
+            ctx.context() as jni::sys::jobject,
```
MarijnS95, line 20:
```suggestion
            ctx.context().cast(),
```

Should work here too?

ndk-glue/src/lib.rs
```diff
@@ -165,6 +165,10 @@ pub unsafe fn init(
     callbacks.onLowMemory = Some(on_low_memory);
 
     let activity = NativeActivity::from_ptr(activity);
+    ndk_context::initialize_android_context(
+        activity.vm() as *mut std::ffi::c_void,
+        activity.activity() as *mut std::ffi::c_void,
```
MarijnS95, line 170: And `.cast()` here. Maybe the `ndk-context` docs too?

README.md
```diff
@@ -8,6 +8,7 @@ Name | Description | Badges
 --- | --- | ---
 [`ndk-sys`](./ndk-sys) | Raw FFI bindings to the NDK | [![crates.io](https://img.shields.io/crates/v/ndk-sys.svg)](https://crates.io/crates/ndk-sys) [![crates.io](https://docs.rs/ndk-sys/badge.svg)](https://docs.rs/ndk-sys)
 [`ndk`](./ndk) | Safe abstraction of the bindings | [![crates.io](https://img.shields.io/crates/v/ndk.svg)](https://crates.io/crates/ndk) [![crates.io](https://docs.rs/ndk/badge.svg)](https://docs.rs/ndk)
+[`ndk-context`](./ndk-context) | Runtime API | [![crates.io](https://img.shields.io/crates/v/ndk-context.svg)](https://crates.io/crates/ndk-context)
```
MarijnS95, line 11: Not sure if this is strictly an API, but I can't really come up with a short enough alternative. `Runtime handle storage` but that's weird perhaps.

ndk-context/src/lib.rs
MarijnS95: Dangerous in the sense that you'll get annoying errors, same as we get right now and are trying to solve by moving to a crate that's a bit more stable.

ndk-context/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "ndk-context"
+version = "0.1.0"
+authors = ["The Rust Windowing contributors"]
+edition = "2021"
+description = "API for accessing the jni."
```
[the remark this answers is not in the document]
dvc94ch: How about "android handles for accessing the jni"? Essentially that's what it is as you can access any android jni api with these.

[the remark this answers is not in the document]
dvc94ch: or "handles for accessing android apis"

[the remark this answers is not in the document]
MarijnS95: Perfect.

ndk-context/src/lib.rs
dvc94ch: well, I guess that's up to us to make sure we don't release breaking changes. I think it's simple enough that we can say we nailed it.

ndk-context/src/lib.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+/// Main entry point to this crate. Returns an [`AndroidContext`].
+pub fn android_context() -> AndroidContext {
+    unsafe { ANDROID_CONTEXT.expect("android context was not initialized") }
+}
+
+/// Initializes the [`AndroidContext`]. [`AndroidContext`] is initialized by [__ndk-glue__](https://crates.io/crates/ndk-glue)
+/// before main is called.
```
MarijnS95, line 76:
```suggestion
/// before `main` is called.
```

MarijnS95:
Looks like GH comment notifications are all over the place. Are we removing `native_activity()`, deprecating it at least, or leaving it to a separate PR entirely?

MarijnS95:
Do we need a changelog for the new crate, and the changes listed in the `ndk-glue` crate? Then I think we are ready for merging and releasing. I'm preparing the necessary changes to backport this to older `ndk-glue` crates which are currently used by at least `winit`.

dvc94ch:
> Looks like GH comment notifications are all over the place. Are we removing native_activity(), deprecating it at least, or leaving it to a separate PR entirely?

Think we can hold off for with releasing until your PR lands, I removed my changes. Got to go afk for a while, will add a CHANGELOG afterwards.

MarijnS95:
We can do a non-breaking patch release (same for the backports) now with `#[deprecated = "Don't use for `vm()` purposes"]`, preparing everyone for this future. Or without the deprecation, as my PR would be breaking anyway.

zarik5:
The crate oboe requires the VM and Context pointers, but also the sdk version. Is there a way of getting the sdk version from the context pointer?

dvc94ch:
yes, should be possible, although will require fiddling a bit with the jni.

https://developer.android.com/reference/android/os/Build

zarik5:
@dvc94ch Oh that's great, thanks!

PR #225, opened 2022-02-15
ndk-glue: Initialize ndk-context on 0.5 stable releases

MarijnS95:
In order to show that [#223] solves our problem ([#211]) with multiple ndk-glue versions in tree, all having their own `static` globals, backport the `ndk-context` initialization to `ndk-glue 0.5` so that everyone on `winit 0.26` (pending a `winit 0.27` release) can be compatible with other crates using the Java VM and/or Android Context.  These crates will need to migrate to `ndk-context` first though.

[#211]: https://github.com/rust-windowing/android-ndk-rs/issues/211
[#223]: https://github.com/rust-windowing/android-ndk-rs/pull/223

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,7 +2,7 @@ name: Publish
 
 on:
   push:
-    branches: [master]
+    branches: [master, '*-stable']
     paths: "**/Cargo.toml"
 
 jobs:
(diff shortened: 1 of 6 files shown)


dvc94ch (approved):

MarijnS95:
https://github.com/rust-windowing/android-ndk-rs/actions/runs/1847041444 Looks like it works, publish is running :crossed_fingers:

PR #235, opened 2022-03-02
Update README.md

ManevilleF:
Fixed the `adb logcat` command


dvc94ch:
This seems to depend on the shell. The command works for me without quotes in sh and bash, but not fish (my default shell). So I'd say the command is fine, users of non posix shells are expected to adapt.

PR #236, opened 2022-03-03
Allow specifying manifest `package` property in `Cargo.toml`

### LINKED ISSUE #234: Control over the AndroidManifest.xml package name
Currently `cargo apk build` and `cargo apk run` will set the Android package name for you, to `"rust.<package name in Cargo.toml>"`.

Can we add a `Cargo.toml` property like `apk_name` that controls the APK package name in `AndroidManifest.xml`?

Something like:
```toml
[package.metadata.android]
apk_package = "com.blah.blahblah" # defaults to "rust.<package.name>"
# ...
```
that generates
```xml
<manifest package="com.blah.blahblah">...</manifest>
```

zjijz:
Allows specifying `package` under the `[package.metadata.android]` section of `Cargo.toml`. This removes an `assert` that `package` is empty and only uses the previous forced value of `rust.<crate_name>` if `package.metadata.android.package` is empty.

cc: @MarijnS95 

Closes #234

### DIFF
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Default `target_sdk_version` to `30` or lower (instead of the highest supported SDK version by the detected NDK toolchain)
   for more consistent interaction with Android backwards compatibility handling and its increasingly strict usage rules:
   https://developer.android.com/distribute/best-practices/develop/target-sdk
+- Allow manifest `package` property to be provided in `Cargo.toml` ([#236](https://github.com/rust-windowing/android-ndk-rs/pull/236)).
 
 # 0.8.2 (2021-11-22)
 
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -27,6 +27,9 @@ Following configuration options are supported by `cargo apk` under `[package.met
 
 ```toml
 [package.metadata.android]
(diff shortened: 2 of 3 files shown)


MarijnS95 (approved):
I think it was me who requested for the assert in https://github.com/rust-windowing/android-ndk-rs/pull/106#discussion_r588291061, as it was unconditionally overridden also before #106. I don't think there's a need for that, and don't see any implication in doing it like this.

Not sure if we should assign the default in `from_subcommand` though, but it wasn't anyway used outside `build()`, perhaps that's something for the cleanups and refactors around `cargo-subcommand` mentioned by @dvc94ch.

MarijnS95:
Please mention this change in `cargo-apk/CHANGELOG.md` :)

dvc94ch (approved):
lgtm, thanks

zjijz:
Sorry for missing your responses earlier!

I updated the `CHANGELOG.md` referencing this PR.

In terms of where to do the default value override check, I think the current logic is split between `from_subcommand` and `build`, but should be easy to refactor into one method. 

Let me know if you want to block landing this on doing that.

zjijz:
@MarijnS95 @dvc94ch bump since I don't have write access to actually land this.

dvc94ch (approved):

PR #238, opened 2022-03-08
Update to `cargo-subcommand 0.8.0` with `clap` argument parser

### LINKED ISSUE #182: Calling `cargo apk` (with e.g. `help`) outside of a crate/workspace should work
I get a `Didn't find Cargo.toml` error in `cargo-subcommand`, so it's not really a bug in `cargo-apk`. However since `cargo-apk` is the only user it might make sense to move it into this repo?

### LINKED ISSUE #217: `cargo-apk`: Parse and act on commandline arguments before finding and panicking on incomplete workspace

dvc94ch:
Fixes #182
Fixes #217


cargo-apk/src/main.rs
```diff
@@ -1,82 +1,75 @@
...
+        #[clap(flatten)]
+        args: Args,
+    },
+    /// Compiles the current package and creates an apk
+    Build {
+        #[clap(flatten)]
+        args: Args,
+    },
+    /// Run a binary or example of the local package
+    Run {
```
MarijnS95, line 33: Can these have `short` annotations as well? I really crave for `cargo apk r` etc, it's in my workflow :)

cargo-apk/src/apk.rs
```diff
@@ -84,7 +84,7 @@ impl<'a> ApkBuilder<'a> {
             if self.cmd.target().is_none() {
                 cargo.arg("--target").arg(triple);
             }
-            cargo.args(self.cmd.args());
+            self.cmd.args().apply(&mut cargo);
```
MarijnS95, line 87: The `Args` struct in `cargo-subcommand` only holds and applies a limited set of args now, how about (new) arguments that are not (yet) recognized by it? Ie. all flags related to features are not in the struct and hence won't be forwarded to `cargo`.

Perhaps we should keep storing and passing the original set of arguments as-is?

cargo-apk/src/main.rs
```diff
@@ -1,82 +1,75 @@
...
+
+#[derive(Parser)]
+struct Cmd {
+    #[clap(subcommand)]
+    apk: ApkCmd,
+}
+
+#[derive(clap::Subcommand)]
+enum ApkCmd {
+    /// Helps cargo build apk's for android
```
MarijnS95, line 13: I think the plural of `apk` is without apostrophe?

MarijnS95: And `android` is capitalized.

cargo-apk/src/main.rs
```diff
@@ -1,82 +1,75 @@
...
+}
+
+#[derive(clap::Subcommand)]
+enum ApkSubCmd {
+    /// Checks that the current package builds without creating an apk
+    Check {
+        #[clap(flatten)]
+        args: Args,
+    },
+    /// Compiles the current package and creates an apk
```
MarijnS95, line 27: I think all doc comments should be written in first person, ie. `Compile` and `build` without `-s`.

MarijnS95:
@dvc94ch Now that https://github.com/dvc94ch/cargo-subcommand/pull/15 is in would you mind rebasing this?

dvc94ch:
@MarijnS95 not sure it's worth it. I'd recommend giving `x new helloworld && cd helloworld && x run` a try. except ios all platforms should work out of the box, ios should catch up in the next few days now that I got an apple developer subscription. Would appreciate your thoughts on it, and what features you think are missing.

- [0] https://github.com/cloudpeers/x

MarijnS95:
@dvc94ch Without even looking at the code or testing it out, I really dislike your approach of throwing the status quo out and starting over from scratch.

https://xkcd.com/927/

dvc94ch:
it has a clearly different scope to cargo-apk. I don't think I'm throwing the status quo out. It doesn't replace `ndk` or `ndk-glue` or `ndk-context`. It's just a build/packaging tool that knows how to build flutter, rust and flutter/rust applications. I don't think extending the scope of cargo-apk would have been the right approach. It's also likely that cargo-apk still has utility in certain niches, especially if you only care about android game development.

MarijnS95:
In fact, looking at your repo, I see a lot of familiar files. Files I've recently fixed bugs in, or contributed new features to, here in `android-ndk-rs` or `cargo-subcommand`.

I am just uncomfortable maintaining the same thing in two disconnected repos, and as such only feel ready to try out and migrate to `x` if we make a clear deprecation and archival plan for `cargo-apk`, here and now. `ndk-build` can stay if and only if it is used in `x`, otherwise I see no reason to keep the maintenance burden up for that too.

(Fwiw why bring ndk/ndk-glue/ndk-context up, when this issue is clearly only about cargo-apk/ndk-build/cargo-subcommand?)

dvc94ch:
> Files I've recently fixed bugs in, or contributed new features to, here in android-ndk-rs or cargo-subcommand.

I'm aware of that, and I appreciate it.

> I see a lot of familiar files

before the renewed effort to maintain cargo-subcommand I did copy some parts. it shares very little or no code with cargo-apk or ndk-build other than some manifest structs `apk/src/manifest.rs`. that's because it takes a different approach. it implements aapt/apksigner/zipalign in rust more or less. A lot of care was taken to avoid dependencies on the environment, so I decided (maybe naively so) that implementing it in rust would mean that `cargo install` would just work on all platforms.

I really don't want to force you to use x if you don't like it :)

dvc94ch:
> Fwiw why bring ndk/ndk-glue/ndk-context up, when this issue is clearly only about cargo-apk/ndk-build/cargo-subcommand?

just for clarification, no particular reason.

anyway, I was asking about how to move forward, I'm happy to factor out shared components to decrease the maintainance burden or whatever. for various reasons it happened to turn out like this (it starting out as a research project). now there is a what I believe is a useful tool which I spent 3 months building. but I'm also fine going our separate ways.

MarijnS95:
> anyway, I was asking about how to move forward, I'm happy to factor out shared components to decrease the maintainance burden or whatever. for various reasons it happened to turn out like this (it starting out as a research project). now there is a what I believe is a useful tool which I spent 3 months building. but I'm also fine going our separate ways.

Fair enough, this had to start somewhere. I don't mind factoring out shared components as long `x` fully supersedes `cargo-apk` (and I'd love for `cargo-ndk` features to be possible with it too), and we deprecate it as a consequence. But do read on.

> it implements aapt/apksigner/zipalign in rust more or less

I'm very happy about this. Setting those up and having a Java dependency (for at least `keytool`, but that isn't needed too often) has been a pain, while we want APK building to be as painless (and with the least possible dependencies outside of `cargo`) as possible.

> I really don't want to force you to use x if you don't like it :)

I _want_ to use `x` if it turns out to be good. What I absolutely don't want is for `cargo-apk+ndk-build` and `x` to coexist with significant enough feature overlap to compete, that's a waste of our development time. As such, I think to proceed we should:

1. I (and hopefully I'm not alone) should check out your tool and evaluate how much all the extra automation improves my workflow;
2. At the same time, evaluate what we're missing in contrast to `cargo-apk`, or just in general;
3. Set up a clear deprecation plan for `cargo-apk`. I don't want to maintain it and I don't think others should either, they should instead contribute to `x`;
4. I can perhaps take some of the burden of porting the bugfixes and features from here to `x`;
5. We deprecate and remove `cargo-apk` and `ndk-build` (caveat: we can also decide to keep `ndk-build` and slim it down to be useful for `x` and others).

I'm then not sure what to do with `cargo-subcommand`. Perhaps we need that crate pulled apart to have a separate library for the manifest format and parsing (perhaps including `.cargo/config.toml`). Some other tools already have/do that, doesn't a separate crate like it exist already?

dvc94ch:
> 1. I (and hopefully I'm not alone) should check out your tool and evaluate how much all the extra automation improves my workflow;

that would be the first step, really looking forward to get your feedback on this.

> 2. At the same time, evaluate what we're missing in contrast to cargo-apk, or just in general;

There are a couple of differences between `cargo-apk` and `x --platform android`:

2.1 `cargo-apk` uses readelf to automatically add third party libraries to apk. We could generalize it using llvm-objdump to make it work cross platform.

2.2 C++ has not been tested, think it's just a matter of including the libc++

2.3 cmake has not been tested, but think we agreed previously that cmake should fix any issues instead of us working around it

2.4 manifest is parsed from the `manifest.yml` `android: manifest: ` section. this is to enable the same workflow in flutter, rust and flutter/rust apps. should be mostly compatible (superset of existing manifest features)

2.5 ndk-debug is not supported. I have a work in progress `x lldb` command which should be cross platform. this is on my todo list.

> 3. Set up a clear deprecation plan for cargo-apk. I don't want to maintain it and I don't think others should either, they should instead contribute to x;
> 4. I can perhaps take some of the burden of porting the bugfixes and features from here to x;

That would be cool, but let's start with 1 first :)

> 5. We deprecate and remove cargo-apk and ndk-build (caveat: we can also decide to keep ndk-build and slim it down to be useful for x and others).

By removing the reliance on shell/python scripts spread out through the ndk and doing everything in rust, ndk-build is not really that useful anymore. Instead there is a bit of code to handle each platform in various places without requiring much special handling for android. You can simply call `cargo.use_ndk(path: &Path)` or `cargo.use_xwin(path: &Path)` or cargo.use_ios_sdk(path: &Path)`.

MarijnS95:
Reopening and resolving conflicts as I really like to get this cleanup/improvement in. Will re-review and dogfood this on my own working env for a while.

Cargo.toml
```diff
@@ -9,3 +9,6 @@ members = [
     "ndk-sys",
     "cargo-apk",
 ]
+
+[patch.crates-io]
+cargo-subcommand = { git = "https://github.com/dvc94ch/cargo-subcommand", branch = "master" }
```
MarijnS95, line 14: Will need to make a new `cargo-subcommand` release before we get this in - but give it some time to finalize this first.

cargo-apk/src/main.rs
```diff
@@ -1,112 +1,96 @@
...
+        ApkSubCmd::Run { args, no_logcat } => {
+            let cmd = Subcommand::new(args.subcommand_args)?;
+            let builder = ApkBuilder::from_subcommand(&cmd, args.device)?;
             anyhow::ensure!(cmd.artifacts().len() == 1, Error::invalid_args());
-            builder.run(&cmd.artifacts()[0])?;
-        }
-        "--" => {
-            builder.default()?;
+            builder.run(&cmd.artifacts()[0], no_logcat)?;
         }
```
MarijnS95, line 84: Let's see if I can add back support for this, as it was quite useful.

MarijnS95: It is sort of working, but getting unknown arguments (outside of `cargo_subcommand::Args`) parsed properly is tricky and left out for now.

msiglreith (approved):
lgtm, checked it locally and functionality seems to be all there

MarijnS95:
I'll push out a `cargo-subcommand` release tomorrow, and merge this while test-driving it locally for a few days. Will then probably release `0.10.0` or `0.9.5` some time next week.

cargo-apk/src/main.rs
MarijnS95: Finally fixed/implemented in https://github.com/rust-windowing/android-ndk-rs/pull/363!

PR #239, opened 2022-03-12
Fix provided libs hashset containing those missing on-device

zer0def:
A small bugfix to a bug introduced with #187, as `libc++_shared.so` is added into the `provided` hashset when crawling Android NDK search paths, which are assumed present on-device, but this particular case is neither on-device, nor ends up included as an artifact when bundling needed libraries.


dvc94ch (approved):
Thanks

MarijnS95:
Thanks, though I'd appreciate if the same comment from below would have been replicated here. OTOH this code can probably use some refactoring again to be simplified and not have an edgecase for `libc++_shared.so` twice.

PR #240, opened 2022-03-13
Initial jni support.

dvc94ch:
Spent a few hours today looking into rewriting `jni-bindgen/jni-android-sys` as besides not working with `cargo-apk` and being unmaintained, the code is overly complicated and messy. Not really sure if it's worth doing at all, as the jni crate seems easy enough to use, but here it is.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,8 @@
 [workspace]
 members = [
+    "android-sys",
+    "cargo-apk",
+    "jni-bindgen",
     "ndk",
     "ndk-context",
     "ndk-macro",
@@ -7,5 +10,4 @@ members = [
     "ndk-examples",
     "ndk-glue",
     "ndk-sys",
-    "cargo-apk",
 ]
--- a/android-sys/.gitignore
+++ b/android-sys/.gitignore
@@ -1,0 +1,1 @@
+src/lib.rs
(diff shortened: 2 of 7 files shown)


zer0def:
Perhaps I could chime in with a use case I'm exploring, as I'm not exactly sure where the dividing lines are.

I'm currently looking into bridging the gap across Qt-Android development using [qmetaobject-rs](https://github.com/woboq/qmetaobject-rs) and, as far as I can tell at the time of writing, the two main ways of doing so are:
- by completely ignoring `android-ndk-rs` by trying to shove the produced `.so` into `androiddeployqt`, similarly to [qt-android-cmake](https://github.com/LaurentGomila/qt-android-cmake)
- by building through `android-ndk-rs`, which results in segfaults, because Qt expects several JNI-related event loop globals [relating to](https://github.com/qt/qtbase/blob/dev/src/corelib/kernel/qjnihelpers.cpp#L220) it's own [JNI class](https://github.com/qt/qtbase/blob/dev/src/android/jar/src/org/qtproject/qt/android/QtNative.java) to be initialized through [`JNI_OnLoad`](https://github.com/qt/qtbase/blob/dev/src/corelib/kernel/qjnihelpers.cpp#L381) → [`initJNI`](https://github.com/qt/qtbase/blob/dev/src/corelib/kernel/qjnihelpers.cpp#L216)

In the latter case, I'm not sure whether it should be (and if so, how) something I should be hijacking in the native activity at the first chance I get (assuming it's not already too late) to hopefully populate, so the proper application itself could limp along.
I'm also not entirely certain this development is related to the presented use case at all, but it's definitely something that piqued my interest.

PR #245, opened 2022-03-16
ndk-sys,ndk: Turn type aliases into newtype wrappers and fix mismatches

MarijnS95:
~Depends on #248~

Some stray `camera_status_t` showed up in the safe media bindings, which is not caught as a mismatch to `media_status_t` unless both are a newtype.

~TODO: `aaudio_result_t` remains a typedef, it is perhaps defined differently in the headers and confusing `bindgen`?~ `aaudio_result_t` is defined as a `typedef` in the C headers, not as `enum` :/


ndk/src/media/error.rs
```diff
@@ -4,77 +4,77 @@ use thiserror::Error;
 #[repr(i32)]
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum MediaErrorResult {
```
MarijnS95, line 6: Should we just drop this type and use/reexport `ffi::media_status_t` directly?

ndk/src/media/error.rs
```diff
@@ -4,77 +4,77 @@ use thiserror::Error;
...
-            ffi::media_status_t_AMEDIA_DRM_VERIFY_FAILED => DrmVerifyFailed,
-            ffi::media_status_t_AMEDIA_DRM_NEED_KEY => DrmNeedKey,
-            ffi::media_status_t_AMEDIA_DRM_LICENSE_EXPIRED => DrmLicenseExpired,
-            ffi::media_status_t_AMEDIA_IMGREADER_ERROR_BASE => ImgreaderErrorBase,
-            ffi::media_status_t_AMEDIA_IMGREADER_NO_BUFFER_AVAILABLE => ImgreaderNoBufferAvailable,
-            ffi::media_status_t_AMEDIA_IMGREADER_MAX_IMAGES_ACQUIRED => ImgreaderMaxImagesAcquired,
-            ffi::media_status_t_AMEDIA_IMGREADER_CANNOT_LOCK_IMAGE => ImgreaderCannotLockImage,
-            ffi::media_status_t_AMEDIA_IMGREADER_CANNOT_UNLOCK_IMAGE => ImgreaderCannotUnlockImage,
-            ffi::media_status_t_AMEDIA_IMGREADER_IMAGE_NOT_LOCKED => ImgreaderImageNotLocked,
+            ffi::media_status_t::AMEDIACODEC_ERROR_RECLAIMED => CodecErrorReclaimed,
```
MarijnS95, line 52: All these conversions are a bit odd too, `MediaErrorResult` appears to be `repr(i32)` for a reason though we don't use it to our advantage. Yes, casting unknown values is UB so perhaps we should just use `ffi::media_status_t` directly as suggested above?

dvc94ch (approved):

PR #248, opened 2022-03-22
media/media_codec: Replace `construct_never_null` and infallible `unwrap()` pattern with something more sensible

MarijnS95:
These functions return a pointer directly instead of an error code with the pointer as argument, making them unsuitable for use with `construct_never_null`.  Because there is no error code, returning a manual success value still requires an infallible `unwrap()` call to unpack the `Result`.

Replace this with a new helper function that does not unnecessarily wrap the resulting pointer in a `Result` when there is no result code to base this off, and match the signature to a function that returns the pointer directly.

At the same time reduce the size of some `unsafe {}` blocks to wrap just what is unsafe.

### DIFF
--- a/ndk/src/media/media_codec.rs
+++ b/ndk/src/media/media_codec.rs
@@ -1,4 +1,4 @@
-use super::{construct_never_null, NdkMediaError, Result};
+use super::{construct_never_null, get_never_null, NdkMediaError, Result};
 use crate::native_window::NativeWindow;
 use std::{
     convert::TryInto,
@@ -40,7 +40,7 @@ impl MediaFormat {
 
     pub fn new() -> Self {
         Self {
-            inner: unsafe { NonNull::new(ffi::AMediaFormat_new()).unwrap() },
+            inner: NonNull::new(unsafe { ffi::AMediaFormat_new() }).unwrap(),
         }
     }
 
@@ -88,7 +88,7 @@ impl MediaFormat {
         let name = CString::new(key).unwrap();
         let mut out = ptr::null();
         unsafe { ffi::AMediaFormat_getString(self.as_ptr(), name.as_ptr(), &mut out) }
-            .then(|| unsafe { CStr::from_ptr(out).to_str().unwrap() })
+            .then(|| unsafe { CStr::from_ptr(out) }.to_str().unwrap())
     }
 
     pub fn set_i32(&self, key: &str, value: i32) {
@@ -193,25 +193,25 @@ impl MediaCodec {
     pub fn from_codec_name(name: &str) -> Option<Self> {
         let c_string = CString::new(name).unwrap();
         Some(Self {
-            inner: unsafe { NonNull::new(ffi::AMediaCodec_createCodecByName(c_string.as_ptr()))? },
+            inner: NonNull::new(unsafe { ffi::AMediaCodec_createCodecByName(c_string.as_ptr()) })?,
         })
     }
 
     pub fn from_decoder_type(mime_type: &str) -> Option<Self> {
         let c_string = CString::new(mime_type).unwrap();
         Some(Self {
-            inner: unsafe {
-                NonNull::new(ffi::AMediaCodec_createDecoderByType(c_string.as_ptr()))?
-            },
+            inner: NonNull::new(unsafe {
+                ffi::AMediaCodec_createDecoderByType(c_string.as_ptr())
+            })?,
         })
     }
 
     pub fn from_encoder_type(mime_type: &str) -> Option<Self> {
         let c_string = CString::new(mime_type).unwrap();
         Some(Self {
-            inner: unsafe {
-                NonNull::new(ffi::AMediaCodec_createEncoderByType(c_string.as_ptr()))?
-            },
+            inner: NonNull::new(unsafe {
+                ffi::AMediaCodec_createEncoderByType(c_string.as_ptr())
+            })?,
         })
     }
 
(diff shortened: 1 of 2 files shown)


MarijnS95:
@zarik5 I trust you have checked all these calls wrapped in `*_never_null` to never return `null` for the given input arguments (ie. they are always valid)? In release `NonNull::new_unchecked` is used, which is UB if there are valid cases where these functions return `null` (ie. I hope you've not picked this function for convenience only).

zarik5:
@MarijnS95 Well no, I didn't check all functions. In general the MediaCodec API requires the user to know what they're doing, because if you call some functions in the wrong context it will probably lead to a crash. For example it's invalid to call `AMediaCodec_getOutputBuffer()` in encoder mode and `AMediaCodec_getInputBuffer()` in decoder mode, unless maybe the surface supports host memory transfer. It's too check all valid usages.

MarijnS95:
> For example it's invalid to call `AMediaCodec_getOutputBuffer()` in encoder mode and `AMediaCodec_getInputBuffer()` in decoder mode

If you want the caller to uphold "undocumented" (I don't see it anywhere in `media_codec.rs`...) constraints such public should be marked `unsafe`, fwiw.

zarik5:
For when the function returns a pointer this makes sense. But specifically for `AMediaCodec_getOutputBuffer` and `AMediaCodec_getInputBuffer` I don't think we can intercept the invalid usage before a crash. More testing is definitely needed.

zarik5:
> If you want the caller to uphold "undocumented" (I don't see it anywhere in media_codec.rs...) constraints such public should be marked unsafe, fwiw.

I'm fine with this.

MarijnS95:
> For when the function returns a pointer this makes sense. But specifically for `AMediaCodec_getOutputBuffer` and `AMediaCodec_getInputBuffer` I don't think we can intercept the invalid usage before a crash. More testing is definitely needed.

Would it crash in the native function or just **correctly** return a `null` value that we should subsequently propagate to the caller with an `Option`?

> > If you want the caller to uphold "undocumented" (I don't see it anywhere in media_codec.rs...) constraints such public should be marked unsafe, fwiw.
> 
> I'm fine with this.

Note again that this should only be necessary if we can't do anything else about it. Ie. if the function just returns `null` on invalid input/state we're better off returning an `Option`/`Result`.

zarik5:
> Would it crash in the native function or just correctly return a null value that we should subsequently propagate to the caller with an Option?

The the thing is I don't know. I'm afraid it might crash inside the native function. I tested only the decoder path, with the proper usage, and at the moment I put that code pause to work on other things. I might try to do more testing, but it will be still be limited. This is more a problem with the (loack of) official documentation.

dvc94ch (approved):

MarijnS95:
Merged this PR for its initial cleanup, @zarik5 one of us should really see about fallibility so that we can fix up the API _before_ the next release.

zarik5:
Unfortunately I can't allocate more time on this. The only help i can provide is [this code](https://github.com/alvr-org/ALVR/blob/master/experiments/client/src/video_decoder/mediacodec.rs), which has been tested to some degree but not fully. I started working (for a job) on other fronts.

MarijnS95:
As such I'm not sure if we should back out of this change entirely, or mark the module private until safety concerns are resolved. Alternatively one could preemptively change the necessary interfaces to return `Option`, or "hide" those interfaces behind a `"possibly-unsafe"` feature flag until we're sure whether the interfaces can return `null`. For example:

- All `construct_never_null` usages seem fine: those functions return an error code and Android better not returns success without assigning to the pointer;
- `AMediaCodec_getOutputBuffer` and `AMediaCodec_getInputBuffer` also seem fine: your Rust wrapper enforces them to only be callable with an argument from, and after a successful result from `AMediaCodec_dequeueOutputBuffer` / `AMediaCodec_dequeueInputBuffer` which I presume is the main reason why those functions don't otherwise have a valid return code.
  It is however unknown when these indices "expire" or how/when the user has to recycle them? I doubt Android keeps buffers for dequeued indices around, that'd be a huge memory leak;
- Same concerns for `AMediaCodec_getBufferFormat` regarding expiry of the index, but otherwise seems like there's no chance to have a buffer without associated format?
- Then `AMediaCodec_getInputFormat` and `AMediaCodec_getOutputFormat` remain, not sure what to do with those.

FWIW I think we're also fine removing `NonNull::new_unchecked` from `get_never_null` and keeping this a panic under `!cfg!(debug_assertions)` conditions.

MarijnS95:
@zarik5 Perhaps if you do manage to find time at some point, would you mind porting that code to a minimal example for inclusion in `ndk-examples`? That ought to help users and maintainers out, thanks!

zarik5:
A "possibly-unsafe" feature is ugly but I don't have any nicer alternative. Removing the whole module would be a loss since it would work fine if used correctly. 

I found [this crate](https://github.com/hardebayho/rust_mediacodec) which wraps the mediacodec API, which seems to have much more extensive checking. Maybe a collaboration is possible? Currently it's not nice to use with ndk (the crate in this repo) since it reimplements all wrappers for pointers like ANativeWindow etc.

zarik5:
I'll open an issue on their repo

MarijnS95:
@zarik5 Good call. Sharing their `media_codec` code is one thing, but it's also a little unfortunate to have a duplicate wrapper around other NDK bits. Unifying it would definitely be good, though that repo seems more like someone publishing their personal project than a crate intended for upstream use.

PR #251, opened 2022-03-23
README: Elaborate quick-start guide to be more clear for new Rust users

### LINKED ISSUE #250: [bug] didnt find Cargo.toml
i have this project

image

lib.rs
```rust
#[cfg_attr(target_os = "android", ndk_glue::main(backtrace = "on"))]
pub fn main() {
    println!("hello world");
}
```

main.rs
```rust
fn main() {
    $crate::main();
}
```

i have installed `cargo-apk` with `cargo install cargo-apk`

when im trying to run it with:
```shell
cargo apk run
```

it says `Error: Didn't find Cargo.toml.`

ummm. what im i doing wrong ?

Cargo.toml is definitely right there!

and btw, how do i install `ndk` (`#[cfg_attr(target_os = "android", ndk_glue::main(backtrace = "on"))]`) or `its built-in` ?

MarijnS95:
Closes #250

Some Rust users seem to take the plunge in deep Android waters directly, without familiarizing themselves with a "normal desktop" crate setup first.  We can accomodate those users by being a bit more clear how to set up a new crate and how to make a crate run on both desktop/PC and Android, while at the same time being short enough for well versed rustaceans to skip over.


README.md
```diff
@@ -71,17 +108,31 @@ We can now directly execute our `Hello World` application on a real connected de
...
+$ cargo apk run
+error: extra arguments to `rustc` can only be passed to one target, consider filtering
+the package by passing, e.g., `--lib` or `--bin NAME` to specify a single target
+Error: Command `cargo rustc --target aarch64-linux-android -- -L hello_world_android/target/cargo-apk-temp-extra-link-libraries` had a non-zero exit code.
+```
+
+To solve this, add `--lib` to the run invocation, like so:
+```console
+$ cargo apk run --lib
+```
```
MarijnS95, line 123: I haven't investigated this properly yet, but perhaps we can force-add `--lib` to our `rustc` invocation just in case?

Additionally, running `cargo apk run --bin(s, or the name of a binary target)` here results in a successful cargo build but finally fails packaging:

```
Error: Path `"hello_world_android/target/aarch64-linux-android/debug/libhello_world_android.so"` doesn't exist.
```

I think we should assert/error in `cargo-apk` when the artifacts from `cargo-subcommand` include non-library targets, but I'm not sure if we can distinguish them at this point.

README.md
```diff
@@ -71,17 +108,39 @@ We can now directly execute our `Hello World` application on a real connected de
...
+
+### 5. Inspect the output
+
+`ndk-glue` redirects stdout and stderr to Android `logcat`, including the `println!("Hello World")` by the example above. See [Logging and stdout](##Logging-and-stdout) below how to access it.
+
+## Rendering to the window
+
+Android native apps have no easy access to [Android's User Interface](https://developer.android.com/guide/topics/ui) functionality (bar [JNI](##jni) interop). Applications can instead draw pixels directly to the window using [`ANativeWindow_lock`](https://developer.android.com/ndk/reference/group/a-native-window#group___a_native_window_1ga0b0e3b7d442dee83e1a1b42e5b0caee6), or use a graphics API like OpenGL or Vulkan for high performance rendering.
```
MarijnS95, line 131: This should perhaps be expanded and added in a separate PR. I need `ANativeWindow_lock` and will probably PR a safe wrapper for it soon which we should reference instead, and this should be elaborated to at least mention `winit` for the activity lifecycle and event handling (we're in the https://github.com/rust-windowing organisation after all...). I'm sure there are some GUI crates available like flutter which (a quick-start guide?) we might reference here.

CC @dvc94ch

dvc94ch (approved):
lgtm

README.md
dvc94ch: flutter requires a different tool. don't see `cargo apk` in its current form being too helpful here.

README.md
MarijnS95: Not going to worry about this for now, behaviour will probably be revamped/improved in `xbuild`.

PR #256, opened 2022-03-28
media_codec: Uphold non-`null` guarantees in some erratic scenarios

MarijnS95:
CC @zarik5, this is the last of things I think we need for `media_codec` to be stable enough?

The Android documentation does not comment on when functions could possibly return `null`, and while discussion in [#248] finds these unlikely enough to be communicated back to the user there is no 100% guarantee and the resulting pointers should as such always be checked (either at all, or also when `cfg!(debug_assertions)` happens to be disabled).

Note that this `NonNull::new(..).unwrap()` pattern is found in many other places in the `ndk`.

[#248]: https://github.com/rust-windowing/android-ndk-rs/pull/248

### DIFF
--- a/ndk/src/media/media_codec.rs
+++ b/ndk/src/media/media_codec.rs
@@ -1,4 +1,4 @@
-use super::{construct_never_null, get_never_null, NdkMediaError, Result};
+use super::{get_unlikely_to_be_null, NdkMediaError, Result};
 use crate::native_window::NativeWindow;
 use std::{
     convert::TryInto,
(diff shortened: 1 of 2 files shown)


MarijnS95:
@zarik5 Thoughts?

zarik5:
Again, I don't have deep knowledge about NDK usage, and I don't have time to do test every scenario. You could work together with @Hardebayho to define a safer API.

zarik5:
@MarijnS95 Alright, I'm ok with removing the Mediacodec wrapper at this point.

MarijnS95:
In any case this PR should address the final safety concerns without altering public API for the time being, after which I think I'm comfortable publishing an `ndk` update _with_ the wrapper.

zarik5:
The assertions and .expect() will make the API safer, but it will not prevent a crash on the NDK side, which I did not verify if that is possible or not.

PR #258, opened 2022-03-30
cargo-apk,ndk-build: Move `ndk-gdb` setup to `ndk-build` and support Windows

MarijnS95:
Depends on #257

Just a draft for now as I haven't really tested this on Windows yet - but the missing `.cmd` is obvious enough, and this command is probably useful to expose from the `ndk-build` library anyway.


MarijnS95:
Tested on Windows and Linux.

PR #263, opened 2022-04-19
ndk-glue,ndk-context: Remove `AndroidContext` when activity is destroyed

### LINKED ISSUE #261: Native Activity finish does not destroy
I'm working on an android application which has a native activity that runs rust. The rust side uses ndk glue, ndk context, and winit. Loading it up works fine and as expected, but when I try to close the native activity and open it again the application crashes. I noticed that when finish is called on the native activity, only the suspend event fires on the rust side even though onDestroy happens in the native activity (android side).

This is the output I get when trying to open the native activity again:
```
2022-04-16 21:52:46.396 30803-30803/rs.ruffle W/ActivityThread: handleWindowVisibility: no activity for token android.os.BinderProxy@1e9fde2
2022-04-16 21:52:46.407 30803-30803/rs.ruffle E/ruffle: log_panics: thread '<unnamed>' panicked at 'assertion failed: previous.is_none()': C:\Users\oxter\.cargo\registry\src\github.com-1ecc6299db9ec823\ndk-context-0.1.0\src\lib.rs:87
2022-04-16 21:52:46.407 30803-30908/rs.ruffle I/RustStdoutStderr: fatal runtime error: failed to initiate panic, error 5
2022-04-16 21:52:46.407 30803-30803/rs.ruffle A/libc: Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 30803 (rs.ruffle), pid 30803 (rs.ruffle)
```

Here's a link to the [repository](https://github.com/oxters168/ruffle-android/tree/AndroidOverlay). Any help would be appreciated, thanks.

MarijnS95:
Fixes #261

Android runs multiple activities of the same app inside the same process, and shutting down followed by starting a `NativeActivity` inside the same process (that's been kept alive by another activity) causes the `static ANDROID_CONTEXT` in `ndk-context` to retain its value.  This value should be cleared to prevent leaking possibly invalid pointers and ultimately crashing in `assert!(previous.is_none());` when a native activity is restarted within that process.

Note that this still doesn't allow for multiple `NativeActivity` instances to be started concurrently.

### DIFF
--- a/ndk-context/src/lib.rs
+++ b/ndk-context/src/lib.rs
@@ -86,3 +86,15 @@ pub unsafe fn initialize_android_context(java_vm: *mut c_void, context_jobject: 
     });
     assert!(previous.is_none());
 }
+
+/// Removes the [`AndroidContext`]. It is released by [__ndk-glue__](https://crates.io/crates/ndk-glue)
+/// when the activity is finished and destroyed.
+///
+/// # Safety
+///
+/// This function must only be called after [`initialize_android_context()`],
+/// when the activity is subsequently destroyed according to Android.
+pub unsafe fn release_android_context() {
+    let previous = ANDROID_CONTEXT.take();
+    assert!(previous.is_some());
+}
--- a/ndk-context/CHANGELOG.md
+++ b/ndk-context/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Add `release_android_context()` function to remove `AndroidContext` when activity is destroyed. (#263)
+
 # 0.1.0 (2022-02-14)
 
 - Initial release! 🎉
--- a/ndk-glue/CHANGELOG.md
+++ b/ndk-glue/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Call `ndk_context::release_android_context()` function to remove `AndroidContext` when activity is destroyed. (#263)
+
 # 0.6.1 (2022-02-14)
 
 - Initialize `ndk-context` for cross-version access to the Java `VM` and Android `Context`.
--- a/ndk-glue/src/lib.rs
+++ b/ndk-glue/src/lib.rs
@@ -251,6 +251,7 @@ unsafe extern "C" fn on_stop(activity: *mut ANativeActivity) {
 
 unsafe extern "C" fn on_destroy(activity: *mut ANativeActivity) {
     wake(activity, Event::Destroy);
+    ndk_context::release_android_context();
 }
 
 unsafe extern "C" fn on_configuration_changed(activity: *mut ANativeActivity) {


MarijnS95:
Note that this will have to be a nonbreaking patch release of `ndk-context`, whose use is supposed to be backported to `ndk-glue` `0.4(.2)` and `0.5(.2)`, and we can safely release `ndk-glue 0.6.2` it seems.

ndk-context/src/lib.rs
```diff
@@ -86,3 +86,15 @@ pub unsafe fn initialize_android_context(java_vm: *mut c_void, context_jobject:
...
 }
+
+/// Removes the [`AndroidContext`]. It is destroyed by [__ndk-glue__](https://crates.io/crates/ndk-glue)
+/// when the activity is finished and destroyed.
+///
+/// # Safety
+///
+/// This function must only be called after [`initialize_android_context()`] when the activity
+/// is subsequently destroyed.
+pub unsafe fn reset_android_context() {
```
dvc94ch, line 97: how about `destroy_android_context`? I think it better mirrors the naming used by android and sounds more like the inverse of initialize to me than reset.

dvc94ch (approved):
Thanks!

ndk-context/src/lib.rs
MarijnS95: That was my first thought, but this function doesn't really "destroy" anything as much as it "gives up" some pointer references to things. It is a better match for `on_destroy` indeed. Perhaps `release` neatly complements `initialize`?

dvc94ch:
> Note that this will have to be a nonbreaking patch release of ndk-context

sounds good

> whose use is supposed to be backported to ndk-glue 0.4(.2) and 0.5(.2), and we can safely release ndk-glue 0.6.2 it seems.

Are many people still using 0.4 and 0.5?

ndk-context/src/lib.rs
dvc94ch: dropping dangling pointers seems to be destroying something, but ok :man_shrugging:

MarijnS95: We're "destroying" the 8 bytes of pointer data, but not the pointee :P

dvc94ch:
I'd say until the next winit release backporting to 0.5 makes sense. but can't see any notable crates using 0.4 here https://crates.io/crates/ndk-glue/reverse_dependencies

MarijnS95:
IIRC someone was still using it on 0.4, and the `0.4.1` backport was mainly to showcase that at least these 3 versions are now all "compatible" - ie. `ndk-context` does what we intend.

MarijnS95:
Added changelog, I'll try to get some releases out for this after merging.

MarijnS95:
@dvc94ch I think the VM is probably valid, but the context pointer can definitely change? I'm wary on the context since we're using the activity pointer here because activity inherits from context, but it should also be possible to extract a "base" context which perhaps remains the same for every activity?

Also, yes, as mentioned in the PR description `ndk-context` is problematic with multiple `NativeActivity` instances too (ie. when configuring to launch another activity instead of reusing the open one, something with task stacks).

MarijnS95:
Thinking about it, `fn native_window()` and `fn input_queue()` don't even have a doc comment and should probably be marked `unsafe` to communicate this implicit synchronization behaviour. Oh my :exploding_head:

dvc94ch:
well if you're just using ndk-context, there is no lock. but thinking about this a bit more, it's probably not that serious, since hopefully all threads spawned by the activity are killed too after destroy and before start.

don't think `native_window` and `input_queue` need to be unsafe, unsafety does not refer to deadlocks or memory leaks, both behaviours are considered safe.

dvc94ch:
ah I see what you mean, but the unsafety happens when dereferencing a raw pointer. so you call native_window, get a raw ptr, that's fine even if it becomes dangling. if you then pass that raw pointer to some graphics api, the api taking the raw pointer should be marked unsafe and state that the pointer needs to be valid for lifetime 'a.

MarijnS95:
Agreed, though the problematic case is usually more insidious than that. You won't be dereferencing invalid pointers directly but using ie. a prior created Vulkan swapchain that - through a Vulkan surface - references a `NativeWindow` which is being destroyed. Like I said, bit far-fetched for `unsafe`.

Fortunately all Ash's functions are `unsafe` by default even if only appearing to handle "safe" data, for exactly this reason :) - it's just hard to know about it sometimes but that's the nature of low-level systems programming and FFI :)

PR #267, opened 2022-04-23
Add `SurfaceTexture` bindings

### LINKED ISSUE #262: SurfaceTexture NDK bindings
~Are there plans to support OpenGL?~

lattice0:
Support for ASurfaceTexture from NDK

Closes #262


ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,72 @@
...
+        unsafe { ffi::ASurfaceTexture_getTimestamp(self.ptr.as_ptr()) }
+    }
+
+    pub fn from_surface_texture(env: *mut JNIEnv, surface_texture: jobject) -> Option<Self> {
+        let a_surface_texture_ptr =
+            unsafe { ffi::ASurfaceTexture_fromSurfaceTexture(env, surface_texture) };
+        let s = NonNull::new(a_surface_texture_ptr)?;
+        Some(unsafe { SurfaceTexture::from_ptr(s) })
+    }
+}
```
dvc94ch, line 72: some doc comments would be nice. guess you can copy from official docs

dvc94ch (commented):
looks good, I assume @MarijnS95 will have some comments too

lattice0:
I'm not sure about 

```
unsafe impl Send for SurfaceTexture {}
unsafe impl Sync for SurfaceTexture {}
```

though, have to think about it

dvc94ch:
good point, it shouldn't be Sync as it will get released mutliple times on drop. Send is fine.

lattice0:
Shouldn't clippy complain about `unsafe impl` without justifications?

Anyways, what about `Send`? As far as I know, the `SurfaceTexture` is created in the main thread by flutter's method channel. On java, stuff from Activity and etc, usually require you to call things from the main thread on it. I know almost nothing about `SurfaceTexture` but I remember having a hard time before when trying to modify Activity related objects from other threads.

JasperDeSutter:
This is missing api level guards, which I think should be 28 for all methods. Perhaps also a general "surface-texture" feature flag can be added which enables this module?

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,85 @@
...
+        } else {
+            Err(r)
+        }
+    }
+
+    /// # Safety
+    ///
+    /// This function should be called by a healthy JVM pointer and with a non null surface texture,
+    /// which must be kept alive on the java/kotlin side.
+    pub unsafe fn from_surface_texture(env: *mut JNIEnv, surface_texture: jobject) -> Option<Self> {
```
MarijnS95, line 80: I feel like such constructors should be written at the top of the `impl` block, together with `from_ptr`.

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,85 @@
...
+            Ok(())
+        } else {
+            Err(r)
+        }
+    }
+
+    /// # Safety
+    ///
+    /// This function should be called by a healthy JVM pointer and with a non null surface texture,
+    /// which must be kept alive on the java/kotlin side.
```
MarijnS95, line 79:
```suggestion
    /// This function should be called with a healthy JVM pointer and with a non-null surface texture,
    /// which must be kept alive on the Java/Kotlin side.
```

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,85 @@
+//! Bindings for [`ffi::ASurfaceTexture`]
+use crate::native_window::NativeWindow;
+use jni_sys::jobject;
+use jni_sys::JNIEnv;
```
MarijnS95, line 4: Feel free to merge.

MarijnS95 (changes requested):
> looks good, I assume @MarijnS95 will have some comments too

Looks like yall covered most of it, and I do agree with all the comments that are still outstanding.

MarijnS95:
> good point, it shouldn't be Sync as it will get released mutliple times on drop. Send is fine.

`Sync` implies that you can have multiple _references_ to an object: references are not ownership, hence cannot drop the object multiple times.

This only becomes a problem if we allow safely cloning/copying the `SurfaceTexture` (implying multiple ownership) which is not possible for this since there are no reference counting functions in the NDK (like we have in `NativeWindow`).
(Note: you can `unsafe`ly alias it by calling `from_ptr(other.ptr())`)

As such, we should implement:
- `Send`, if it's save to hand over this structure from one thread to another: I don't know if this is true for GL contexts as those are afaik also bound to a thread if I remember correctly. That may or may not impose a limitation here;
- `Sync` - presuming we can implement `Send` - if the NDK allows calling these function from multiple threads, concurrently.

It does not seem like the NDK docs clarify any of this, so should perhaps look at `Surface` and if any of this is allowed on the Java side? How would that work?

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,85 @@
...
+        self.ptr
+    }
+
+    pub fn acquire_native_window(&mut self) -> Option<NativeWindow> {
+        let native_window = unsafe { ffi::ASurfaceTexture_acquireANativeWindow(self.ptr.as_ptr()) };
+        let n = NonNull::new(native_window)?;
+        Some(unsafe { NativeWindow::from_ptr(n) })
+    }
+
+    pub fn attach_to_gl_context(&self, tex_name: u32) -> Result<(), i32> {
```
MarijnS95, line 39: All the `i32` results are POSIX error codes according to the documentation. We should perhaps check if the `ndk` already has an appropriate error wrapper for that or add one - even just wrapping it in a new `struct PosixError(pub i32)` seems fine by me.

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub fn attach_to_gl_context(&self, tex_name: u32) -> Result<(), PosixError> {
+        let r = unsafe { ffi::ASurfaceTexture_attachToGLContext(self.ptr.as_ptr(), tex_name) };
+        if r == 0 {
+            Ok(())
+        } else {
+            Err(PosixError(r))
+        }
+    }
+
+    pub fn detach_from_gl_context(&self) -> Result<(), i32> {
```
MarijnS95, line 51:
```suggestion
    pub fn detach_from_gl_context(&self) -> Result<(), PosixError> {
```

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,88 @@
...
+        let mut r = [0f32; 16];
+        unsafe { ffi::ASurfaceTexture_getTransformMatrix(self.ptr.as_ptr(), r.as_mut_ptr()) };
+        r
+    }
+
+    pub fn timestamp(&self) -> i64 {
+        unsafe { ffi::ASurfaceTexture_getTimestamp(self.ptr.as_ptr()) }
+    }
+
+    pub fn update_tex_image(&self) -> Result<(), i32> {
```
MarijnS95, line 70: Is this also a posix error?

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,88 @@
...
+
+#[derive(Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct SurfaceTexture {
+    ptr: NonNull<ffi::ASurfaceTexture>,
+}
+
+unsafe impl Send for SurfaceTexture {}
+
+#[derive(Debug)]
+pub struct PosixError(pub i32);
```
MarijnS95, line 15: This type seems pretty useless, should it at least implement `Error` and `Display`?

I wonder if there already exists a helper for it _somewhere_ though, nicely stringifying everything based on `errno-base.h`. I thought we already had such a thing somewhere in this crate but I might be mistaken.

MarijnS95 (changes requested):
> Is it ok to merge now?

There are some unresolved comments still open, and quite a few marked resolved which haven't actually been addressed - I've unmarked those as unresolved.

> PS: what is happening at the `format` task?

You can click on it to see what it wants. Basically make sure you run `cargo fmt` before pushing your contribution upstream.

---

After that, all that remains is checking what's up with the `Send`/`Sync`.

I presume `Send` is safe to have here _unless there's some GL context magic going on that disallows a bound texture to be used on a different thread before it is unbound_.

`Sync` requires _at least_ the above constraints too.

lattice0:
Ok, resolved all conversations except the one asking if there's a Posix struct somewhere. Couldn't find one also. What do you guys think?

lattice0:
Added api guards as suggested by @JasperDeSutter

MarijnS95:
I don't think `SurfaceTexture` is really usable without methods, so we should probably hide the entire type/module with `#![cfg(feature = "api-level-28")]`.

In addition I like @JasperDeSutter's suggestion to alias this behind a `"surface-texture"` feature flag. In the same way, should we rename `"hardware_buffer"` to `"hardware-buffer"` @JasperDeSutter?

MarijnS95:
@lattice0 Don't forget to add a changelog entry :)

You might want to wait with this before the other PRs are in and rebase + fix it at once, the changelog likes to conflict :)

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+unsafe impl Send for SurfaceTexture {}
+
+#[derive(Debug, Error)]
+pub struct PosixError(pub i32);
+
+impl std::fmt::Display for PosixError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "Posix Error: {}", self.0)
```
MarijnS95, line 21: Good enough for now, we can stringify these later.

Since more Android API (to be added in the future) shares these error codes I think we should move this type to a separate module though.

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,112 @@
...
+impl SurfaceTexture {
+    /// Assumes ownership of `ptr`
+    ///
+    /// # Safety
+    /// `ptr` must be a valid pointer to an Android [`ffi::ASurfaceTexture`].
+    pub unsafe fn from_ptr(ptr: NonNull<ffi::ASurfaceTexture>) -> Self {
+        Self { ptr }
+    }
+
+    #[cfg(feature = "api-level-28")]
```
MarijnS95, line 40: Nit: typically write attributes _below_ the doc comment (but in this case I don't think we should have these at all, and just `cfg()` once on the entire module altogether).

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    /// Returns native internal pointer
+    pub fn ptr(&self) -> NonNull<ffi::ASurfaceTexture> {
+        self.ptr
+    }
+
+    #[cfg(feature = "api-level-28")]
+    /// Returns a reference to an ANativeWindow (i.e. the Producer)
+    /// for this SurfaceTexture. This is equivalent to Java's:
+    /// Surface sur = new Surface(surfaceTexture);
```
MarijnS95, line 59: How about newlining this and adding some more intradoc-links, so that the end-result looks like (feel free to apply this suggestion):


```suggestion
    /// Returns a reference to a [`NativeWindow`] (i.e. the Producer) for this [`SurfaceTexture`].
    ///
    /// This is equivalent to Java's:
    /// ```java
    /// Surface sur = new Surface(surfaceTexture);
    /// ```
```

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub fn acquire_native_window(&mut self) -> Option<NativeWindow> {
+        let native_window = unsafe { ffi::ASurfaceTexture_acquireANativeWindow(self.ptr.as_ptr()) };
+        let n = NonNull::new(native_window)?;
+        Some(unsafe { NativeWindow::from_ptr(n) })
+    }
+
+    #[cfg(feature = "api-level-28")]
+    /// Attach the SurfaceTexture to the OpenGL ES context that is current on the calling thread.
```
MarijnS95, line 67:
```suggestion
    /// Attach the [`SurfaceTexture`] to the OpenGL ES context that is current on the calling thread.
```

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,112 @@
...
+        let r = unsafe { ffi::ASurfaceTexture_attachToGLContext(self.ptr.as_ptr(), tex_name) };
+        if r == 0 {
+            Ok(())
+        } else {
+            Err(PosixError(r))
+        }
+    }
+
+    #[cfg(feature = "api-level-28")]
+    /// Detach the SurfaceTexture from the OpenGL ES context that owns the OpenGL ES texture object.
```
MarijnS95, line 78:
```suggestion
    /// Detach the [`SurfaceTexture`] from the OpenGL ES context that owns the OpenGL ES texture object.
```

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,112 @@
...
+        let r = unsafe { ffi::ASurfaceTexture_detachFromGLContext(self.ptr.as_ptr()) };
+        if r == 0 {
+            Ok(())
+        } else {
+            Err(PosixError(r))
+        }
+    }
+
+    #[cfg(feature = "api-level-28")]
+    /// Retrieve the 4x4 texture coordinate transform matrix associated with the texture image set by the most recent call to updateTexImage.
```
MarijnS95, line 89:
```suggestion
    /// Retrieve the 4x4 texture coordinate transform matrix associated with the texture image set by the most recent call to [`update_tex_image()`][Self::update_tex_image()].
```

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,112 @@
...
+    pub fn get_transform_matrix(&self) -> [f32; 16] {
```
MarijnS95, line 90: Just like `fn timestamp()` there should be no `get_` prefix here.

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,112 @@
...
+        let mut r = [0f32; 16];
+        unsafe { ffi::ASurfaceTexture_getTransformMatrix(self.ptr.as_ptr(), r.as_mut_ptr()) };
+        r
+    }
+
+    #[cfg(feature = "api-level-28")]
+    /// Retrieve the timestamp associated with the texture image set by the most recent call to updateTexImage.
```
MarijnS95, line 97:
```suggestion
    /// Retrieve the timestamp associated with the texture image set by the most recent call to [`update_tex_image()`][Self::update_tex_image()].
```

MarijnS95 (changes requested):

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,112 @@
...
+        Some(SurfaceTexture::from_ptr(s))
+    }
+
+    /// Returns native internal pointer
+    pub fn ptr(&self) -> NonNull<ffi::ASurfaceTexture> {
+        self.ptr
+    }
+
+    #[cfg(feature = "api-level-28")]
+    /// Returns a reference to an ANativeWindow (i.e. the Producer)
```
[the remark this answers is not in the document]
MarijnS95: Resolving: see below for a more complete suggested change.

ndk/src/lib.rs
```diff
@@ -26,5 +26,6 @@ pub mod looper;
 pub mod media;
 pub mod native_activity;
 pub mod native_window;
+#[cfg(feature = "api-level-28")]
```
MarijnS95, line 29: That doesn't quite follow the standard of the other modules, I think [I requested](https://github.com/rust-windowing/android-ndk-rs/pull/267#issuecomment-1134595941) a top-level `#![]` annotation _inside_ the `surface_texture` module.

lattice0:
is it ok now?

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,98 @@
+#![cfg(feature = "api-level-28")]
+//! Bindings for [`ffi::ASurfaceTexture`]
+use super::posix::PosixError;
```
MarijnS95, line 3:
```suggestion
//! Bindings for [`ffi::ASurfaceTexture`]
#![cfg(feature = "api-level-28")]

use super::posix::PosixError;
```

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,99 @@
...
+        unsafe { ffi::ASurfaceTexture_getTransformMatrix(self.ptr.as_ptr(), r.as_mut_ptr()) };
+        r
+    }
+
+    /// Retrieve the timestamp associated with the texture image set by the most recent call to [`update_tex_image()`][Self::update_tex_image()].
+    pub fn timestamp(&self) -> i64 {
+        unsafe { ffi::ASurfaceTexture_getTimestamp(self.ptr.as_ptr()) }
+    }
+
+    /// Update the texture image to the most recent frame from the image stream.
```
MarijnS95, line 90: This is missing from the original documentation (both in Java and NDK land):

```suggestion
    /// Update the texture image to the most recent frame from the image stream.
    ///
    /// This may only be called while the OpenGL ES context that owns the texture is current on the calling thread. It will implicitly bind its texture to the `GL_TEXTURE_EXTERNAL_OES` texture target.
```

MarijnS95 (approved):
I think we're good here. We can add back `Sync` once we're sure the NDK/Java internally synchronize these threads, but the [documentation on `SurfaceTexture`](https://developer.android.com/reference/android/graphics/SurfaceTexture) mentions that certain operations are only valid when called on the thread that has an OpenGL context bound (which was previously used elsewhere), to me implying that it is safe to `Send` a `SurfaceTexture` across threads.

It seems like you've only copied the documentation summary of every function. Please make sure the entire paragraph is available or otherwise link to the appropriate function in the NDK docs: https://developer.android.com/ndk/reference/group/surface-texture

MarijnS95 (changes requested):
That shouldn't have been an approval in light of missing documentation.

lattice0:
What about now? :)

ndk/src/surface_texture.rs
```diff
@@ -32,6 +32,11 @@ impl SurfaceTexture {
     ///
     /// This function should be called with a healthy JVM pointer and with a non-null surface texture,
     /// which must be kept alive on the Java/Kotlin side.
+    /// Get a reference to the native ASurfaceTexture from the corresponding java object.
```
MarijnS95, line 35: This doesn't belong under `# Safety`, and a newline is otherwise missing that would fold this sentence into the previous pagraph.

ndk/src/surface_texture.rs
```diff
@@ -32,6 +32,11 @@ impl SurfaceTexture {
...
+    //
```
MarijnS95, line 36: Wrong.

ndk/src/surface_texture.rs
```diff
@@ -66,6 +76,12 @@ impl SurfaceTexture {
     }
 
     /// Detach the [`SurfaceTexture`] from the OpenGL ES context that owns the OpenGL ES texture object.
+    /// This call must be made with the OpenGL ES context current on the calling thread. The OpenGL
```
MarijnS95, line 79: Again, newline. Have you looked at the rendered docs before pushing this?

ndk/src/surface_texture.rs
```diff
@@ -66,6 +76,12 @@ impl SurfaceTexture {
...
+    /// ES texture object will be deleted as a result of this call. After calling this method all
+    /// calls to [`update_tex_image()`][Self::update_tex_image()] will fail until a successful call to
+    /// [`attach_to_gl_context()`][Self::attach_to_gl_context()] is made.
+    /// This can be used to access the [`SurfaceTexture`] image contents from multiple OpenGL ES contexts.
```
MarijnS95, line 83: ...

ndk/src/surface_texture.rs
```diff
@@ -82,12 +98,23 @@ impl SurfaceTexture {
         r
     }
 
-    /// Retrieve the timestamp associated with the texture image set by the most recent call to [`update_tex_image()`][Self::update_tex_image()].
+    /// Retrieve the 4x4 texture coordinate transform matrix associated with the texture image set by the most recent call to updateTexImage.
```
MarijnS95, line 101: You replaced `fn timestamp()` documentation with that for the matrix??

ndk/src/surface_texture.rs
```diff
@@ -56,6 +61,11 @@ impl SurfaceTexture {
     }
 
     /// Attach the [`SurfaceTexture`] to the OpenGL ES context that is current on the calling thread.
+    /// A new OpenGL ES texture object is created and populated with the SurfaceTexture image frame
+    /// that was current at the time of the last call to ASurfaceTexture_detachFromGLContext.
+    /// This new texture is bound to the GL_TEXTURE_EXTERNAL_OES texture target.
+    /// This can be used to access the [`SurfaceTexture`] image contents from multiple OpenGL ES contexts.
+    /// Note, however, that the image contents are only accessible from one OpenGL ES context at a time.
```
MarijnS95, line 68: Newline. Backticks/intradoc-links.

ndk/src/surface_texture.rs
```diff
@@ -82,12 +98,23 @@ impl SurfaceTexture {
...
+    /// a camera should be strictly monotonic but for a MediaPlayer may be reset when the position is set. The specific meaning and zero point of the timestamp depends
+    /// on the source providing images to the SurfaceTexture. Unless otherwise specified by the image source, timestamps cannot generally be compared across SurfaceTexture
+    /// instances, or across multiple program invocations. It is mostly useful for determining time offsets between subsequent frames
+    /// For EGL/Vulkan producers, this timestamp is the desired present time set with the EGL_ANDROID_presentation_time or VK_GOOGLE_display_timing extensions
     pub fn timestamp(&self) -> i64 {
         unsafe { ffi::ASurfaceTexture_getTimestamp(self.ptr.as_ptr()) }
     }
 
     /// Update the texture image to the most recent frame from the image stream.
+    /// This may only be called while the OpenGL ES context that owns the texture is
```
MarijnS95, line 115: Another newline.

MarijnS95 (changes requested):
> What about now? :)

No need to keep pinging/asking, I get notifications for your pushes and handle them as soon as appropriate. **Especially** when you ask, make sure that your PR is in tip-top shape first... Which it really isn't :(

MarijnS95:
@lattice0 Mind if I take over?

ndk/src/surface_texture.rs
```diff
@@ -0,0 +1,126 @@
...
+        self.ptr
+    }
+
+    /// Returns a reference to a [`NativeWindow`] (i.e. the Producer) for this [`SurfaceTexture`].
+    ///
+    /// This is equivalent to Java's:
+    /// ```java
+    /// Surface sur = new Surface(surfaceTexture);
+    /// ```
+    pub fn acquire_native_window(&mut self) -> Option<NativeWindow> {
```
MarijnS95, line 57: I don't think this needs to be `mut`, it seems I can safely crate and use multiple `NativeWindow`s (`Surface`, producers) on a single `SurfaceTexture` and the result doesn't hold a (mutable) borrow on `self` anyway.

Otherwise `update_tex_image` should probably be `mut` as well based on the name, but it should all be internally synchronize so there should be very little reason to force mutable (unique, really) access here.

MarijnS95 (approved):

MarijnS95:
> Ok, do you want to finish the PR? I can apply the changes otherwise, haven't seen your message

Not sure what was going on with your last push, but `master` was merged in and subsequently reverted, and lots of links were corrupted.

I've reverted those changes and redone the docs on this PR. Still looking for a balance to make doc links more consistent across the `ndk` crate (when to link to the wrapper type, when to link to the `ffi` type, and when to link to the Android docs) but I think I struck the right path now. Will be updated for other types separately.

PR #269, opened 2022-05-03
Release ndk-build-0.5.0, cargo-apk-0.9.0

MarijnS95:
Despite transitioning over to [xbuild] I still care about `cargo-apk` and especially releasing new features and bugfixes that already made it in before we started the process.  This release has many little additions and fixes, though some might be considered API-breaking?

[xbuild]: https://github.com/cloudpeers/xbuild/


MarijnS95:
- Default `target_sdk_version` to `30` or lower (instead of the highest supported SDK version by the detected NDK toolchain)
  for more consistent interaction with Android backwards compatibility handling and its increasingly strict usage rules:
  https://developer.android.com/distribute/best-practices/develop/target-sdk
- Remove default insertion of `MAIN` intent filter through a custom serialization function, this is better filled in by
  the default setup in `cargo-apk`. ([#241](https://github.com/rust-windowing/android-ndk-rs/pull/241))

These two changes in `ndk-build`, and the relevant API/SDK level changes in `cargo-apk` may warrant a breaking bump?

dvc94ch:
yes, will be a while before we should recommend people using xbuild by default. hope we can get xbuild there within the next 3 months.

MarijnS95:
Yeah, so better get this at least the merged features out there, and we can do more releases following that. We'll start recommending PRs to transition to `xbuild` when they're either large or nearing the "cutoff point" where `xbuild` successfully supersedes `cargo-apk`.

dvc94ch (approved):

MarijnS95:
@dvc94ch @msiglreith what are your thoughts on making this a breaking minor release?

dvc94ch:
well, either it breaks existing users or not. did we change any api's in a way that would cause build failures? but I really don't care much either way.

MarijnS95:
We changed the `target_sdk_version` which may now be different depending on what SDKs the user had installed, so it does have an impact on compiled binaries. As such `ndk-build` and `cargo-apk` should probably be breaking (and I'll add the `**Breaking**` prefix in the changelog if everyone agrees).

dvc94ch:
well, bug fixes/behavioral changes are not considered breaking according to semver rules. otherwise you could never have meaningful patch releases (whitespace changes don't count). but don't really care.

MarijnS95:
That's a fair observation. I think behavioural changes are still part of the API and should as such be guarded by the same semver principles (at least that's what I could find online about the matter) but both functions can just as much be bugfixes...

msiglreith:
Given `ndk-build` is not only used by `cargo-apk` anymore (https://crates.io/crates/ndk-build/reverse_dependencies) I would vote for a breaking change as upper levels may relay on the default intent filter for example.

MarijnS95:
@msiglreith Thanks, I've been leaning towards that as well. Breaking version for `cargo-apk` is probably less of an issue given that `cargo install` by default picks the latest, but it's more correct nevertheless (and there are probably installers / arguments to stick to a specific version).

msiglreith (approved):

PR #280, opened 2022-05-25
ndk-glue: Drop `deprecated` annotation from `native_activity()`

MarijnS95:
It has been quite some time since we introduced this, and our alternative implementation for more safely passing a native activity to the user while interoperating cleanly with `ndk_context` still hasn't found its way.

There are legitimate uses for `native_activity()` that cannot be performed through the suggested `ndk_context::android_context().vm()`, making the load of this deprecation warning heavier than the problem it tries to prevent.

### DIFF
--- a/ndk-glue/src/lib.rs
+++ b/ndk-glue/src/lib.rs
@@ -53,19 +53,30 @@ static LOOPER: Lazy<Mutex<Option<ForeignLooper>>> = Lazy::new(Default::default);
 
 static mut NATIVE_ACTIVITY: Option<NativeActivity> = None;
 
-#[deprecated = "Use `ndk_context::android_context().vm()` instead."]
+/// This function accesses a `static` variable internally and must only be used if you are sure
+/// there is exactly one version of `ndk_glue` in your dependency tree.
+///
+/// If you need access to the `JavaVM` through [`NativeActivity::vm()`] or Activity `Context`
+/// through [`NativeActivity::activity()`], please use the [`ndk_context`] crate and its
+/// [`ndk_context::android_context()`] getter to acquire the `JavaVM` and `Context` instead.
 pub fn native_activity() -> &'static NativeActivity {
     unsafe { NATIVE_ACTIVITY.as_ref().unwrap() }
 }
 
+/// This function accesses a `static` variable internally and must only be used if you are sure
+/// there is exactly one version of `ndk_glue` in your dependency tree.
 pub fn native_window() -> RwLockReadGuard<'static, Option<NativeWindow>> {
     NATIVE_WINDOW.read().unwrap()
 }
 
+/// This function accesses a `static` variable internally and must only be used if you are sure
+/// there is exactly one version of `ndk_glue` in your dependency tree.
 pub fn input_queue() -> RwLockReadGuard<'static, Option<InputQueue>> {
     INPUT_QUEUE.read().unwrap()
 }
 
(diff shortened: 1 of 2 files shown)


dvc94ch:
Can you provide an example? Unless you're building a windowing library like winit, your code will be non portable. Still don't quite understand the use case. There is nothing that isn't specific to the NativeActivity that can't be done via the jni.

MarijnS95:
I want to access the path variables that Android has already stored for me in `NativeActivity`, without going through the hassle of JNI or using `app_dirs2` (which I finally fixed the broken `SharedData`/`SharedConfig` retrieval for: https://github.com/app-dirs-rs/app_dirs2/pull/20).

At the same time I want to access new features like `set_window_flags()` (#278), which I have yet to come up with a `winit` binding for nor know if I can fully map all the flags there.

Finally it's just inconvenient for windowing library writers like `winit` to have to allow deprecated-but-not-really function in their codebase, especially when starting to use use newer window/activity functions on that side.

Since my approach suggested in #211 has effectively been shot down as being too complicated (pass `NativeActivity` into the function wrapped/called by `ndk_glue::main`, then force the user to pass that into `winit` somehow, and remove `ndk_glue::native_activity()` and `native_window()` altogether) the only alternative is storing pointers to those objects in `ndk-context` too. Was there a particular reason why we didn't/don't/can't do that? I don't remember :sweat_smile:

MarijnS95:
@dvc94ch Yet another idea - albeit a stupid one because it affects `ndk` instead of `ndk-glue` - move the deprecation markers to `NativeActivity::vm()` and `::activity()`.

dvc94ch:
> I want to access the path variables that Android has already stored for me in NativeActivity, without going through the hassle of JNI or using app_dirs2

using something like app_dirs2 is really what we should be incouraging. it seems like a weird place to micro optimize. we can remove the deprecation warning if you feel strongly about it. I believe I added it because you suggested it?

MarijnS95:
I think [you suggested the deprecation attribute](https://github.com/rust-windowing/android-ndk-rs/pull/223#issuecomment-1038985106), I [originally disapproved of it](https://github.com/rust-windowing/android-ndk-rs/pull/223#discussion_r805871322), you removed it, then [I requested it again](https://github.com/rust-windowing/android-ndk-rs/pull/223#issuecomment-1039283197) in light of https://github.com/rust-windowing/android-ndk-rs/compare/master...MarijnS95:anativeactivity-instance removing static globals...

As per the PR description that branch isn't anywhere near landing, so I retract my recommendation/request for deprecating it :grimacing: - I think there are valid use-cases for this function, users should just be careful using it.

We have this problem anyway. In a recent test project I didn't call `native_activity()` myself, but accidentally paired `ndk-glue 0.6` with `winit 0.26` and wondered why no events were arriving. That can only be solved with the suggested branch above, it is not something this `#[deprecated]` attribute guards against.

dvc94ch (approved):

MarijnS95:
And of course I should at least maintain the original `deprecation` message through a doc-comment. English isn't really with me today, let me know if it reads today :sweat_smile:

Can we do anything else to mitigate this issue, or make the user aware? How about defining a `#[no_mangle]` symbol (does it need to be `pub`?) that carries this message in its name (elaborated in a doc-comment), that could fail linking when two `ndk_glue`s appear in the dependency graph? Only one way to find out, I guess :)

dvc94ch (approved):
thanks

PR #282, opened 2022-05-28
ndk-glue: Switch to `parking_lot` for `map`pable and `Send`able lock guards

MarijnS95:
Depends on #282

It has come up previously that the `native_window()` function is easier to use if we could transpose a `LockGuard<Option<>>` into an `Option<LockGuard<>>`.  After all, as soon as you have the lock you only need to check the `Option<>` value once and are thereafter guaranteed that its value won't change, until the lock is given up.

At the same time `parking_lot` has a `send_guard` feature which allows moving a lock guard to another thread (as long as `deadlock_detection` is disabled); a use case for this recently came up [in glutin] where the `NativeWindow` handle lock should be stored with a GL context so that our `fn on_window_destroyed()` callback doesn't return until after the `Surface` created on it is removed from the context and destroyed.  Glutin forces this context to be `Send`, and a lock guard already allows `Sync` if `NativeWindow` is `Sync` (which it is).

[in glutin]: https://github.com/rust-windowing/glutin/pull/1411#discussion_r883677680

### DIFF
--- a/ndk-glue/src/lib.rs
+++ b/ndk-glue/src/lib.rs
@@ -7,6 +7,7 @@ use ndk::native_activity::NativeActivity;
 use ndk::native_window::NativeWindow;
 use ndk_sys::{AInputQueue, ANativeActivity, ANativeWindow, ARect};
 use once_cell::sync::Lazy;
+use parking_lot::{MappedRwLockReadGuard, RwLock, RwLockReadGuard};
 use std::ffi::{CStr, CString};
 use std::fmt;
 use std::fs::File;
@@ -15,7 +16,7 @@ use std::ops::Deref;
 use std::os::raw;
 use std::os::unix::prelude::*;
 use std::ptr::NonNull;
-use std::sync::{Arc, Condvar, Mutex, RwLock, RwLockReadGuard};
+use std::sync::{Arc, Condvar, Mutex};
 use std::thread;
 
 #[cfg(feature = "logger")]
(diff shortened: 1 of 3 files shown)


MarijnS95:
I mentioned in #285 that we can `const`ly construct the `RwLock`s from `parking_lot` _without_ `Lazy` wrapper, but that would bump our MSRV to Rust 1.61 (current stable as of writing): https://github.com/Amanieu/parking_lot/pull/325.

Is that something we want to do or have hard feelings against? We never cared much about MSRV and I wouldn't even know where this crate (the chain of these crates) are at right now.

If we do, we should _at least_ test the MSRV that we set in the CI and specify it in `Cargo.toml` so that users get a sensible error, rather than a mid-crate compiler error stating that these functions are not `const`.

ndk-glue/src/lib.rs
```diff
@@ -67,7 +68,20 @@ pub fn native_activity() -> &'static NativeActivity {
...
 }
 
-pub struct LockReadGuard<T: ?Sized + 'static>(RwLockReadGuard<'static, T>);
+pub struct LockReadGuard<T: ?Sized + 'static>(MappedRwLockReadGuard<'static, T>);
+
+impl<T> LockReadGuard<T> {
+    /// Transpose an [`Option`] wrapped inside a [`LockReadGuard`]
+    ///
+    /// This is a _read_ lock for which the contents can't change; hence allowing the user to only
+    /// check for [`None`] once and hold a a lock containing `T` directly thereafter, without
```
msiglreith, line 77:
```suggestion
    /// check for [`None`] once and hold a lock containing `T` directly thereafter, without
```

msiglreith (approved):

PR #286, opened 2022-06-10
cargo-apk,ndk-build: Move NDK r23 `-lgcc` workaround to `cargo_ndk` to reuse in `--` subcommand

MarijnS95:
Winit is currently being hit by `-lgcc not found` linker errors because GH Actions' [virtual-environments recently migrated to Android NDK r23], and it turns out it's using the niche `--` subcommand to invoke regular `cargo` commands under an NDK environment.

The workaround using `RUSTFLAGS` is only patched into the `build()` command; by moving it into our `cargo_ndk()` environment preparation function this workaround is universally available and usable.

Note that this is a breaking change, we're changing the API signature of `ndk-build`.

[virtual-environments recently migrated to Android NDK r23]: https://github.com/actions/virtual-environments/issues/5595


msiglreith (approved):

msiglreith:
seems to work locally with ndk-r24 👍

PR #287, opened 2022-06-10
cargo-apk: Append `--target` to blanket `cargo apk --` calls when not provided

MarijnS95:
Depends on #286, #283

When the user doesn't provide a `--target` we default the triple to the currently connected device over `adb`, or otherwise fall back to `aarch64`.

While this triple is always used to determine what NDK environment to provide, it's never added to the arguments passed after `--`; for example a `cargo apk -- test --no-run` will try to build tests for the host unless explicitly called with `--target`.

---

I can't currently think of any command that one might want to run under `cargo apk --` with an NDK environment, but without support for `--target`. If there is, that's a valid argument to not pass this flag (but perhaps clearly document it, maybe even warn about it if we don't find `--target` in the call?).


MarijnS95:
@msiglreith Thoughts? If we want this, we should probably get it in right in the next patch release (that I'd like to do soon™ to fix the `winit` CI with #286), and see how it fares.

Or is there a command you think may not support `--target` but may be relevant to run here?

msiglreith (approved):
fine for me
`check`, `build`, `test`, `clippy`, `doc` and `bench` seem all to support it which are probably the most used CI configurations and which might user want through `cargo-apk`

MarijnS95:
Let's go for it then, this is probably a net-improvement and otherwise we'll come up with an alternative :+1:

PR #290, opened 2022-06-11
ndk: Move library documentation to modules

MarijnS95:
Instead of having an outdated overview of modules and their contents in the library documentation, move this to the module documentation header.  Rustdoc already provides a clickable link for every module (these manual list didn't include _any_ links), and also includes the first sentence/paragraph from this module documentation.

This also inserts some missing `struct` documentation and cleans up existing references to link to the NDK docs more often and specify exactly what type it is wrapping, instead of linking to the "rather useless" `ffi` representation for that NDK type.


MarijnS95:
With this PR we go from:

image

To:

image

msiglreith (approved):
nice!

PR #291, opened 2022-06-11
ndk: Replace `posix` module with `std::io::Error::from_raw_os_error`

MarijnS95:
Platform specific - i.e. POSIX - error codes can be wrapped in humand readable form by Rust's `std` through [`Error::from_raw_os_error()`].  Certain NDK interfaces use these codes directly from `errno.h` through [`Errors.h`].

[`Error::from_raw_os_error()`]: https://doc.rust-lang.org/std/io/struct.Error.html#method.from_raw_os_error
[`Errors.h`]: https://cs.android.com/android/platform/superproject/+/master:system/core/libutils/include/utils/Errors.h


MarijnS95:
@lattice0 this is what I was thinking about in your PR (#262), but couldn't quite find it back then :sweat_smile:

lattice0:
Awesome! Much cleaner now 😍

msiglreith (approved):

PR #292, opened 2022-06-11
ndk/input_queue: Replace `InputQueueError` with `std::io::Error` and add missing docs

MarijnS95:
These `AInputQueue` functions all return [standard `errno.h` error codes] which can and should be communicated back to the user instead of appearing as some opaque magic struct.

In addition import the NDK documentation for all these functions and make the scope of `unsafe` blocks smaller.

[standard `errno.h` error codes]: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/jni/android_view_InputQueue.cpp;l=112;drc=7b3193e1e86ab5c0fcb784cb8616864045dd2515

### DIFF
--- a/ndk/src/input_queue.rs
+++ b/ndk/src/input_queue.rs
@@ -2,11 +2,13 @@
 //!
 //! [`AInputQueue`]: https://developer.android.com/ndk/reference/group/input#ainputqueue
 
+use std::io::{Error, Result};
 use std::os::raw::c_int;
-use std::ptr;
-use std::ptr::NonNull;
+use std::ptr::{self, NonNull};
 
 use crate::event::InputEvent;
+#[cfg(doc)]
+use crate::event::KeyEvent;
 use crate::looper::ForeignLooper;
 
 /// A native [`AInputQueue *`]
@@ -23,9 +25,6 @@ pub struct InputQueue {
 unsafe impl Send for InputQueue {}
 unsafe impl Sync for InputQueue {}
 
-#[derive(Debug)]
-pub struct InputQueueError;
-
 impl InputQueue {
     /// Construct an [`InputQueue`] from the native pointer.
     ///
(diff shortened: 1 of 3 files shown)


ndk/src/input_queue.rs
```diff
@@ -33,45 +38,61 @@ impl InputQueue {
...
 
-    pub fn get_event(&self) -> Option<InputEvent> {
-        unsafe {
-            let mut out_event = ptr::null_mut();
-            if ffi::AInputQueue_getEvent(self.ptr.as_ptr(), &mut out_event) < 0 {
-                None
-            } else {
+    /// Returns the next available [`InputEvent`] from the queue.
+    ///
+    /// Returns [`std::io::ErrorKind::WouldBlock`] if no event is available.
```
MarijnS95, line 43: Should perhaps catch this case and return `Result<Option<>>` instead? That way they can easily write `while let Some(event) = get_event()` followed by either `?` or `.unwrap()`.

ndk/src/input_queue.rs
```diff
@@ -33,45 +38,61 @@ impl InputQueue {
...
     }
 
+    /// Sends the key for standard pre-dispatching that is, possibly deliver it to the current IME
+    /// to be consumed before the app.
+    ///
+    /// Returns [`Some`] if it was not pre-dispatched, meaning you can process it right now. If
+    /// [`None`] is returned, you must abandon the current event processing and allow the event to
+    /// appear again in the event queue (if it does not get consumed during pre-dispatching).
+    ///
+    /// Also returns [`None`] if `event` is not a [`KeyEvent`].
```
MarijnS95, line 73: Not sure if this is always true, but we can change the API signature to take a `KeyEvent` directly?

msiglreith: seems fine implementation-wise: https://android.googlesource.com/platform/frameworks/base.git/+/refs/heads/master/core/jni/android_view_InputQueue.cpp#115

ndk/src/input_queue.rs
msiglreith: it doesn't seem to be possible to differentiate between 'would block' and other errors, or is it? All error cases seem to map to `-1` /:

ndk/src/input_queue.rs
MarijnS95: Yeah exactly, but that's going off of the current implementation and I don't know if this is modified by vendors. Also not sure how pointer gestures like swipes are handled by an IME... touches on a screen are pointer/motions as well. Perhaps I should read up on what this does :sweat_smile:

ndk/src/input_queue.rs
MarijnS95: Not sure I understand, the only error [this codepath] currently returns is `WOULD_BLOCK`, but there's no guarantee that it won't return other codes in the future. So for now it's either:

1. Yes there is an event;
2. No there is no event (`WouldBlock`);
3. Some other error that the user probably doesn't know/care about, abort.

[this codepath]: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/jni/android_view_InputQueue.cpp;l=112;drc=7b3193e1e86ab5c0fcb784cb8616864045dd2515

MarijnS95: > All error cases seem to map to `-1` /:

In other words, @msiglreith what "map" are you referring to exactly?

ndk/src/input_queue.rs
MarijnS95: Let's leave this for now, it's documented and users can figure out themselves what to do.

ndk/src/input_queue.rs
msiglreith: ah sorry, I don't remember where I looked but apparently at the wrong place 😅  - makes more sense now with the WOULD_BLOCK path

MarijnS95: @msiglreith :+1: - it's now changed to return `None` instead of putting the burden on the user to separate `e.kind() == WouldBlock` - which is a case you _must_ handle for this API to work/be useful (or call `has_events()` all the time?) - also implemented on `from_raw_os_error()` + `e.kind()` because there's no clean constant `-11` for `EWOULDBLOCK` in Rust.

msiglreith (approved):

PR #293, opened 2022-06-11
Enable `doc_cfg` and `doc_auto_cfg` for all `[features]` crates on https://docs.rs

MarijnS95:
The documentation currently doesn't expose what features a user needs to enable before certain functionality becomes available, requiring them to look at the sourcecode as we don't really provide much documentation or guideance towards this.

That's totally fine because it is a job for `rustdoc`, one that it performs very well when the experimental `doc_cfg` feature is enabled, and specifically `doc_auto_cfg` to not have to manually re-specify every `#[cfg(<guard>)]` within a `#[cfg_attr(docs_rs, doc(cfg(<guard>)))]`.

docs.rs builds on `nightly` which makes this possible, and docs can be built with the `docsrs` cfg enabled locally to see the result, e.g.:

    RUSTDOCFLAGS="-Dwarnings --cfg docsrs" cargo +nightly doc --no-deps --workspace --target aarch64-linux-android --all-features --document-private-items --open

### DIFF
--- a/ndk-glue/Cargo.toml
+++ b/ndk-glue/Cargo.toml
@@ -27,6 +27,7 @@ test = ["ndk/test", "ndk-sys/test"]
 logger = ["android_logger", "ndk-macro/logger"]
 
 [package.metadata.docs.rs]
+rustdoc-args = ["--cfg", "docsrs"]
 targets = [
     "aarch64-linux-android",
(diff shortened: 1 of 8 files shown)


MarijnS95:
The module, for example:

image

msiglreith (approved):

PR #295, opened 2022-06-12
ndk: Replace `HardwareBufferError` with `std::io::Error` through new helper conversion function

MarijnS95:
Depends on #294

These `AHardwareBuffer` functions all return [standard `errno.h` error codes] which can be communicated back to the user in ergonomic form instead of appearing as a newtyped struct holding a "meaningless" `i32`.

Perform the same conversion to just `NativeWindow::set_buffers_geometry` which also returns `errno.h` codes: this change is small enough to not warrant _yet another_ PR, as we've already changed the exact same for `InputQueueError` in [#292].  (Note that this is intentionally omitted from the changelog since `set_buffers_geometry()` is only making its debut in the upcoming release, this change isn't breaking).

This pattern is now prevalent across the codebase (`hardware_buffer`, `surface_texture`, and `native_window` all use it), warranting the new `status_to_io_result()` helper function to be introduced in an also-new `utils.rs` module.

[standard `errno.h` error codes]: https://cs.android.com/android/platform/superproject/+/master:frameworks/native/libs/nativewindow/AHardwareBuffer.cpp;drc=4120991d92a0265ce480672f704612556328057e
[#292]: https://github.com/rust-windowing/android-ndk-rs/pull/292

### DIFF
--- a/ndk/CHANGELOG.md
+++ b/ndk/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Improve library and structure documentation, linking back to the NDK docs more rigorously. (#290)
 - **Breaking:** input_queue: `get_event()` now returns a `Result` with `std::io::Error`; `InputQueueError` has been removed. (#292)
 - **Breaking:** input_queue: `has_events()` now returns a `bool` directly without being wrapped in `Result`. (#294)
+- **Breaking:** hardware_buffer: `HardwareBufferError` has been removed and replaced with `std::io::Error` in return types. (#295)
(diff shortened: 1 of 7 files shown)


MarijnS95:
@dvc94ch Assuming you have more rights to this repo, can you turn on `Always suggest updating pull request branches`?

image

Thanks!

dvc94ch:
Nope, don't have access

MarijnS95:
@francesca64 maybe?

PR #296, opened 2022-06-12
ndk: Fix `HardwareBuffer` leak for `AndroidBitmap` and clarify handling in docs

MarijnS95:
Depends on #294, #295

According to [`AndroidBitmap_getHardwareBuffer`]:

    outBuffer   On success, is set to a pointer to the AHardwareBuffer
                associated with bitmap. This acquires a reference on the
                buffer, and the client must call AHardwareBuffer_release
                when finished with it.

By returning the `outBuffer` wrapped in `HardwareBuffer` we represent a weak instead of strong reference, and don't `_release` it on `Drop`.  If we instead wrap it in a `HardwareBufferRef` we get the desired `release` semantics on `Drop` and prevent resource leaks for the user.

Note that a similar API on [`AImage_getHardwareBuffer`] does **not** aquire a strong reference, and should remain returning a weak `HardwareBuffer`.

This feat isn't very well described in the docs, so we now echo what was only stated on `HardwareBufferRef` across the `from_ptr()`/`from_jni()` implementations.

Finally, just like `ForeignLooper` and `NativeWindow` this `acquire`+`release` kind of type can implement `Clone` to allow the user to take advantage of internal refcounting instead of having to re-wrap it in `Rc` or `Arc`.

[`AndroidBitmap_getHardwareBuffer`]: https://developer.android.com/ndk/reference/group/bitmap#androidbitmap_gethardwarebuffer
[`AImage_getHardwareBuffer`]: https://developer.android.com/ndk/reference/group/media#aimage_gethardwarebuffer

### DIFF
--- a/ndk/src/media/image_reader.rs
+++ b/ndk/src/media/image_reader.rs
@@ -322,6 +322,20 @@ impl Image {
         construct(|res| unsafe { ffi::AImage_getNumberOfPlanes(self.as_ptr(), res) })
     }
 
+    /// Get the hardware buffer handle of the input image intended for GPU and/or hardware access.
+    ///
+    /// Note that no reference on the returned [`HardwareBuffer`] handle is acquired automatically.
+    /// Once the [`Image`] or the parent [`ImageReader`] is deleted, the [`HardwareBuffer`] handle
+    /// from previous [`Image::get_hardware_buffer()`] becomes invalid.
+    ///
+    /// If the caller ever needs to hold on a reference to the [`HardwareBuffer`] handle after the
+    /// [`Image`] or the parent [`ImageReader`] is deleted, it must call
+    /// [`HardwareBuffer::acquire()`] to acquire an extra reference, and [`drop()`] it when
+    /// finished using it in order to properly deallocate the underlying memory managed by
+    /// [`HardwareBuffer`]. If the caller has acquired an extra reference on a [`HardwareBuffer`]
+    /// returned from this function, it must also register a listener using
+    /// [`ImageReader::set_buffer_removed_listener()`] to be notified when the buffer is no longer
+    /// used by [`ImageReader`].
     #[cfg(feature = "hardware_buffer")]
     pub fn get_hardware_buffer(&self) -> Result<HardwareBuffer> {
         unsafe {
--- a/ndk/CHANGELOG.md
+++ b/ndk/CHANGELOG.md
@@ -16,6 +16,7 @@
 - **Breaking:** input_queue: `get_event()` now returns a `Result` with `std::io::Error`; `InputQueueError` has been removed. (#292)
 - **Breaking:** input_queue: `has_events()` now returns a `bool` directly without being wrapped in `Result`. (#294)
 - **Breaking:** hardware_buffer: `HardwareBufferError` has been removed and replaced with `std::io::Error` in return types. (#295)
+- Fixed `HardwareBuffer` leak on buffers returned from `AndroidBitmap::get_hardware_buffer()`. (#296)
 - **Breaking:** Update `jni` crate (used in public API) from `0.18` to `0.19`. (#300)
 
 # 0.6.0 (2022-01-05)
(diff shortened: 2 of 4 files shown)


MarijnS95:
@JasperDeSutter I've recently implemented this differently for `NativeWindow` in #207 by always making sure the `NativeWindow` is acquired, hence _always_ release it on `Drop`. That way it can't accidentally become invalid either, since there exist no checks here to see whether the object is still alive.

What do you think, is it advantageous to have an ever-so-slightly-faster API that doesn't call `acquire`, or should we remove `HardwareBufferRef` entirely? IMO seems advantageous to have a somewhat consistent API around this.

For the cases where we have to call acquire (`clone_from_ptr()` in #207) to construct a valid `NativeWindow` I wish I could have returned (through transmuting) a `&NativeWindow` instead. That immediately makes this clear through the type system, puts an implicit borrow on the source type, and obviates the `acquire` operation through a `.clone()` when the user wishes to store an owned version of it.

But I didn't manage to write such a transmute since `NativeWindow` is already a pointer reference inside, `&NativeWindow` would be a double indirection and not type compatible in any way.

What do you think?

JasperDeSutter:
Nice catch about the bitmap api, some of these gotchas are easy to miss in the documentation.

The current design which is based on camera/video usage where frames are transmitted frequently and usually has a need to be well-optimised. You wouldn't normally lock the buffer if you can synchronously process or memcpy it before the system callback returns. However I didn't do any profiling to see if this actually matters at all so it is premature optimization.

I agree with just removing the *Ref system if it leads to a cleaner API, let's not delve into more unsafe than is needed with the transmutes. Having the reference (`&`) explicitly in the type system is very clear for users, but the extra work (or risk?) doesn't seem worth it. Note that std library types like Path and PathBuf also don't always do this.

MarijnS95:
> Nice catch about the bitmap api, some of these gotchas are easy to miss in the documentation.

I wouldn't be surprised if they added that doc bit later, certain functions (`onWindowCreated` anyone?) don't document this at all (last I looked at them).

Lets get this PR in with all its documentation improvements and a leak-fix for `Bitmap`, and we can make a note to revisit this API later. In hindsight I don't think we need lots of unsafe transmuting if we have a `HardwareBuffer(ffi::AHardwareBuffer)` (that we make unsized to prevent the user owning it), hand out `&` to it and perhaps a custom `.clone()` function that hands out a `HardwareBufferRef` which is basically a `NonNull<HardwareBuffer>` then, performing the `_acquire` and `_release`. That's easy to replicate for `NativeWindow` too, execpt that a `'_` lifetime on `native_window()` isn't very descriptive :grimacing:

ndk/src/media/image_reader.rs
```diff
@@ -322,6 +322,20 @@ impl Image {
...
+    ///
+    /// Note that no reference on the returned [`HardwareBuffer`] handle is acquired automatically.
+    /// Once the [`Image`] or the parent [`ImageReader`] is deleted, the [`HardwareBuffer`] handle
+    /// from previous [`Image::get_hardware_buffer()`] becomes invalid.
+    ///
+    /// If the caller ever needs to hold on a reference to the [`HardwareBuffer`] handle after the
+    /// [`Image`] or the parent [`ImageReader`] is deleted, it must call
+    /// [`HardwareBuffer::acquire()`] to acquire an extra reference, and [`drop()`] it when
+    /// finished using it in order to properly deallocate the underlying memory managed by
+    /// [`HardwareBuffer`]. If the caller has acquired extra reference on an [`HardwareBuffer`]
```
msiglreith, line 335:
```suggestion
    /// [`HardwareBuffer`]. If the caller has acquired an extra reference on a [`HardwareBuffer`]
```

msiglreith (approved):

ndk/src/media/image_reader.rs
MarijnS95: @msiglreith I've copied the documentation verbatim from Android docs for the most part and only replaced intradoclinks + markdown (and rework sections about lifetime when we implicitly manage it through `Drop`), but sure we can patch up their broken grammar :)

Would you mind reviewing dependent PRs too? Then we can rebase and merge this whole train :)

PR #298, opened 2022-06-16
ndk-build: Switch NDK r23 -lgcc workaround to `CARGO_ENCODED_RUSTFLAGS` env var for paths with spaces

EmilioLaiso:
On Windows user paths usually include people's full name, with a space. If they have their Rust project somewhere in their user profile `cargo` fails to parse the workaround path:

    process didn't exit successfully: `rustc - --crate-name ___ --print=file-names -L '"C:\Users\Bla' 'Haj\Desktop\Breda\target\cargo-apk-temp-extra-link-libraries"' ... (exit code: 1)

As per the [documentation for `RUSTFLAGS`] its contents are split by a very simple `.split(" ")` algorithm without caring for quotation delimiters or escape characters. To retain backwards compatibility this has been implemented in a new environment variable named `CARGO_ENCODED_RUSTFLAGS`, which separates by ASCII Unit Separators (`0x1f`) instead.

[documentation for `RUSTFLAGS`]: https://doc.rust-lang.org/cargo/reference/environment-variables.html

### DIFF
--- a/ndk-build/CHANGELOG.md
+++ b/ndk-build/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Allow NDK r23 `-lgcc` workaround to work for target directories containing spaces. ([#298](https://github.com/rust-windowing/android-ndk-rs/pull/298))
+
 # 0.6.0 (2022-06-11)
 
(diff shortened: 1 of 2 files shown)


MarijnS95:
I'll merge this after adding some space-testing in the CI and seeing it fail first :)

Issue #140, opened 2021-04-15
support using a user supplied signing key


dvc94ch:
at work we have some custom android builds using gradle with a java wrapper. there is no real business case to change what already works, but using a user supplied signing key would be a requirement.

MarijnS95:
> there is no real business case to change what already works

Not sure whether this applies to your app or `cargo-apk`... But hardcoding a debug key like we do right now means `cargo-apk` is not production ready yet, that's only for hobbyists and developers playing around.

dvc94ch:
Well, I might one day port our setup if I have a week of vaccation and no other projects on my list. Not sure, haven't put much thought into it yet. Since we release via ci, a cli option would be fine.

MarijnS95:
Implemented in #299!

PR #299, opened 2022-06-17
cargo-apk: Configure APK signing keystore location through manifest

MarijnS95:
All builds currently use a fixed debug keystore.  This is cumbersome on machines where `keytool` isn't installed, and odd for `release` builds.

To make sharing debug APKs more consistent across developers this allows them to check in their `debug.keystore` in a repository and reuse it for everyone.

At the same time there's no sensible default-debug keystore for release builds; these require explicit configuration through the manifest (yet nothing withholds the user from passing their debug keystore here, if they so desire - but at least it's explicit).

### DIFF
--- a/cargo-apk/src/apk.rs
+++ b/cargo-apk/src/apk.rs
@@ -6,7 +6,7 @@ use ndk_build::cargo::{cargo_ndk, VersionCode};
 use ndk_build::dylibs::get_libs_search_paths;
 use ndk_build::error::NdkError;
 use ndk_build::manifest::{IntentFilter, MetaData};
-use ndk_build::ndk::Ndk;
+use ndk_build::ndk::{Key, Ndk};
 use ndk_build::target::Target;
 use std::path::PathBuf;
 use std::process::Command;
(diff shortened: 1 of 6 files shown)


msiglreith (approved):

cargo-apk/src/apk.rs
```diff
@@ -139,6 +139,12 @@ impl<'a> ApkBuilder<'a> {
 
         let crate_path = self.cmd.manifest().parent().expect("invalid manifest path");
 
+        let is_debug_profile = match self.cmd.profile() {
+            Profile::Dev => true,
+            Profile::Release => false,
+            Profile::Custom(c) => todo!("Read `debug` field from `[profile.{}]`?", c),
```
msiglreith, line 145: I'm a bit about this one - seems to effectively disallow custom profiles for now.

MarijnS95: @msiglreith Fair enough, perhaps I should just implement it in this PR or choose a "sensible default" - which would be?

msiglreith: restricting it to `Profile::Dev` should be fine imo

MarijnS95: We're doing it the same as a default value for `manifest.android_manifest.application.debuggable`, fwiw.

kanerogers:
Nice work! Does this also close #140 ?

MarijnS95:
@kanerogers Correct, thanks for mentioning!

kanerogers:
No worries! Was just googling around for some documentation to help a user with keysigning - we're huge fans of `cargo apk` over at Hotham! https://github.com/leetvr/hotham

PR #305, opened 2022-06-21
ndk-build,cargo-apk: Rename `intent_filters` back to `intent_filter`

### LINKED ISSUE #304: Custom intent broken since cargo-apk 0.9.0
Hello!

As the title suggests the custom intent filtration appears to been broken since release 0.9.0. In cargo-apk version 0.8.2 things worked as expected. And looking into what code changed between 0.8.2 and 0.9.0 I found this https://github.com/rust-windowing/android-ndk-rs/pull/241 though I'm not sure if this the culprit.

This issue can be reproduced by building using cargo-apk 0.8.2 and view what intents the app have registered to and building the same app using a newer of cargo-apk.

MarijnS95:
Fixes #304, CC @hoodad

We commonly name vectors of items with their singular form so that they read nicer in TOML and XML (i.e. `[[intent_filter]]` adds a single entry to that list, and gets serialized as a single `<intent-filter>` XML element) and this matches Android's `intent-filter` naming too.

This change in a6f3e13 ("ndk-build: Move default serialization of `MAIN` intent filter to `cargo-apk` (#241)") was made purely with `Vec` in mind, but breaks existing manifest parsing nor was anticipated to be a breaking change in this area as the README still carries `intent_filter` instead of `intent_filters` in its `Cargo.toml` metadata reference.

---

I want to try landing this without yet-another breaking release for `ndk-build`: can we do that by yanking the previous releases? We don't need a breaking release for `cargo-apk` either way, but it may be nice to yank their older releases too.

### DIFF
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Unreleased
 
+# 0.9.3 (2022-07-05)
+
 - Allow configuration of alternate debug keystore location; require keystore location for release builds. ([#299](https://github.com/rust-windowing/android-ndk-rs/pull/299))
+- **Breaking:** Rename `Activity::intent_filters` back to `Activity::intent_filter`. ([#305](https://github.com/rust-windowing/android-ndk-rs/pull/305))
 
 # 0.9.2 (2022-06-11)
 
--- a/cargo-apk/Cargo.toml
+++ b/cargo-apk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-apk"
-version = "0.9.2"
+version = "0.9.3"
 authors = ["The Rust Windowing contributors"]
 edition = "2018"
 description = "Helps cargo build APKs"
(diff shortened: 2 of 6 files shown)


MarijnS95:
Behind keeping the name as it was, to be very precise :grimacing: - it's more in-line with all the other properties too.

Now let's see why NDK r23 is hitting us _again_ - we already had to work around the `libgcc` bug and now it's breaking our CI :/

MarijnS95:
A relaunch seems to have fixed it, but we should perhaps bump the cache (newer `api-level`) at some point.

This is also the time ping @rust-windowing/publishers for suggestions/approval on breaking changes in a patch release if all the other versions are yanked; and I also need you to do the yanking if this is okay :)

msiglreith (approved):

MarijnS95:
After some thought I still don't think it's a good idea to publish a breaking change in a patch release and yank the older versions, so we'll release it as breaking change instead.

@rust-windowing/publishers it'd still be nice to yank `cargo-apk 0.9.0-0.9.2` and `ndk-build 0.6.0` though, thanks!

PR #312, opened 2022-07-05
media_codec: `NativeWindow` can be optional in `fn configure()`

lattice0:
For rendering outside a surface, you can pass a null in the C++ bindings, so translate this to a `None` in Rust side.

https://developer.android.com/reference/android/media/MediaCodec#configure(android.media.MediaFormat,%20android.view.Surface,%20android.media.MediaCrypto,%20int)

### DIFF
--- a/ndk/src/media/media_codec.rs
+++ b/ndk/src/media/media_codec.rs
@@ -229,14 +229,18 @@ impl MediaCodec {
     pub fn configure(
         &self,
         format: &MediaFormat,
-        surface: &NativeWindow,
+        surface: Option<&NativeWindow>,
         direction: MediaCodecDirection,
     ) -> Result<()> {
         let status = unsafe {
             ffi::AMediaCodec_configure(
                 self.as_ptr(),
                 format.as_ptr(),
-                surface.ptr().as_ptr(),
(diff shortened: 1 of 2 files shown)


MarijnS95 (changes requested):
1. Missing a changelog entry;
2. Where is this documented? You should link to it;
3. This is more concise to write as a `.map_or()`.

ndk/src/media/media_codec.rs
```diff
@@ -229,14 +229,14 @@ impl MediaCodec {
...
     ) -> Result<()> {
         let status = unsafe {
             ffi::AMediaCodec_configure(
                 self.as_ptr(),
                 format.as_ptr(),
-                surface.ptr().as_ptr(),
+                surface.map_or(ptr::null_mut(), |s| s.ptr().as_ptr()),
                 ptr::null_mut(),
                 if direction == MediaCodecDirection::Encoder {
                     1
```
MarijnS95, line 242: While at it, can you replace this with:
```suggestion
                    ffi::AMEDIACODEC_CONFIGURE_FLAG_ENCODE
```
No extra PR needed for that.

MarijnS95 (approved):

PR #315, opened 2022-07-08
ndk-sys: Switch remaining enums to newtypes with associated constants

MarijnS95:
Searching for `pub type` reveals a lot of typedefs for enums that could have been a newtype with associated constants.  Unfortunately there are still a lot of private enums in the source, some with a `typedef` following: these cannot be converted yet, otherwise we might have gotten away with globally enabling `--default-enum-style` to apply this style to _every_ enum.

At the same time certain enums could use `--bitfield-enum` yet those eligible also suffer from the same lack of a type, requiring the user to work with `_bindgen_ty_XX` which is far from ergonomic.

### DIFF
--- a/ndk/src/hardware_buffer.rs
+++ b/ndk/src/hardware_buffer.rs
@@ -18,72 +18,82 @@ pub struct HardwareBufferUsage(pub ffi::AHardwareBuffer_UsageFlags);
 
 impl HardwareBufferUsage {
     pub const CPU_READ_NEVER: Self =
-        Self(ffi::AHardwareBuffer_UsageFlags_AHARDWAREBUFFER_USAGE_CPU_READ_NEVER);
+        Self(ffi::AHardwareBuffer_UsageFlags::AHARDWAREBUFFER_USAGE_CPU_READ_NEVER);
     pub const CPU_READ_RARELY: Self =
-        Self(ffi::AHardwareBuffer_UsageFlags_AHARDWAREBUFFER_USAGE_CPU_READ_RARELY);
+        Self(ffi::AHardwareBuffer_UsageFlags::AHARDWAREBUFFER_USAGE_CPU_READ_RARELY);
     pub const CPU_READ_OFTEN: Self =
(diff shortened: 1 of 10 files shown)


ndk/src/hardware_buffer.rs
```diff
@@ -18,72 +18,82 @@ pub struct HardwareBufferUsage(pub ffi::AHardwareBuffer_UsageFlags);
 
 impl HardwareBufferUsage {
     pub const CPU_READ_NEVER: Self =
-        Self(ffi::AHardwareBuffer_UsageFlags_AHARDWAREBUFFER_USAGE_CPU_READ_NEVER);
+        Self(ffi::AHardwareBuffer_UsageFlags::AHARDWAREBUFFER_USAGE_CPU_READ_NEVER);
```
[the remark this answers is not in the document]
MarijnS95: https://github.com/rust-lang/rust-bindgen/issues/777 Unfortunately this has gotten stale on the `bindgen` side :/

ndk/src/hardware_buffer.rs
```diff
@@ -263,7 +273,7 @@ impl HardwareBuffer {
             None => std::ptr::null(),
         };
         construct(|res| unsafe {
-            ffi::AHardwareBuffer_lock(self.as_ptr(), usage.0, fence, rect, res)
+            ffi::AHardwareBuffer_lock(self.as_ptr(), usage.0 .0, fence, rect, res)
```
msiglreith, line 276: these spaces between the `.0`s look a bit weird to me, was this intentional as `cargo fmt` doesnt seem to complain?

msiglreith (approved):

ndk/src/hardware_buffer.rs
MarijnS95: @msiglreith It was actually `cargo fmt` that did this, perhaps to prevent some possible mis-interpretation of `0.0`?

We run `cargo fmt` in the CI and it's green, so it doesn't complain :)

PR #319, opened 2022-07-15
Update raw-window-handle to v0.5.0

kchibisov:


ndk/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - Fixed `HardwareBuffer` leak on buffers returned from `AndroidBitmap::get_hardware_buffer()`. (#296)
 - **Breaking:** Update `jni` crate (used in public API) from `0.18` to `0.19`. (#300)
 - hardware_buffer: Made `HardwareBufferDesc` fields `pub`. (#313)
+- **Breaking:** Update `raw-window-handle` to `v0.5` removing `HasRawWindowHandle` methods.
```
MarijnS95, line 22: Not a breaking change because it was only added in the same release, see:

```
- Implement `HasRawWindowHandle` directly on `NativeWindow`. (#274)
```

right above.

Can you add `, #319` to it though?

ndk/src/native_window.rs
```diff
@@ -37,14 +37,6 @@ impl Clone for NativeWindow {
     }
 }
 
-unsafe impl HasRawWindowHandle for NativeWindow {
```
MarijnS95, line 40: Why/when has the trait been removed?

MarijnS95: https://github.com/rust-windowing/raw-window-handle/blob/0ee134f812cc7191cc2ee33a1fdf2d266f0cf1e4/src/lib.rs#L60 It's still there on master?

MarijnS95: https://github.com/rust-windowing/raw-window-handle/blob/master/CHANGELOG.md#050-2022-07-14 This is not in the changelog for `0.5.0` either.

kchibisov: This trait is a compatibility for raw window handle `v0.3`, there's no need for it with `v0.5.0` it was left to simplify transition, but there's no need in it, since the handle is cross platform.

ndk/CHANGELOG.md
kchibisov: Sure.

ndk/CHANGELOG.md
```diff
@@ -6,7 +6,7 @@
 - **Breaking:** Upgrade to [`ndk-sys 0.4.0`](../ndk-sys/CHANGELOG.md#040-TODO-YET-UNRELEASED) and use new `enum` newtype wrappers. (#245)
 - native_window: Use `release`/`acquire` for `Drop` and `Clone` respectively. (#207)
 - **Breaking:** audio: Rename from `aaudio` to `audio` and drop `A` prefix. (#273)
-- Implement `HasRawWindowHandle` directly on `NativeWindow`. (#274)
+- Implement `raw_window_handle` directly on `NativeWindow`. (#319)
```
MarijnS95, line 9:
```suggestion
- Implement `raw_window_handle` directly on `NativeWindow`. (#273, #319)
```

kchibisov: I don't think that we need a first PR hash here given that it  this invalidates the old one, no?

ndk/src/native_window.rs
kchibisov: Hm, I see. Will update.

kchibisov:
Updated. Should be good now.

MarijnS95 (approved):
Thanks, perfect!

PR #320, opened 2022-07-18
ndk: Remove "irrelevant" `hardware_buffer`/`trace` feature in favour of API level

### LINKED ISSUE #318: Replace `hardware_buffer` and `trace` features with their respective `api-level-23/26`
Features thus far seem to be used specifically in `ndk-sys` to prevent linking against additional libraries, for content that is provided outside of `libandroid.so`:

https://github.com/rust-windowing/android-ndk-rs/blob/d104d01f116b537babdc94d8c9889581f8d63a02/ndk-sys/src/lib.rs#L43-L53

However, `hardware_buffer` and `trace` have their content available directly in `libandroid.so` - making these features only guard Rust code, in my eyes unnecessarily. It seems much more clear to enable everything `libandroid.so` has to offer solely when a certain API level is turned on. After all, we don't replicate this feature guarding pattern for other "subsystems" of the Android API, only for tracing and hardware_buffer.

@JasperDeSutter What's your thought on this?

MarijnS95:
Fixes #318

Thus far feature flags have been used specifically in `ndk-sys` to control what native libraries are linked to, for content that is provided outside of the standard linked `libandroid.so`:

https://github.com/rust-windowing/android-ndk-rs/blob/d104d01f116b537babdc94d8c9889581f8d63a02/ndk-sys/src/lib.rs#L43-L53

However, `hardware_buffer` and `trace` are available in `libandroid.so` directly making these features only guard our Rust bindings.  This is unnecessarily tedious and inconsistent with the rest of the NDK wrappers which are only guarded by their minimum API level.

### DIFF
--- a/ndk-examples/Cargo.toml
+++ b/ndk-examples/Cargo.toml
@@ -9,7 +9,7 @@ publish = false
 jni = "0.19"
 libc = "0.2"
 log = "0.4.14"
-ndk = { path = "../ndk", features = ["trace"] }
+ndk = { path = "../ndk", features = ["api-level-23"] }
 ndk-context = { path = "../ndk-context" }
 ndk-glue = { path = "../ndk-glue", features = ["logger"] }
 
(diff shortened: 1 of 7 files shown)


dvc94ch (approved):
Lgtm

PR #328, opened 2022-08-17
ndk-build: Add an opt-in alternative to adb install 

lovelyyfiaaa:
Sometimes, I ocassionally find that adb install takes too long to install!!
This feature is a lovely workaround to that issue. <3

Instead of using adb install, we manually push the APK file to the device.
The location is in /data/local/tmp/.
Then, we manually install the app with pm install~!
After that, we remove the apk from the install location.

As this is just a workaround for now, it needs to be enabled by adding the CARGO_APK_MANUAL_INSTALL=1 environment variable.

Some questions though!!
- Do we want to use `CARGO_APK_MANUAL_INSTALL=1` or a flag like `--manual-backend`?
- Should `CARGO_APK_MANUAL_INSTALL=1` be rather an argument in the install() function that will be set by cargo-apk?
 Soo we will find this
```rs
   pub fn install(&self, manual_install: bool) -> Result<(), NdkError> {
      ...
   }
```

and..
```rs
 pub fn run(&self, artifact: &Artifact) -> Result<(), Error> {
        let apk = self.build(artifact)?;
        apk.install(std::env::var("CARGO_APK_MANUAL_INSTALL")
            .map(|var| var == "1")
            .unwrap_or(false))?;
        apk.start()?;
        Ok(())
    }

```

### DIFF
--- a/ndk-build/src/apk.rs
+++ b/ndk-build/src/apk.rs
@@ -177,10 +177,46 @@ impl Apk {
     }
 
     pub fn install(&self) -> Result<(), NdkError> {
-        let mut adb = self.ndk.platform_tool(bin!("adb"))?;
-        adb.arg("install").arg("-r").arg(&self.path);
-        if !adb.status()?.success() {
-            return Err(NdkError::CmdFailed(adb));
+        if std::env::var("CARGO_APK_MANUAL_INSTALL")
+            .map(|var| var == "1")
+            .unwrap_or(false)
+        {
+            let dist_path = String::from("/data/local/tmp/")
+                + &self.path.file_name().unwrap().to_str().unwrap();
+            println!("Pushing {:?} to the device", &self.path);
+            {
+                let mut cmd = self.ndk.platform_tool(bin!("adb"))?;
+
+                cmd.arg("push").arg(&self.path).arg(&dist_path);
+                if !cmd.status()?.success() {
+                    return Err(NdkError::CmdFailed(cmd));
+                }
+            }
+            println!("Installing application");
+            {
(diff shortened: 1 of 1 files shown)


MarijnS95:
There's supposed to be an `adb install` flag that does exactly this, have you tried that? I'm not near a computer so can't look up the output of `--help`.

lovelyyfiaaa:
> There's supposed to be an `adb install` flag that does exactly this, have you tried that? I'm not near a computer so can't look up the output of `--help`.

Yupp!! I'm aware of that..

But this hack works everytime for me.. while I don't really have a nice experience with ADB Install frontends.

lovelyyfiaaa:
I assume `--no-streaming` is the one I'm looking for?

lovelyyfiaaa:
Umm.. as I said, this is just a lovely workaround should one liner ADB doesn't work!!

dvc94ch:
I don't understand why one should be faster than the other. Can you provide evidence that that is the case and a hypothesis why this is?

lovelyyfiaaa:
I guess, maybe it's better to close this PR for now? I think having this never-merged PR existing itself is already nice, since people can use the patch if things go wrong.

dvc94ch:
just think it would help to decide what to do if the problem is better understood. I was of the impression that adb install works exactly like your workaround.

MarijnS95:
> Yupp!! I'm aware of that..
> 
> But this hack works everytime for me.. while I don't really have a nice experience with ADB Install frontends.

Same question as @dvc94ch: without any more context/logs/explanation it is really hard to justify a bunch of extra code to maintain, we need to understand _why_ all those `adb` flags to tweak the install process don't solve whatever vague issue you're facing.

PR #329, opened 2022-08-20
Add --device arg to be able to run on specific android devices

Jasper-Bekkers:
This feels quite hacky, but it doesn't look like cargo_subcommand has adequate functionality to filter or add new arguments to some of the commands. 

This adds support for a `--device` arg in both `cargo apk run` and `cargo apk gdb` to run on a specific device whenever one is present.

### DIFF
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Add `--device` argument to select `adb` device by serial (see `adb devices` for connected devices and their serial). ([#329](https://github.com/rust-windowing/android-ndk-rs/pull/329))
+
 # 0.9.3 (2022-07-05)
 
 - Allow configuration of alternate debug keystore location; require keystore location for release builds. ([#299](https://github.com/rust-windowing/android-ndk-rs/pull/299))
--- a/cargo-apk/src/apk.rs
+++ b/cargo-apk/src/apk.rs
@@ -234,15 +234,15 @@ impl<'a> ApkBuilder<'a> {
         Ok(apk.align()?.sign(signing_key)?)
     }
 
-    pub fn run(&self, artifact: &Artifact) -> Result<(), Error> {
+    pub fn run(&self, artifact: &Artifact, device_name: Option<String>) -> Result<(), Error> {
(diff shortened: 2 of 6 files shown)


MarijnS95:
> This feels quite hacky, but it doesn't look like cargo_subcommand has adequate functionality to filter or add new arguments to some of the commands.

This is exactly what `|_, _| Ok(false)` exists for. Granted that's a bit of an ugly API which is why we replaced it with `clap` in https://github.com/dvc94ch/cargo-subcommand/commit/6c77c51, allowing you to write the usual:

```rust
#[derive(Parser)]
struct CargoApkArgs {
    #[clap(flatten)]
    pub subcommand_args: cargo_subcommand::Args,
    /// Device serial to use, from `adb devices`
    #[clap(short, long)]
    pub device: Option<String>,
}
```

@dvc94ch Was there anything pending for `cargo-subcommand` that is blocking a release? Or were we pending for a rather substantial change here in `cargo-apk` to migrate to the new/improved API?

MarijnS95:
> @dvc94ch Was there anything pending for `cargo-subcommand` that is blocking a release? Or were we pending for a rather substantial change here in `cargo-apk` to migrate to the new/improved API?

Found it, it got stale while we shifted to `x`: https://github.com/rust-windowing/android-ndk-rs/pull/238

MarijnS95 (approved):

dvc94ch:
Hey @MarijnS95, sorry for being absent. Started a new job. Let me know if you need anything like repo access, if you send an email I'll be more likely to read it than if it's a gh notification.

MarijnS95:
@dvc94ch No worries, we already decided that PR would go stale in favour of X, though I might pick it up at some point to keep `cargo-apk` in a somewhat neat state.

I/we have also been super busy with not much time left for these tools, which doesn't really matter since they (both `cargo-apk` and `xbuild`, we use them both!) have been working out just fine for us besides the odd maintenance and usability improvements.

No need for anything else, I have repo access to both `xbuld` and `android-ndk-rs`, and even have publisher access for the latter (https://github.com/orgs/rust-windowing/teams/publishers) in case I need to yank a crate (other than that, publishing is automated on GitHub :tada:).

PR #331, opened 2022-08-21
ndk-build: Return PID after launching the app

Jasper-Bekkers:
Depends on  #329 to apply cleanly.

Shows the process PID after launching. Requires `-W` as an `am start` arg to wait for process completion to prevent a race condition. Having the PID is useful in a lot of cases, such as filtering logs with `adb logcat --pid=...`

### DIFF
--- a/ndk-build/src/apk.rs
+++ b/ndk-build/src/apk.rs
@@ -200,19 +200,37 @@ impl Apk {
         Ok(())
     }
 
-    pub fn start(&self, device_serial: Option<&str>) -> Result<(), NdkError> {
-        let mut adb = self.ndk.adb(device_serial)?;
-
-        adb.arg("shell")
+    pub fn start(&self, device_serial: Option<&str>) -> Result<u32, NdkError> {
+        let mut am_start = self.ndk.adb(device_serial)?;
+        am_start
+            .arg("shell")
             .arg("am")
             .arg("start")
+            .arg("-W")
             .arg("-a")
             .arg("android.intent.action.MAIN")
             .arg("-n")
             .arg(format!("{}/android.app.NativeActivity", &self.package_name));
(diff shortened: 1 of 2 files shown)


ndk-build/src/apk.rs
```diff
@@ -176,27 +176,57 @@ impl Apk {
...
 
-    pub fn start(&self) -> Result<(), NdkError> {
-        let mut adb = self.ndk.platform_tool(bin!("adb"))?;
-        adb.arg("shell")
+    pub fn start(&self, device_name: Option<String>) -> Result<(), NdkError> {
+        let mut am_start = self.shell(&device_name)?;
+        am_start
             .arg("am")
             .arg("start")
+            .arg("-W")
```
MarijnS95, line 198: That's neat, I was wondering how this was implemented by @dvc94ch in `x`, turns out to be a sleep-loop: https://github.com/cloudpeers/xbuild/blob/e0ebb50bfa78cfad61931bb30c70a624b85fdddd/xbuild/src/devices/adb.rs#L175-L186

MarijnS95: It's probably semi-smart to keep a loop so that duplicate PIDs are not causing issues:

https://github.com/cloudpeers/xbuild/blob/e58cc09b4f2cb07c1fa0591f9d65cea435dd5646/xbuild/src/devices/adb.rs#L179

But IMO `am start` should be able to emit the `pid` instead of this hackery.

MarijnS95 (approved):
I think the pid thing is good enough for now. Was seeing duplicate PIDs because one of my `gdb` sessions got stuck and the processes couldn't get killed and detached even under root.

If we ever observe problems with it we'll make it more fallible or change it altogether.

PR #332, opened 2022-08-21
cargo-apk: Print and follow `adb logcat` output after starting app

Jasper-Bekkers:
Depends on #331

This starts a logcat session for our PID so we have something that's a bit more akin to how `cargo run` regularly works. If this is undesired, I would also be OK with adding a `--log` argument.

### DIFF
--- a/.github/workflows/android_test.sh
+++ b/.github/workflows/android_test.sh
@@ -10,7 +10,7 @@ adb uninstall rust.example.hello_world || true
 
 if [ -z "$1" ];
 then
-    cargo apk run -p ndk-examples --target x86_64-linux-android --example hello_world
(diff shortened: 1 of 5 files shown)


Jasper-Bekkers:
@MarijnS95 I suspect the CI script to potentially require an update after this, but I'm not entirely sure. Looks like it got stuck as well after about ~5 hours.

MarijnS95 (approved):
Good to have this by default, I added a `--no-logcat` option to back out of this, for example in the CI.

(We could/should also use the `cargo apk run` `logcat` output in CI to test if the right strings are printed, but there's currently no neat way to pass the `-d` option to `logcat` so that it exits pretty soon. At the same time that doesn't guarantee the app has properly reached the print statement, but the CI has always been running on good faith given that `-W` was never passed to `am start` either yet the app always started in time.)

PR #338, opened 2022-09-05
ndk-build: Support `android:resizeableActivity`

RagnarListig:
This is a trivial pull request to add support for the [`android:resizeableActivity`](https://developer.android.com/guide/topics/manifest/activity-element#resizeableActivity) attribute on `<activity>`, needed e.g. for [manifests for the Oculus Quest](https://developer.oculus.com/documentation/native/android/mobile-native-manifest/) VR headset.

### DIFF
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Upgrade to latest `ndk-build` to deduplicate libraries before packaging them into the APK. ([#333](https://github.com/rust-windowing/android-ndk-rs/pull/333))
+- Support `android:resizeableActivity`.
 
 # 0.9.3 (2022-07-05)
 
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Unreleased
 
 - Upgrade to latest `ndk-build` to deduplicate libraries before packaging them into the APK. ([#333](https://github.com/rust-windowing/android-ndk-rs/pull/333))
-- Support `android:resizeableActivity`.
+- Support `android:resizeableActivity`. ([#338](https://github.com/rust-windowing/android-ndk-rs/pull/338))
 
 # 0.9.3 (2022-07-05)
 
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -159,6 +159,11 @@ orientation = "landscape"
 # Unset by default, or "true" when targeting Android >= 31 (S and up).
 exported = "true"
 
+# See https://developer.android.com/guide/topics/manifest/activity-element#resizeableActivity
+#
+# Unset by default, or "true" when targeting Android >= 24 (Nougat and up).
+resizeable_activity = "true"
+
 # See https://developer.android.com/guide/topics/manifest/meta-data-element
 #
 # Note: there can be several .meta_data entries.
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -156,13 +156,13 @@ orientation = "landscape"
 
 # See https://developer.android.com/guide/topics/manifest/activity-element#exported
 #
-# Unset by default, or "true" when targeting Android >= 31 (S and up).
-exported = "true"
+# Unset by default, or true when targeting Android >= 31 (S and up).
+exported = true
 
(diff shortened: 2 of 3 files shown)


cargo-apk/README.md
```diff
@@ -159,6 +159,11 @@ orientation = "landscape"
 # Unset by default, or "true" when targeting Android >= 31 (S and up).
 exported = "true"
 
+# See https://developer.android.com/guide/topics/manifest/activity-element#resizeableActivity
+#
+# Unset by default, or "true" when targeting Android >= 24 (Nougat and up).
+resizeable_activity = "true"
```
MarijnS95, line 165: Wondering if this - and `exported = "true"` - should be the TOML literal `true` instead of the string `"true"`, just like `required = true` under `[[package.metadata.android.uses_feature]]`.

cargo-apk/README.md
```diff
@@ -159,6 +159,11 @@ orientation = "landscape"
 # Unset by default, or "true" when targeting Android >= 31 (S and up).
 exported = "true"
 
+# See https://developer.android.com/guide/topics/manifest/activity-element#resizeableActivity
+#
+# Unset by default, or "true" when targeting Android >= 24 (Nougat and up).
```
MarijnS95, line 164: Not entirely true, `cargo-apk` doesn't set this value at all. The matching comment on `exported` above exists specifically to document this logic on the `cargo-apk` side:

https://github.com/rust-windowing/android-ndk-rs/blob/93558ca7b03f1818bb2bfc832f9b2bf52742b557/cargo-apk/src/apk.rs#L83-L88

Instead, what you described here is how Android _treats_ this field by default on Android >= 24.

cargo-apk/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Upgrade to latest `ndk-build` to deduplicate libraries before packaging them into the APK. ([#333](https://github.com/rust-windowing/android-ndk-rs/pull/333))
+- Support `android:resizeableActivity`.
```
MarijnS95, line 4:
```suggestion
- Support `android:resizeableActivity`. ([#338](https://github.com/rust-windowing/android-ndk-rs/pull/338))
```

cargo-apk/README.md
RagnarListig: Probably yes. I just used the other boolean attribute, `exported`, as a template. Shall I change both?

cargo-apk/README.md
RagnarListig: Ah, I see, that's true. I'll change the comment.

cargo-apk/README.md
RagnarListig: Fixed in c388b497bc146e8d625b20f30118d266a1a82007. I also tested that the non-string values work fine as documented in the README.

RagnarListig:
@MarijnS95 Thanks for your review. I've addressed your feedback, please have another look.

cargo-apk/README.md
```diff
@@ -156,8 +156,13 @@ orientation = "landscape"
...
 # See https://developer.android.com/guide/topics/manifest/activity-element#exported
 #
-# Unset by default, or "true" when targeting Android >= 31 (S and up).
-exported = "true"
+# Unset by default, or true when targeting Android >= 31 (S and up).
+exported = true
+
+# See https://developer.android.com/guide/topics/manifest/activity-element#resizeableActivity
+#
+# Defaults to true on Android >= 24, no effect on earlier API levels
```
MarijnS95, line 164: There's another detail on Android >=31, but it seems a bit much to describe here.

MarijnS95 (approved):
Thanks!

cargo-apk/README.md
MarijnS95: Perfect, thanks a lot for fixing `exported` too!

RagnarListig:
Incidentally, looks like users have already previously (incorrectly) assumed that this attribute was supported:

https://github.com/Ralith/openxrs/blob/05a0f442d93e627bf45b341100ad09c621d6ab06/openxr/Cargo.toml#L61

MarijnS95:
@RagnarListig I don't fully remember if the previous implementation serialized all tables parsed from TOML into XML, which won't suffer from that. Regardless, that's what you sometimes get with external contributions :grimacing:

PR #341, opened 2022-09-08
cargo-apk: Pass device serial to `ndk-gdb`

MarijnS95:
`ndk-gdb` must use the same device as `.run()` previously launched the app on.


MarijnS95:
Cc @Jasper-Bekkers :grimacing:

Jasper-Bekkers (approved):

PR #342, opened 2022-09-08
cargo-apk: Force-disable logcat in `fn run()` when using `gdb`

MarijnS95:
The terminal must be available for GDB to start and interact with, instead of indefinitely blocking on a running instance of `logcat`.


MarijnS95:
Also CC @Jasper-Bekkers, I thought `ndk-gdb` took care of everything but turns out we have to start the app first and let it attach in hindsight.

Jasper-Bekkers (approved):

PR #343, opened 2022-09-09
ndk-build: Provide `adb` path to `ndk-gdb`

MarijnS95:
`adb` is not necessarily available in `PATH`, hence why `cargo-apk` and `ndk-build` detect and use this executable from the SDK path.  We should do the same for `ndk-gdb` which strangely doesn't search for `adb` within the SDK it resides in.


MarijnS95:
Tested on Windows and Linux.

PR #344, opened 2022-09-09
ndk-build: Remove unparsable quotes from `Android.mk` for `ndk-gdb`

MarijnS95:
On Windows it seems the `ndk-gdb.py` script that invokes `make` to dump the variables from `Android.mk` fails to strip quotation marks from the file correctly, resulting in erroneous output like (notice double quotes at the end of the path):

	OSError: [WinError 123] The filename, directory name, or volume
	label syntax is incorrect:
	'android-ndk-rs\\target\\debug\\apk\\examples\\""'

And:

	Application cannot run on the selected device.
	Application ABIs: "arm64-v8a"
	Device ABIs: arm64-v8a, armeabi-v7a, armeabi

Solve this by removing the quotes, allowing `ndk-gdb` to launch sucessfully.


MarijnS95:
Tested on Windows and Linux.

Issue #327, opened 2022-08-16
Is the wrong environment variable being stated as deprecated?


Hoodad:
Hello! 

As noted in the title if you have the environment variable `ANDROID_HOME` set you are greeted with a warning telling you to use `ANDROID_SDK_ROOT` instead and remove the `ANDROID_HOME` variable. However from reading the [Android documentation](https://developer.android.com/studio/command-line/variables) I get the impression its the opposite relation, that `ANDROID_SDK_ROOT` is deprecated and shall not be used? 

From the Android documentation.
> ANDROID_HOME | Sets the path to the SDK installation directory. Once set, the value does not typically change, and can be shared by multiple users on the same machine. ANDROID_SDK_ROOT, which also points to the SDK installation directory, is deprecated. If you continue to use it, Android Studio and the Android Gradle plugin will check that the old and new variables are consistent.

The error in question can be found here, https://github.com/rust-windowing/android-ndk-rs/blob/caeb591038a4c3642188cd17db19dca69a2f9086/ndk-build/src/ndk.rs#L21-L23

MarijnS95:
https://github.com/rust-windowing/android-ndk-rs/pull/39

That's what you get with PRs that fail to quote any reference material.

PR #346, opened 2022-09-12
ndk-build: Consider ANDROID_SDK_ROOT as deprecated instead of ANDROID_HOME

MarijnS95:
Fixes https://github.com/rust-windowing/android-ndk-rs/issues/327, CC @Hoodad

According to the [official documentation] `ndk-build` is considering the wrong variable as deprecated.  The wrong message was introduced in [#39] which bizarrely failed to quote any reference matterial on the matter, making it impossible to understand whether this has always been the case or recently switched in precedence.

[official documentation]: https://developer.android.com/studio/command-line/variables
[#39]: https://github.com/rust-windowing/android-ndk-rs/pull/39


Hoodad:
Thank you for fixing it 👍🙏

msiglreith (approved):

PR #348, opened 2022-09-15
cargo-apk: Automate `adb reverse` port forwarding through `Cargo.toml` metadata

Jasper-Bekkers:
This adds a feature so we can set up reverse port forwarding from our Cargo.toml file before the app starts

```toml
[package.metadata.android.reverse_port_fwd]
"tcp:1338" = "tcp:1338"
"tcp:1339" = "tcp:1339"
```

### DIFF
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -192,6 +192,12 @@ port = "8080"
 path = "/rust-windowing/android-ndk-rs/tree/master/cargo-apk"
 path_prefix = "/rust-windowing/"
 mime_type = "image/jpeg"
+
+# Set up reverse port forwarding through `adb reverse`, meaning that if the
+# Android device connects to `https://localhost:1338` it will be routed to 
+# the localhost on the host device instead.
+[package.metadata.android.reverse_port_fwd]
+"tcp:1338" = "tcp:1338"
 ```
 
 If a manifest attribute is not supported by `cargo apk` feel free to create a PR that adds the missing attribute.
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -194,8 +194,9 @@ path_prefix = "/rust-windowing/"
 mime_type = "image/jpeg"
 
 # Set up reverse port forwarding through `adb reverse`, meaning that if the
-# Android device connects to `https://localhost:1338` it will be routed to 
-# the localhost on the host device instead.
+# Android device connects to `localhost` on port `1338` it will be routed to 
+# the host on port `1338` instead. Source and destination ports can differ,
+# see the `adb` help page for possible configurations.
 [package.metadata.android.reverse_port_forward]
 "tcp:1338" = "tcp:1338"
 ```
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -196,7 +196,7 @@ mime_type = "image/jpeg"
 # Set up reverse port forwarding through `adb reverse`, meaning that if the
 # Android device connects to `https://localhost:1338` it will be routed to 
 # the localhost on the host device instead.
-[package.metadata.android.reverse_port_fwd]
+[package.metadata.android.reverse_port_forward]
 "tcp:1338" = "tcp:1338"
 ```
 
--- a/cargo-apk/src/manifest.rs
+++ b/cargo-apk/src/manifest.rs
@@ -73,7 +73,7 @@ struct AndroidMetadata {
     /// Maps profiles to keystores
     #[serde(default)]
     signing: HashMap<String, Signing>,
-    /// Set up reverse port forwarding before the application launches
+    /// Set up reverse port forwarding before launching the application
     #[serde(default)]
     reverse_port_forward: HashMap<String, String>,
 }
--- a/cargo-apk/src/manifest.rs
+++ b/cargo-apk/src/manifest.rs
@@ -17,7 +17,7 @@ pub(crate) struct Manifest {
     pub(crate) runtime_libs: Option<PathBuf>,
     /// Maps profiles to keystores
     pub(crate) signing: HashMap<String, Signing>,
-    pub(crate) reverse_port_fwd: HashMap<String, String>,
+    pub(crate) reverse_port_forward: HashMap<String, String>,
 }
 
 impl Manifest {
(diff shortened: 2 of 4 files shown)


msiglreith (commented):
Would be nice to have the example from the readme added in our [manifest example](https://github.com/rust-windowing/android-ndk-rs/blob/master/cargo-apk/README.md#manifest)

Jasper-Bekkers:
Updated the readme in [4638053](https://github.com/rust-windowing/android-ndk-rs/pull/348/commits/46380537fe37a89ba72fc65a402941028872fc08)

cargo-apk/src/manifest.rs
```diff
@@ -71,6 +73,9 @@ struct AndroidMetadata {
     /// Maps profiles to keystores
     #[serde(default)]
     signing: HashMap<String, Signing>,
+    /// Set up reverse port forwarding before the application launches
+    #[serde(default)]
+    reverse_port_fwd: HashMap<String, String>,
```
MarijnS95, line 78: `adb` calls these `reverse socket connections` and `forward socket connections`, and doesn't use the "port forwarding" terminology in the `reverse` docs - can we do the same?

Otherwise at least spell out `fwd` :grimacing:

cargo-apk/README.md
```diff
@@ -192,6 +192,12 @@ port = "8080"
 path = "/rust-windowing/android-ndk-rs/tree/master/cargo-apk"
 path_prefix = "/rust-windowing/"
 mime_type = "image/jpeg"
+
+# Set up reverse port forwarding through `adb reverse`, meaning that if the
+# Android device connects to `https://localhost:1338` it will be routed to 
+# the localhost on the host device instead.
```
MarijnS95, line 198: This isn't strictly for `http`, I'd write ``Android device connects to `localhost` on port `1338` ...``.

For demonstration purposes, can we use a different source and destination purposes? e.g. ``it will be routed to the host on port `XYZ whatever you choose` instead.``

cargo-apk/src/manifest.rs
MarijnS95: While at it, I think it's only a minimal stretch to also implement `adb forward` as part of this PR :grimacing:

ndk-build/src/apk.rs
```diff
@@ -187,9 +190,29 @@ impl Apk {
...
     }
 
+    pub fn reverse_port_forwarding(&self, device_serial: Option<&str>) -> Result<(), NdkError> {
+        for (from, to) in &self.reverse_port_fwd {
+            println!("Reverse port forwarding {} {}", from, to);
+            let mut adb = self.ndk.platform_tool(bin!("adb"))?;
+
+            if let Some(device_serial) = device_serial {
+                adb.arg("-s").arg(device_serial);
+            }
```
MarijnS95, line 204:
```suggestion
            let mut adb = self.ndk.adb(device_serial)?;
```

cargo-apk/src/manifest.rs
```diff
@@ -71,6 +73,9 @@ struct AndroidMetadata {
     /// Maps profiles to keystores
     #[serde(default)]
     signing: HashMap<String, Signing>,
+    /// Set up reverse port forwarding before the application launches
```
MarijnS95, line 76:
```suggestion
    /// Set up reverse port forwarding before launching the application
```

MarijnS95:
I still don't feel entirely confident about having this in `adb`, we're just building out more and more passthrough commands for `adb` that don't have much in common with `apk` building anymore, _but I don't have any other alternative :sob:_ (i.e. a server on the host that makes itself visible to connected `adb` devices instead, but `adb wait-for-device` isn't powerful enough). A scriptable pre-apk-start might solve many such things but isn't platform-agnostic either :(

cargo-apk/src/manifest.rs
Jasper-Bekkers: > While at it, I think it's only a minimal stretch to also implement adb forward as part of this PR 😬

I don't think that belong in this PR tbh, it's a different feature that I have no means of testing.

Jasper-Bekkers: > adb calls these reverse socket connections and forward socket connections, and doesn't use the "port forwarding" terminology in the reverse docs - can we do the same?

Regardless of how adb calls it, reverse port forwarding is the normal terminology for this. I can change it but I don't think it makes it more clear.

cargo-apk/src/manifest.rs
Jasper-Bekkers: I don't understand the correction, both sentences describe the exact same thing in the exact same way?

Jasper-Bekkers:
> I still don't feel entirely confident about having this in adb, we're just building out more and more passthrough commands for adb that don't have much in common with apk building anymore

For us, in practice, it means a significantly smoother and less error prone workflow for launching apk's, regardless of how much it has to do with building an apk or not, for us it's fundamental to running one which is why I'm proposing adding this to this tool.

cargo-apk/src/manifest.rs
MarijnS95: This way of wording makes it ever so slightly more clear that `cargo-apk` is (obviously) in control of launch_ing_ the application straight after. It is by all means a nit though, but I find it more clear.

MarijnS95:
Sure, and as said above we don't really have any clean/proper alternative. At the same time we're one of the very few actively contributing to `cargo-apk` so I think we can bend it towards our use-cases somewhat.

I'll apply the nits above and merge this, not sure when I'll push out a new (patch) release though (`cargo install --git` exists, fortunately).

cargo-apk/README.md
MarijnS95, line 198:
```suggestion
# Set up reverse port forwarding through `adb reverse`, meaning that if the
# Android device connects to `localhost` on port `1338` it will be routed to 
# the host on port `1338` instead. Source and destination ports can differ,
# see the `adb` help page for possible configurations.
```

ndk-build/src/apk.rs
```diff
@@ -187,9 +190,25 @@ impl Apk {
             path: config.apk(),
             package_name: config.manifest.package.clone(),
             ndk,
+            reverse_port_forward: config.reverse_port_forward.clone(),
         }
     }
 
+    pub fn reverse_port_forwarding(&self, device_serial: Option<&str>) -> Result<(), NdkError> {
+        for (from, to) in &self.reverse_port_forward {
+            println!("Reverse port forwarding {} {}", from, to);
```
MarijnS95, line 199:
```suggestion
            println!("Reverse port forwarding from {} to {}", from, to);
            
```

MarijnS95 (approved):

PR #352, opened 2022-09-27
Add 300ms sleep before calling apk.start.

### LINKED ISSUE #351: Error: String `` is not a PID
```
Verification succesful
Performing Incremental Install
Serving...
Unknown command: install-incremental
Performing Streamed Install
Success
Starting: Intent { act=android.intent.action.MAIN cmp=rust.openxr_test/android.app.NativeActivity }
Status: ok
LaunchState: COLD
Activity: com.oculus.shellenv/.ShellEnvActivity
TotalTime: 23
WaitTime: 24
Complete
Error: String `` is not a PID

Caused by:
    cannot parse integer from empty string
```

`cargo apk run` succeeds sometimes, but I run into this error frequently (>50% of the time).  Repeatedly running the command will eventually lead to success.  Appears to be a race condition where `apk.start` is called before the system is ready to start the app and return a PID, leading to a failure on this line: https://github.com/rust-windowing/android-ndk-rs/blob/master/cargo-apk/src/apk.rs#L251

jingchan:
Adding a short delay here appears to fix #351 at least for my system.

First time contributing, so please let me know if there are other conventions I should follow.

Thanks!

### DIFF
--- a/cargo-apk/src/apk.rs
+++ b/cargo-apk/src/apk.rs
@@ -9,6 +9,8 @@ use ndk_build::manifest::{IntentFilter, MetaData};
 use ndk_build::ndk::{Key, Ndk};
 use ndk_build::target::Target;
 use std::path::PathBuf;
+use std::thread;
+use std::time;
 
 pub struct ApkBuilder<'a> {
     cmd: &'a Subcommand,
@@ -239,6 +241,10 @@ impl<'a> ApkBuilder<'a> {
     pub fn run(&self, artifact: &Artifact, no_logcat: bool) -> Result<(), Error> {
         let apk = self.build(artifact)?;
         apk.install(self.device_serial.as_deref())?;
+
+        // Delay 300ms while waiting for process ID to become available.
+        thread::sleep(time::Duration::from_millis(300));
+
         let pid = apk.start(self.device_serial.as_deref())?;
 
         if !no_logcat {


Jasper-Bekkers:
The commit seems empty, though I don't think adding a 300ms delay to fix a race condition is a proper way to fix these things.

MarijnS95:
The app is started with `am start -W` which _should_ guarantee that is has started? If that isn't the case we might want to spinloop on `pidof` until a given timeout - let's check how `ndk-gdb` does it instead of hand-rolling our own solution (300ms is after all very long _and_ arbitrary).

Jasper-Bekkers:
One thing that occurred to me - we don't strictly need the PID, it's just to output the logcat results. If we can wrangle logcat to filter based on activity name we might also be fine (though - we may also introduce problems if we have multiple of the same activities running).

MarijnS95:
@Jasper-Bekkers It doesn't seem like the `logcat` buffer includes the package name (or activity name), there's only log tags which are application-defined and impossible for us to guess (set to `RustStdoutStderr` for the forwarding code in `ndk-glue`, and our own application uses module names through the `log` facade).

All apps run under a unique user (`UID`), and that's included in the logcat buffer. Our app tends to switch UIDs though (perhaps because it is being reinstalled, e.g. when switching debug signatures) but at least `logcat --uid` supports a comma-separated list. We just need a clean way to process `pm dump rust.<appname>`.

Alas, Google has already solved this problem for Android Studio - we just have to look up how they solve this problem instead of reinventing the wheel.

jingchan:
Sorry for the empty commit- I believe it's resolved now.  Thanks for being patient!

I'd agree a more proper solution is warranted here, though I'm somewhat unfamiliar with the rust (and android-rust) ecosystem, so I'd appreciate some guidance along the way.

For reference, this is how Google achieves this in their bazel tooling, which I would presume could be considered close to if not the "gold standard" for Android deployment (this is likely similar or the same as is done internally): https://github.com/bazelbuild/bazel/blob/f65a0d6a87619022de5a6a28ddd7a30ea3cf3186/tools/android/incremental_install.py#L785-L788

MarijnS95:
@jingchan Are you sure this change change solves the problem? You inserted the timeout in between `apk.install()` and `apk.start()`, meaning that the Activity won't get launched (but the install _might_ start up the process) before the timeout.

Instead, should the timeout reside in between `am start -W` and `pidof`? In #351 you mention "Appears to be a race condition where `apk.start` is called before the system is ready to start the app and return a PID" yet the logs seem to indicate that the activity was started successfully - which is strange given that it should be running after that, thanks to `-W`.

Do you think the installation/app-replacement is still running in the background at that point, causing `am start` to start an older instance of the `Activity` which is killed pretty much immediately after, when the installation completes?

---

Alas, the Bazel build script doesn't seem to wait for the activity to start nor tries to acquire a PID, so it's different.

For testing, can you test if the `Activity` _always_ launches correctly after commenting out the `pidof` and `logcat` invocations? I'm curious if the app doesn't start at all in your case (as this always worked before, AFAIK) or if it's merely an issue with the freshly-added `pidof`+`logcat` logic.
In other words, even if `cargo apk run` fails on `Error: String `` is not a PID`, is the app still launched?

MarijnS95:
Closing for not seeing any followup on the questions and suggestions posed above. Besides, the `cargo-apk` portion of this repo migrated to https://github.com/rust-mobile/cargo-apk :tada:

PR #353, opened 2022-10-04
ndk: Add AMidi interface

paxbun:
Just noticed that `ndk` doesn't have an interface for AMidi, so I made one. Please let me know if there're any tests or fixes I have to do before merging!


ndk/src/midi.rs
```diff
@@ -0,0 +1,271 @@
+//! Bindings for [`AMidiDevice`], [`AMidiInputPort`], and [`AMidiOutputPort`]
+//!
+//! See [the NDK guide](https://developer.android.com/ndk/guides/audio/midi) for
+//! design and usage instructions, and [the NDK reference](https://developer.android.com/ndk/reference/group/midi)
+//! for an API overview.
+//!
+//! [`AMidiDevice`]: https://developer.android.com/ndk/reference/group/midi#amididevice
+//! [`AMidiInputPort`]: https://developer.android.com/ndk/reference/group/midi#amidiinputport
+//! [`AMidiOutputPort`]: https://developer.android.com/ndk/reference/group/midi#amidioutputport
+#![cfg(feature = "media")]
```
MarijnS95, line 10: Shouldn't this live under the `media` module, given that it heavily ties into all its infrastructure and types?

ndk/src/midi.rs
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    /// Receives the next pending MIDI message.
+    ///
+    /// To retrieve all pending messages, the client should repeatedly call this method until it
+    /// returns `Some(_)`.
+    ///
+    /// Note that this is a non-blocking call. If there are no Midi messages are available, the
+    /// function returns `None` immediately (for 0 messages received).
+    pub fn receive(&self, buffer: &mut [u8]) -> Result<(MidiOpcode, i64)> {
```
MarijnS95, line 240: I'd ask to turn these monospace doc-comments into intradoc links, but also change them to `Ok`/`Err` to match the `Result` here.

ndk/src/midi.rs
```diff
@@ -0,0 +1,271 @@
...
+        } else {
+            Err(
+                NdkMediaError::from_status(ffi::media_status_t(num_output_ports as c_int))
+                    .unwrap_err(),
+            )
+        }
+    }
+
+    /// Gets the MIDI device type.
+    pub fn get_type(&self) -> Result<MidiDeviceType> {
```
MarijnS95, line 92: No `get_` prefix, please: https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

ndk/src/midi.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+impl MidiDevice {
+    fn from_ptr(inner: NonNull<ffi::AMidiDevice>) -> Self {
+        Self { inner }
+    }
+
+    fn as_ptr(&self) -> *mut ffi::AMidiDevice {
+        self.inner.as_ptr()
+    }
```
MarijnS95, line 53: Is this intentionally private? Looks like a copy-paste from `image_reader.rs`, all the other wrappers use a `pub unsafe` with some documentation explaining if ownership is taken or not.

paxbun: I referred `audio.rs`, but seems like `audio.rs` also does not comply with the rules you mentioned. I'll make a fix for this.

ndk/src/midi.rs
paxbun: Fixed!

ndk/src/midi.rs
paxbun: Fixed!

ndk/src/midi.rs
paxbun: I changed `from_ptr` into a `pub unsafe` function, do I have to do the same to functions in `audio.rs`?

ndk/src/midi.rs
paxbun: I also hesitated to move `midi.rs` to `media`, but since the Android docs describes `AMidi` as a separate module, I left it there. It also seems that `AMidi` does not need `mediasdk` to be linked, so I think it is better to define a new feature "midi" and make `media::{Result, construct, NdkMediaError}` to be compiled when either "media" or "midi" is enabled. But this needs a lot of `#[cfg(any(feature = "media", feature = "midi"))]`s in `media/mod.rs`; would there be a better solution?

ndk/src/midi.rs
paxbun: Do I have to do the same fix for other modules? Seems like other modules have similar problems as well...
image

ndk/src/midi.rs
paxbun: Great! Since it's before the dawn here, I'll make another commit for this in the morning. Seems like I may have to modify `ndk_sys/src/lib.rs` for `libamidi.so` as well. Thanks for the review!

paxbun: Fixed!

ndk/Cargo.toml
```diff
@@ -17,6 +17,7 @@ all = ["audio", "bitmap","media", "api-level-30"]
 audio = ["ffi/audio", "api-level-26"]
 bitmap = ["ffi/bitmap"]
 media = ["ffi/media"]
+midi = ["ffi/midi", "api-level-29"]
```
MarijnS95, line 20: Don't these exist since API level 23?

ndk-sys/src/lib.rs
```diff
@@ -51,3 +51,7 @@ extern "C" {}
 #[cfg(all(feature = "audio", target_os = "android"))]
 #[link(name = "aaudio")]
 extern "C" {}
+
+#[cfg(all(feature = "midi", target_os = "android"))]
+#[link(name = "amidi")]
+extern "C" {}
```
MarijnS95, line 57: Good catch!

ndk/src/midi.rs
```diff
@@ -0,0 +1,286 @@
...
+
+    /// Receives the next pending MIDI message.
+    ///
+    /// To retrieve all pending messages, the client should repeatedly call this method until it
+    /// returns [`Ok(MidiOpcode::NoMessage)`].
+    ///
+    /// Note that this is a non-blocking call. If there are no Midi messages are available, the
+    /// function returns [`Ok(MidiOpcode::NoMessage)`] immediately (for 0 messages received).
```
MarijnS95, line 254: Nit: this returns a tuple with a `timestamp`.

Is that timestamp still valid if there was `NoMessage`? If so, I think the timestamp should be moved into the `MidiOpcode` variants where it is considered valid/useful instead.

ndk/src/midi.rs
```diff
@@ -0,0 +1,286 @@
...
+        Self { inner }
+    }
+
+    fn as_ptr(&self) -> *mut ffi::AMidiDevice {
+        self.inner.as_ptr()
+    }
+
+    /// Connects a native Midi Device object to the associated Java MidiDevice object.
+    ///
+    /// Use this AMidiDevice to access the rest of the native MIDI API.
```
MarijnS95, line 62:
```suggestion
    /// Use this [`MidiDevice`] to access the rest of the native MIDI API.
```

And even better: replace `this` with `the returned`.

ndk/src/midi.rs
```diff
@@ -0,0 +1,286 @@
...
+            )
+        }
+    }
+
+    /// Gets the MIDI device type.
+    pub fn device_type(&self) -> Result<MidiDeviceType> {
+        let device_type = unsafe { ffi::AMidiDevice_getType(self.as_ptr()) };
+        if device_type >= 0 {
+            let device_type =
+                MidiDeviceType::try_from_primitive(device_type as u32).map_err(|e| {
```
MarijnS95, line 101: Other code seems to use `try_from()` instead of `try_from_primitive()`?

ndk/src/midi.rs
```diff
@@ -0,0 +1,286 @@
...
+                buffer.as_mut_ptr(),
+                buffer.len() as ffi::size_t,
+                &mut num_bytes_received,
+                &mut timestamp,
+            )
+        };
+
+        if result < 0 {
+            Err(NdkMediaError::from_status(ffi::media_status_t(result as c_int)).unwrap_err())
+        } else if result == 0 {
```
MarijnS95, line 272: The docs say that this function only ever returns 0 or 1 (and negative for error codes) - should we validate that it never returns anything `> 1` as we do in other functions with a `match` block?

ndk/src/midi.rs
MarijnS95: Fwiw we don't use the ``Creates an `MidiDevice` from a pointer.`` language anywhere else. Since this struct takes/inherits ownership of the pointer (it doesn't _create_ anything, these structs are merely thin wrappers around a pointer), also by closing/releasing these objects on `Drop`, can you use the same ``Assumes ownership of `ptr` `` terminology I used for `NativeWindow`/`SurfaceTexture`?

Also, it's probably fine to make `fn ptr()` public.

ndk/src/midi.rs
MarijnS95: Now that I'm at a computer: I meant to look at `hardware_buffer_format.rs` and how I extracted an `enum` from a file that's otherwise guarded by an `api-level` CFG #276. I think we can do a similar thing here, without spreading `cfg(media)` all across `media/mod.rs`?

ndk/src/midi.rs
```diff
@@ -0,0 +1,294 @@
...
+use num_enum::{IntoPrimitive, TryFromPrimitive};
+use std::fmt;
+use std::marker::PhantomData;
+use std::os::raw::{c_int, c_uint};
+use std::ptr::NonNull;
+
+// There is no mention about thread-safety in the NDK reference, nut the official Android C++ MIDI
+// sample stores `AMidiDevice*` and `AMidi{InputOutput}Port*` in global variables and accesses the
+// ports from separate threads.
+// See https://github.com/android/ndk-samples/blob/main/native-midi/app/src/main/cpp/AppMidiManager.cpp
```
MarijnS95, line 24:
```suggestion
// There is no mention about thread-safety in the NDK reference, but the official Android C++ MIDI
// sample stores `AMidiDevice *` and `AMidi{Input,Output}Port *` in global variables and accesses the
// ports from separate threads.
// See https://github.com/android/ndk-samples/blob/main/native-midi/app/src/main/cpp/AppMidiManager.cpp
```

ndk/src/midi.rs
paxbun: Fixed!

ndk/src/midi.rs
paxbun: Fixed!

ndk/src/midi.rs
paxbun: I have never used MIDI with timestamps - I will run a test and make a commit as you suggested.

ndk/Cargo.toml
MarijnS95: Looks like you're right, I may have ended up accidentally scrolling the Java docs and seeing all the `Available since API 23` notes.

The NDK docs don't mention a bound at all, except for [`AMidiDevice_getDefaultProtocol` which is set at 33](https://developer.android.com/ndk/reference/group/midi#amididevice_getdefaultprotocol), so let's keep these all at `29` :+1:

ndk/src/midi.rs
paxbun: I haven't run the test yet but seems like it's not. 
https://cs.android.com/android/platform/superproject/+/master:frameworks/base/media/native/midi/amidi.cpp;l=344;drc=89b01aeaa9e19377ff547baab791277719b8aaf3?q=AMidi&ss=android%2Fplatform%2Fsuperproject
image

MarijnS95: Neat, let's not expose it to the user then :)

ndk/src/midi.rs
paxbun: That's from the official sample:

https://github.com/android/ndk-samples/blob/7f6936ea044ee29c36b5c3ebd62bb3a64e1e6014/native-midi/app/src/main/cpp/AppMidiManager.cpp#L115-L129

AMidi always returns negative value for errors (`AMEDIA_*` enums are all negative except for `AMIDIA_OK`), I thought there's no possibility that `AMidiOutputPort_receive` is updated to return positive error values in future versions.

MarijnS95: Hmm.

We have a similar case for `fn pre_dispatch()` at #325, but there the docs don't specify whether non-boolean, non-error values are possible and it seems Android on certain devices has been tweaked to return these.

paxbun: If that's the case, should we return `Err(NdkMediaError::from_status(...))` as well as we did in `if result < 0 { ... }`?

ndk/src/midi.rs
paxbun: image
image
Fixed!

ndk/src/midi.rs
paxbun: image
Fixed!

ndk/src/midi.rs
MarijnS95: I was thinking something like:

```rust
match result {
    r if r < 0 => Err(NdkMediaError::from_status(ffi::media_status_t(r as c_int)).unwrap_err()),
    0 => Ok(MidiOpcode::NoMessage),
    1 => Ok(MidiOpcode::Data(num_bytes_received)),
    r => unreachable!("Result is positive integer {}", r),
}
```

Akin to https://github.com/rust-windowing/android-ndk-rs/blob/ddc2757f5abd002916c86576eaffd3f502e334b5/ndk/src/utils.rs#L8-L14

ndk/src/midi.rs
paxbun: Oh, I got it. Maybe I can add `media_errors.rs` and do similar things for this case as well. I'll make a fix for this.

ndk/src/midi.rs
paxbun: Great! Since it seems that the original version `AMidiOutputPort_receive` definitely does not return any value but 0, 1, and `AMEDIA_ERROR_UNKNOWN`, I think this is better. I'll make a commit according to your suggestion.

ndk/src/midi.rs
```diff
@@ -0,0 +1,305 @@
...
+        let mut opcode = 0i32;
+        let mut timestamp = 0i64;
+        let mut num_bytes_received: ffi::size_t = 0;
+        let result = unsafe {
+            ffi::AMidiOutputPort_receive(
+                self.as_ptr(),
+                &mut opcode,
+                buffer.as_mut_ptr(),
+                buffer.len() as ffi::size_t,
+                &mut num_bytes_received,
```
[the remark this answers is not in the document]
paxbun: I want to make the APIs as raw as possible, so we can give more control to those porting their C++ Midi app to Rust. I firstly put Vec<u8> inside the MidiOutputPort and made MidiOpcode::Data contain a &[u8], but I thought that design is not friendly to those who read the Android docs and the official sample and to those who are not used to read other's code, I removed it and modified the code as it is now.

ndk/src/midi.rs
MarijnS95: Probably no need to have _both_ `fn ptr()` _and_ `as_ptr()`?

It looks like we have more of the former, probably stick with that.

ndk/src/midi.rs
paxbun: Fixed!

ndk/src/midi.rs
```diff
@@ -0,0 +1,305 @@
...
+        match num_messages_received {
+            r if r < 0 => {
+                Err(NdkMediaError::from_status(ffi::media_status_t(r as c_int)).unwrap_err())
+            }
+            0 => Ok(MidiOpcode::NoMessage),
+            1 if opcode as c_uint == ffi::AMIDI_OPCODE_DATA => Ok(MidiOpcode::Data {
+                length: num_bytes_received as usize,
+                timestamp,
+            }),
+            1 => Ok(MidiOpcode::Flush),
```
MarijnS95, line 295: Sorry, I forgot to add this to my original suggestion: do you want to check for [`AMIDI_OPCODE_FLUSH`](https://developer.android.com/ndk/reference/group/midi#group___midi_1gga16af7b253440dadd46a80a4b9fddba4dacda48ee7ade411b83e9b731230c2ad45) specifically, and error out if we come across `r == 1` with an unknown opcode?

ndk/src/midi.rs
paxbun: Fixed!

ndk/src/midi.rs
```diff
@@ -0,0 +1,297 @@
...
+use std::os::raw::{c_int, c_uint};
+use std::ptr::NonNull;
+
+// There is no mention about thread-safety in the NDK reference, but the official Android C++ MIDI
+// sample stores `AMidiDevice *` and `AMidi{Input,Output}Port *` in global variables and accesses the
+// ports from separate threads.
+// See https://github.com/android/ndk-samples/blob/7f6936ea044ee29c36b5c3ebd62bb3a64e1e6014/native-midi/app/src/main/cpp/AppMidiManager.cpp
+unsafe impl Send for MidiDevice {}
+unsafe impl<'a> Send for MidiInputPort<'a> {}
+unsafe impl<'a> Send for MidiOutputPort<'a> {}
```
MarijnS95, line 28: Can they be accessed concurrently (i.e., `Sync`), or are these only safe to be used on different threads as long as there aren't two or more threads poking it at the same time (`Send`)?

ndk/src/midi.rs
paxbun: I think we need to strictly check the opcodes as well.
image
Fixed!

ndk/src/midi.rs
```diff
@@ -0,0 +1,300 @@
...
+                Err(NdkMediaError::from_status(ffi::media_status_t(r as c_int)).unwrap_err())
+            }
+            0 => Ok(MidiOpcode::NoMessage),
+            1 => match opcode as c_uint {
+                ffi::AMIDI_OPCODE_DATA => Ok(MidiOpcode::Data {
+                    length: num_bytes_received as usize,
+                    timestamp,
+                }),
+                ffi::AMIDI_OPCODE_FLUSH => Ok(MidiOpcode::Flush),
+                _ => unreachable!("Opcode is {}", opcode),
```
MarijnS95, line 289: Nit: if you can call that `Unrecognized opcode {}` that makes it a bit more clear :)

ndk/src/midi.rs
```diff
@@ -0,0 +1,300 @@
...
+            ptr,
+            _marker: PhantomData,
+        }
+    }
+
+    pub fn ptr(&self) -> NonNull<ffi::AMidiInputPort> {
+        self.ptr
+    }
+
+    /// Sends data to the specified input port.
```
MarijnS95, line 173: Nit (here and below): I know this documentation was inherited from Android, but `specified` is a bit vague here since you can only "specify" `self` :)

ndk/src/midi.rs
MarijnS95: Looks like we're pretty much ready for merge except this point?

ndk/src/midi.rs
paxbun: image
Fixed!

ndk/src/midi.rs
paxbun: aa26247 
Fixed!

ndk/src/midi.rs
paxbun: `MidiIntputPort` is definitely not `Sync`: it puts `write` in the loop, so there's a possibility of interleaved writing.
image

`MidiOutputPort` is definitely `Sync`:`MidiReceiver`, which is called by `AMidiOutputPort_receive` implements synchronization logic. `MidiOutputPort::receive` will immediately return an `Err` if a simultaneous read call was invoked by another thread.
image

`MidiDevice` seems to be `Sync`: as `MidiDeviceServer`, which is called by `MidiDevice`, uses `synchronized` in every method except for `getDeviceInfo()` and `setDeviceInfo()`.
image

ndk/src/midi.rs
paxbun: I'm working on separating `media::*` types that `midi.rs` is dependent on. Maybe we can merge after this.

ndk/src/midi.rs
MarijnS95: That's an interesting design choice and discrepancy between input and output ports. We can also enforce the write with a `&mut self` borrow but I think keeping them as `Send` without `Clone`/`Copy` is also fine.

ndk/src/midi.rs
MarijnS95: Thanks! Feel free to do so in a separate preliminary PR to make it easier to review / keep things separate.

Let's see if we can do it in a nonbreaking kind of way.

paxbun: Fixed! Still have ugly `#![cfg(any(feature = "media", feature = "midi"))]` at the top of `media_errors.rs`, but `crate::media::{NdkMediaError, Result}` is available via `pub use`:
```rust
// In Line 9 of media/mod.rs:
pub use crate::media_error::{MediaErrorResult, NdkMediaError, Result};
```

ndk/src/media/image_reader.rs
```diff
@@ -5,7 +5,8 @@
 #![cfg(feature = "api-level-24")]
 
 use super::NdkMediaError;
-use super::{construct, construct_never_null, error::MediaErrorResult, Result};
+use super::{construct, construct_never_null, Result};
+use crate::media_error::MediaErrorResult;
```
MarijnS95, line 9: These can still be imported from the same crate, right?

ndk/src/media/mod.rs
```diff
@@ -3,20 +3,13 @@
 //! See also [the NDK docs](https://developer.android.com/ndk/reference/group/media)
 #![cfg(feature = "media")]
 
-mod error;
```
MarijnS95, line 6: That might be a breaking change; but we have more breaking PRs queued up for `ndk` anyway.

ndk/src/media/image_reader.rs
paxbun: They can, but If I put `use crate::media_error::MediaErrorResult` to `media/mod.rs` without `features = "api-level-24"`, `MediaErrorResult` is pointed out as an unused member. Should I put `pub use crate_media_error::MediaErrorResult` to `media/mod.rs`?

ndk/src/midi.rs
MarijnS95: Ah, I don't think we strictly need to have those `cfg`s. They're only to guard what to link to in `ndk-sys` (and to limit API levels).

ndk/src/media/image_reader.rs
MarijnS95: In `media/mod.rs` it would have been `pub`.

Alas, since it seems we need to make a breaking change anyway (otherwise need a local `mod error { pub use super:...; }`) let's not reexport (`pub use`) anything.

paxbun: So you mean removing `pub use crate::media_error::{NdkMediaError, Result};` from `media/mod.rs` as well and reflect the changes to `media/{image_reader, media_codec}.rs`?

MarijnS95: Yup.

While at it, let's also update `ndk/CHANGELOG.md` to finalize this.

paxbun: Updated!

ndk/src/midi.rs
paxbun: I re-read the source code of `MidiDevice` and `MidiDeviceServer` - I think we can assume that `MidiDevice` is `Sync` as well.

paxbun: Would the lack of mentions of thread safety in the NDK reference mean that there would be potential changes that make `AMidiDevice` thread-unsafe in the future version?

ndk/src/media/mod.rs
MarijnS95: Never mind, it's not `pub` so this module was never visible.

ndk/src/media/mod.rs
```diff
@@ -3,20 +3,11 @@
...
-use std::{mem::MaybeUninit, ptr::NonNull};
-
-pub type Result<T, E = NdkMediaError> = std::result::Result<T, E>;
-
-fn construct<T>(with_ptr: impl FnOnce(*mut T) -> ffi::media_status_t) -> Result<T> {
-    let mut result = MaybeUninit::uninit();
-    let status = with_ptr(result.as_mut_ptr());
-    NdkMediaError::from_status(status).map(|()| unsafe { result.assume_init() })
-}
+use crate::media_error::{construct, NdkMediaError, Result};
```
MarijnS95, line 9: As discussed previously, in order to not make this a breaking change (which should otherwise be clearly marked in the `README`) `NdkMediaError` and `Result` should be `pub use`. In fact, as a bugfix `MediaErrorResult` should be made publicly accessible too.

I'd like to get that in as a preliminary PR to separate concerns from the `AMidi` feature, and also revise how the errors are structured currently. Mind if I address that first?

ndk/src/midi.rs
MarijnS95: Let's not preemptively add `Send`/`Sync` in that case. If it's not documented there's nothing holding the NDK folks from changing the underling implementation and comments.

We recently [requested clarification](https://github.com/android/ndk/discussions/1768) on other APIs as well, perhaps you can do the same here?

TLDR: Let's keep it `Send` for now until we have official confirmation for `Sync`.

MarijnS95:
> Are there any additional fixes needed? Please let me know if there are any :)

There are a few review comments still unresolved, but nothing major. I just ended up having to deal with some high(er) priority items the past few days, thanks for your patience.

ndk/src/media/mod.rs
paxbun: No, please go ahead. I'll find other potential issues in the current pull request by running some tests. Thanks for explaining the current situation!

ndk/src/midi.rs
paxbun: I found that `AMidiDevice` must be released in the thread where the JNI is attached, i.e., the app crashes when `AMidiDevice_release` is called before `AttachCurrentThread[AsDaemon]` is called from that thread. I'll remove `Send` from `MidiDevice`.

```rust
let midi_device = MidiDevice::from_java(...);
thread::spawn(move || {
  midi_device. ...
  // JNIEnv access error on AMidiDevice_release
});
```

```rust
let midi_device = MidiDevice::from_java(...);
thread::spawn(move || {
  AttachCurrentThread(vm, &env, ...);
  midi_device. ...
  // No error
});
```

paxbun:
I accidently merged another branch to `feat/amidi`, I will make a new branch (`feat/amidi-2`) and a PR from the commit where we've left.

MarijnS95:
@paxbun please don't, just force-push the old commit back.

paxbun:
I didn't know that force-pushing also works in this case... Actually, this is my first time accidentally merging branches already published for a PR. Thanks for your reply!

ndk/src/media/mod.rs
MarijnS95: This separation is now open at https://github.com/rust-mobile/ndk/pull/399

paxbun:
I found some occurrences of `as ffi::size_t` in the current code. Would it be better to change that into `as _`? (#370)

MarijnS95:
Hehe, those should all be gone now :)

Would you mind providing your feedback on #399 and rebase on that so we have consistent error types?

MarijnS95:
> I changed the base branch from `master` to `ndk-rework-media-error`; should I revert it back to `master`?

That gives a nice overview of what this PR entails after #399 has been merged, but will have to be changed back when I inevitably merge that PR and auto-delete the branch (hope we can still do that, but I think we cannot...), so better to leave it at `master` and will re-review this once that PR is in :crossed_fingers:

MarijnS95:
@paxbun can you rebase this to make the changes from #399 trickle through and get rid of the conflicts?

ndk/src/midi.rs
```diff
@@ -0,0 +1,292 @@
...
+    /// Sends a message with a 'MIDI flush command code' to this port.
+    ///
+    /// This should cause a receiver to discard any pending MIDI data it may have accumulated and
+    /// not processed.
+    pub fn send_flush(&self) -> Result<()> {
+        let result = unsafe { ffi::AMidiInputPort_sendFlush(self.ptr.as_ptr()) };
+        MediaError::from_status(result)
+    }
+
+    pub fn send_with_timestamp(&self, buffer: &[u8], timestamp: i64) -> Result<usize> {
```
MarijnS95, line 196:
```suggestion
    /// Sends data to the this port with a timestamp.
    pub fn send_with_timestamp(&self, buffer: &[u8], timestamp: i64) -> Result<usize> {
```

Do we also need extra docs what the timestamp is supposed to mean (and if it is e.g. monotonic with the system, or since the device was opened)? I have zero experience with MIDI so this may be common knowledge :)

ndk/src/midi.rs
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    /// Receives the next pending MIDI message.
+    ///
+    /// To retrieve all pending messages, the client should repeatedly call this method until it
+    /// returns [`Ok(MidiOpcode::NoMessage)`].
+    ///
+    /// Note that this is a non-blocking call. If there are no Midi messages are available, the
+    /// function returns [`Ok(MidiOpcode::NoMessage)`] immediately (for 0 messages received).
+    pub fn receive(&self, buffer: &mut [u8]) -> Result<MidiOpcode> {
```
MarijnS95, line 255: Maybe add an extra paragraph explaining that the `MidiOpcode` contains the `length` which the user must use to bound their slice?

Alternatively I've been toying with `buffer: &mut &mut [u8]`-like APIs, where I update the caller slice to `*buffer = &mut buffer[..num_bytes_received]` so that the user only gets the valid number of bytes back. This becomes more important when you start supporting `[MaybeUninit<u8>]` inputs (which _may_ be useful here!). Alternatively `MidiOpcode` could borrow and return the valid sub-slice (instead of `length`?) so that the user has direct access to these.

Unfortunately all slice helpers around `MaybeUninit` are eternally stuck in "nightly-only experimental" land: https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.slice_assume_init_mut

I'll leave this choice up to you, depending on convenience. The user quite likely already has something along the form of (i.e. it is quite hard to ignore/overlook `MidiOpcode::Data::length`):

```rust
match receive(&mut buffer)? {
    NoMessage => break,
    Data { length, timestamp: _ } => {
        let useful = &buffer[..length as usize];
        // ...
    }
    Flush => ...,
}
```

MarijnS95 (commented):
> Just noticed that `ndk` doesn't have an interface for AMidi, so I made one. Please let me know if there're any tests or fixes I have to do before merging!

Reading this PR description again, do you actually have a use for this API? I wouldn't add these just for the sake of adding them unless there's an actual consumer that can also comment on the ergonomics, efficiency and consistency of these _high level_ wrappers. Though in comments you seem to have done experiments with this, so you were able to put this to use and make sure it is workable?

---

On the note of `MediaError::from_status(...).unwrap_err()`, I feel like this is an API limitation of `MediaError` and might think of something better at some point, since the case of handling any non-negative code as a useful value, any known negative number as an error code, and any unknown error code as `UnknownStatus` is prevalent here and elsewhere.

---

Other than that this looks simple and concise, nice job! Left a few more minor cleanup comments and closed some old, resolved conversations.

paxbun:
Yes, we're using this in our [app](https://play.google.com/store/apps/details?id=com.clebrain.pianokiwis), though our app only uses MIDI input, not output. All my contributions to this repo were derived from that app.

As you can see our app features some audio-based ones, and those are based on `ndk::audio`. I then needed to implement MIDI input, and found that there's no `ndk::midi`, and that's why I said "Just noticed that `ndk` doesn't have an interface for AMidi."

Maybe making an example app for testing this would be great, but I need some time for that.

MarijnS95:
No need to make an example, just pointing out that it sounded like this was added for the sake of adding it rather than needing it, all good if you're using it!

ndk-sys/src/lib.rs
```diff
@@ -51,3 +51,7 @@ extern "C" {}
 #[cfg(all(feature = "audio", target_os = "android"))]
 #[link(name = "aaudio")]
 extern "C" {}
+
+#[cfg(all(feature = "midi", target_os = "android"))]
+#[link(name = "amidi")]
```
MarijnS95, line 56: Noticed that I named this `libmidi.so` in #399 the module-level doc-comment of `media_error.rs`. Can you fix that for me (should be `libamidi.so` apparently) as part of this PR? :pray:

ndk/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - event: Add `tool_type` getter for `Pointer`. (#323)
+- Add `AMidi` bindings. (#353)
 - input_queue: Allow any non-zero return code from `pre_dispatch()` again, as per documentation. (#325)
 - asset: Use entire asset length when mapping buffer. (#387)
```
MarijnS95, line 6:
```suggestion
- event: Add `tool_type` getter for `Pointer`. (#323)
- input_queue: Allow any non-zero return code from `pre_dispatch()` again, as per documentation. (#325)
- Add `AMidi` bindings. (#353)
- asset: Use entire asset length when mapping buffer. (#387)
```

If we want to keep these sorted by PR number :)

paxbun:
> I was thinking of something along the lines of:
>
> https://github.com/MarijnS95/ndk/compare/negative-error-code-helper
>
> WDYT?

LGTM! There are 6 branches checking whether the return value is negative; midi.rs will benefit from `from_status_if_negative`.

ndk/src/midi.rs
paxbun: It's the return value of `long java.lang.System.nanoTime()`, but the NDK documents it as a `CLOCK_MONOTONIC` time... I initially tried to use `Instant`, but I'm not sure if I can assume that the current implementation of `Instant` on Unix systems will keep using `CLOCK_MONOTONIC`.

ndk/Cargo.toml
```diff
@@ -18,6 +18,7 @@ all = ["audio", "bitmap","media", "api-level-30"]
 audio = ["ffi/audio", "api-level-26"]
 bitmap = ["ffi/bitmap"]
 media = ["ffi/media"]
+midi = ["ffi/midi", "api-level-29", "media"]
```
MarijnS95, line 21: This wasn't dependent on `libmediandk.so` before; why is it now?

ndk/src/midi.rs
MarijnS95: Looks like you resolved this. No clue about removing `impl Send` from the ports, the NDK makes no guarantees but it seems logical that you might want to process such data on another thread even if the JNI device can only be accessed on a VM-attached thread?

ndk/src/midi.rs
MarijnS95: Yeah don't cast into `Instant` unless it has a public API detailing how to manually create such a type. Please document those findings in a doc comment.

Sounds like `java.lang.System.nanoTime()` returns `CLOCK_MONOTONIC` (on Unix systems)? If there's a source I'd love to reference it in the doc-comment.

MarijnS95 (approved):
Looking really good, almost there! Collecting all the open conversations in one place:

- [ ] Sudden dependency on `libmediandk.so`?
- [x] Missing docs on `send_with_timestamp()`;
- [ ] `receive()` should:
  - [ ] document that the user should limit the `buffer` to `length` after the function returns;
  - [ ] perhaps accept a `MaybeUninit` slice?
  - [ ] Optionally, if you like it: have a different API that returns the `lengh`-subsliced `buffer` straight away;
- [ ] Remove `impl Send` from `Midi*Port`?

I will tackle the `.unwrap_err()` generalization separately, doesn't matter if this PR lands first.

ndk/src/midi.rs
MarijnS95: Another question, just for my understanding: is this timestamp detailing when this message was put together (same for the `timestamp` field in `receive()`), or should it be some time in the future when the receiver should process/play back the given message/note?

ndk/Cargo.toml
paxbun: This is because media_error.rs needs the "media" feature, would adding a separate "media_error" feature and making both "media" and "midi" be dependent on "media_error" okay?

MarijnS95: If I remember correctly I designed this to be extended (since it's not really a public feature):

https://github.com/rust-mobile/ndk/blob/e66b4fd71bc0fa1ae895a98506c51eb7cf588b68/ndk/src/media_error.rs#L5

paxbun:
Sorry for the late reply; I'm working on another project. I'll try to make the work done by this weekend. Thanks for your patience!

ndk/Cargo.toml
paxbun: Building `ndk` only with "midi" and "api-level-29" makes error because `ndk::midi` depends on `ndk::media_error`. `ndk::media` is guarded with `#![cfg(feature = "media")]` while `ndk::midi` is not...

MarijnS95:
@paxbun all good, though you might miss out on the next breaking release window.

That doesn't hurt as we can always do a non-breaking patch release for these purely-additive APIs, and spend that _tiny bit_ more time fleshing out the functions so that we don't have to make breaking changes in the near future.

paxbun:
@MarijnS95 I think keeping `Midi*Port` `Send` is okay; they're just a holder of a file descriptor and some values representing the state of the port.
image
The Java counterpart of `AMidi*Port` implements the MIDI handling logic in pure Java, since that logic only requires file IO functions. Likewise, I think assuming that `AMidi*Port` won't depend on JNI in the future is okay, as the NDK implementation will also contain IO logic only.

Moreover, I think making `Midi*Port` `Sync` is also okay; As I said above, `AMidiOutputPort` is guarded by an atomic integer (it's like `Mutex::try_lock`), like Java's `MidiOutputPort`'s implementations are also guarded with `synchronized` blocks. I thought `AMidiInputPort_send` was not thread-safe, but I just found this [comments](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/media/native/midi/amidi.cpp;l=68) in the Android source code:
image
Since the MIDI socket is opened with `SOCK_SEQPACKET`, messages won't be interleaved even when `AMidiInputPort_send` is called from multiple threads.

Making `MidiOutputPort` `Send` is necessary because in the current NDK API `AMidiOutputPort_receive` is designed to be called in a separate thread; Though `AMidiInputPort_send` does not need a dedicated thread, I think some users would need to call it in another thread.

In addition, since `Midi*Port<'a>` has a lifetime parameter that depends on its associated `MidiDevice` and `MidiDevice` is not `Send` nor `Sync`, so most users who send `Midi*Port<'a>` to another thread will attach JNI to that thread.

If you think it's okay to keep `Midi*Port` `Send` (and `Sync` as well), I'll add documentation for this.

paxbun:
@MarijnS95 You're right, that makes the lifetime parameter meaningless, I overlooked it.
But allowing MidiDevice to be used with multithreading is important for its practical use. How about adding `SafeMidiDevice`, `SafeMidiInputPort`, and `SafeMidiOutputPort` that ensure the safety conditions (e.g., JavaVM thread attachment) and hold their unsafe counterpart?

MarijnS95:
> @MarijnS95 You're right, that makes the lifetime parameter meaningless, I overlooked it.
> But allowing MidiDevice to be used with multithreading is important for its practical use. How about adding `SafeMidiDevice`, `SafeMidiInputPort`, and `SafeMidiOutputPort` that ensure the safety conditions (e.g., JavaVM thread attachment) and hold their unsafe counterpart?

I had concerns about the lifetime still making it hard to work around this as long as the `MidiDevice` is either refcounted and clonable, or wrapped in an `Arc` (same effect) and it seems you've gone for the latter.

Not sure if we really need a "safe" counterpart or can just make this the default implementation?

Anyways, I'll involve @rib since he did a lot on the JNI crate and is probably better at evaluating whether this is the right - and safe - way to go.

ndk/src/midi/safe.rs
```diff
@@ -0,0 +1,241 @@
...
+use crate::media_error::Result;
+
+use super::*;
+
+/// The owner of [`MidiDevice`]. [`SafeMidiDevice`], [`SafeMidiInputPort`], and
+/// [`SafeMidiOutputPort`] holds an [`Arc<SafeMidiDeviceBox>`]. to ensure that the underlying
+/// [`MidiDevice`] is not dropped while the safe wrappers are alive.
+///
+/// [`SafeMidiDeviceBox`] also holds a pointer to the current Java VM to attach the calling thread
+/// of [`Drop::drop`] to the VM, which is required by [`ffi::AMidiDevice_release`].
```
MarijnS95, line 27: So it is only the `drop()` that requires to be attached to the Java VM, all other calls are safe to call elsewhere since this implements `Deref` and all `MidiDevice` functions take an immutable borrow.

You said something similar on https://github.com/rust-mobile/ndk/pull/353#discussion_r987970649, perhaps good to mention that here (as doc-comment on the `Deref`) that it is send/sync except drop?

Otherwise let's continue discussing whether we can make this the default implementation, rather than having a threadsafe and not-threadsafe variant.

MarijnS95: Also don't forget to have `()` on all intradoc-links to functions :wink:

paxbun: For my use, it's 100% okay to make the "safe" one default implementation; I'm just concerned about whether there's some demand for a low-level but easy-to-use wrapper for AMidi.

MarijnS95: > For my use, it's 100% okay to make the "safe" one default implementation; I'm just concerned about whether there's some demand for a low-level but easy-to-use wrapper for AMidi.

In my experience these are intended to be high level wrappers, and I wouldn't worry about sneaking an `Arc` in there (as long as it is clearly annotated at the places that clone/own it, so that the caller is aware that they need to drop a bunch of things) in order to make various operations lifetime-safe, which would otherwise be impossible with Rust lifetimes and borrows. Especially if this API is intended to be used in multithreaded fashion.

Then later someone may wants to have weak references instead so that they can drop their `MidiDevice` in place and make the ports dysfunct :grimacing: 

But let's again wait for @rib, I have a hunch he has a better feeling for such APIs and use-cases.

> I mentioned that MidiDevice is Send and Sync except for Drop in the module-level doc-comment; would it be better to move it to SafeMidiDeviceBox's doc-comment?

Didn't see that, only that it holds the Java VM to implement drop but missing the `, all other functions are safe to call from any thread`.

paxbun: > Didn't see that, only that it holds the Java VM to implement drop but missing the , all other functions are safe to call from any thread.

You're right. Just added that to the module-level doc-comment of `ndk::midi::safe`.

paxbun: > Also don't forget to have () on all intradoc-links to functions 😉

Fixed!

ndk/src/midi/safe.rs
```diff
@@ -0,0 +1,244 @@
+//! Safe bindings for [`AMidiDevice`], [`AMidiInputPort`], and [`AMidiOutputPort`]
+//!
+//! Provides implementation of `SafeMidiDeviceBox` that ensures the current thread is attached to
+//! the Java VM when being dropped, which is required by [`AMidiDevice_release`]. All other types
```
MarijnS95, line 4:
```suggestion
//! the Java VM when being dropped, which is required by [`ffi::AMidiDevice_release()`]. All other types
```

ndk/src/midi/safe.rs
```diff
@@ -0,0 +1,244 @@
...
+use crate::media_error::Result;
+
+use super::*;
+
+/// The owner of [`MidiDevice`]. [`SafeMidiDevice`], [`SafeMidiInputPort`], and
+/// [`SafeMidiOutputPort`] holds an [`Arc<SafeMidiDeviceBox>`]. to ensure that the underlying
+/// [`MidiDevice`] is not dropped while the safe wrappers are alive.
+///
+/// [`SafeMidiDeviceBox`] also holds a pointer to the current Java VM to attach the calling thread
+/// of [`Drop::drop`] to the VM, which is required by [`ffi::AMidiDevice_release()`].
```
MarijnS95, line 30:
```suggestion
/// of [`Drop::drop()`] to the VM, which is required by [`ffi::AMidiDevice_release()`].
```

ndk/src/midi/safe.rs
```diff
@@ -0,0 +1,244 @@
...
+struct SafeMidiDeviceBox {
+    midi_device: ManuallyDrop<MidiDevice>,
+    java_vm: NonNull<jni_sys::JavaVM>,
+}
+
+// SAFETY: [`SafeMidiDeviceBox::drop`] attaches the calling thread to the Java VM if required.
```
MarijnS95, line 36:
```suggestion
// SAFETY: [`SafeMidiDeviceBox::drop()`] attaches the calling thread to the Java VM if required.
```

ndk/src/midi/safe.rs
```diff
@@ -0,0 +1,244 @@
...
+unsafe impl Send for SafeMidiDeviceBox {}
+
+// SAFETY: [`SafeMidiDeviceBox::drop`] attaches the calling thread to the Java VM if required.
```
MarijnS95, line 39:
```suggestion
// SAFETY: [`SafeMidiDeviceBox::drop()`] attaches the calling thread to the Java VM if required.
```

ndk/src/midi/safe.rs
```diff
@@ -0,0 +1,244 @@
...
+    }
+}
+
+// SAFETY: a AMidi port is a mere holder of an atomic state, a pointer to the associated MIDI
+// device, a binder, and a file descriptor, all of which are safe to be sent to another thread.
+// https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/media/native/midi/amidi.cpp?q=symbol%3A%5CbAMIDI_Port%5Cb%20case%3Ayes
+unsafe impl Send for SafeMidiOutputPort {}
+
+// SAFETY: AMidiOutputPort is guarded by a atomic state ([`AMIDI_Port::state`]), which enables
+// [`ffi::AMidiOutputPort_receive`] to detect accesses from multiple threads and return error.
```
MarijnS95, line 242:
```suggestion
// [`ffi::AMidiOutputPort_receive()`] to detect accesses from multiple threads and return error.
```

MarijnS95 (commented):
Some more `()` nits, though not sure if they're relevant at this stage.

ndk/Cargo.toml
paxbun: Yes that works, I just didn't come up with that... 😢

PR #354, opened 2022-10-10
readme: Document compatibility with NDK r25 and remove workaround mention

MarijnS95:
NDK r25 works just as well as r24, and is the current LTS.  There are no other releases planned between it and r26 launching in 2023, and scheduled to be the next NDK.

The mention of our workaround is irrelevant given that it's something we deal with for the user in the background, and the feature has evolved many times through multiple PRs.  This allows us to drop the beta notion for r23 which is long deprecated by now.


Jasper-Bekkers (approved):

msiglreith (approved):
I'm fine with removing the workaround note

Issue #355, opened 2022-10-11
Reducing the size of build (stripping)?


Animeshz:
The hello_world example produces shared library of 19MB (maybe reasonable for the std lib), and adding simple library of axum increases it to 42MB.

Doesn't seem good, seems like a lot of unused symbols are packed into the library, is there way to strip it before packaging?

MarijnS95:
@Animeshz This is the default for Rust debug libraries/binaries. You can use [standard `Cargo.toml` configuration](https://doc.rust-lang.org/cargo/reference/profiles.html#strip) to control this behaviour.

PR #356, opened 2022-10-17
cargo-apk: Add `strip` configuration to manifest

Jake-Shadle:
I've been extremely annoyed with the time that `cargo-apk` adds onto our build for a long time and today was the day that I got annoyed enough to fix it.

This PR adds a configuration field, `package.metadata.android.strip` that allows users to configure how they want to treat debug symbols _after_ cargo build has finished, but _before_ the .so libraries are appended to the APK.

- `default` - This acts the exact same as the current version of cargo-apk, and is the value used if `strip` is not explicitly set. The .so produced by cargo is simply copied to the `apk` output directory with no modification.
- `strip` - The library copied into the apk output directory is stripped of debug symbols
- `split` - The library copied into the apk output directory is stripped of debug symbols, in addition, the debug symbols are copied into the directory as well with the same filename as the library, but a `.dwarf` extension.

#355 was closed with the suggestion to just use the [strip](https://doc.rust-lang.org/cargo/reference/profiles.html#strip) option in cargo, but unfortunately that doesn't work for our use case. We absolutely do want debug symbols, since we upload all debug symbols to our symbol server so that we can either use them for debugging, or most likely, as the source for the debugging information our error/crash reporting system uses to resolve symbolicated stack traces. Using `strip` to remove debug info is therefore not an option for us, and unfortunately [`split-debuginfo`](https://doc.rust-lang.org/rustc/codegen-options/index.html#split-debuginfo) doesn't actually do anything when targeting android, at least at the moment. That being said I think this option still makes sense once `split-debuginfo` does work for android, since it keeps the split symbols next to the library they are paired with, which makes it nicer for any tooling that is doing any kind of processing to handle them, as otherwise they would need to look in the regular target dir as well as the apk dir.

This change make our Android release step, which currently produces ~2.4GiB of DWARF debug data, go from a ridiculous +3m30s (surpassing the time it actually takes the code to build!) to a more reasonable (though still not good) +17s to strip debug symbols and write the stripped binaries and other things to the APK.

### DIFF
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -60,7 +60,7 @@ apk_name = "myapp"
 # debug symbols are present in the `.so` file(s) produced by your build, enabling
 # https://doc.rust-lang.org/cargo/reference/profiles.html#strip or
 # https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo
-# in your cargo manifest will cause debug symbols will no longer be present
+# in your cargo manifest can cause debug symbols to no longer be present
 # in the `.so`.
 strip = "default"
 
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -45,12 +45,23 @@ assets = "path/to/assets_folder"
 # Defaults to package name.
 apk_name = "myapp"
 
-# `default` (or unspecified) - Debug symbols are not treated specially
-# `strip` - Debug symbols are stripped from the shared libraries before being
-#   copied into the APK
-# `split` - Functions the same as strip, except the debug symbols are written to
-#   the apk output directory alongside the stripped shared libraries, but with a
-#   `.dwarf` extension.
+# `default` (or unspecified) - Debug symbols, if they exist, are not treated
+#                              specially
+#
+# `strip`                    - Debug symbols are stripped from the shared
+#                              libraries before being copied into the APK
+#
+# `split`                    - Functions the same as strip, except the debug
+#                              symbols are written to the apk output directory
+#                              alongside the stripped shared libraries, but with
+#                              a `.dwarf` extension.
+#
+# Note that the `strip` and `split` options will only have an effect if the
+# debug symbols are present in the `.so` file(s) produced by your build, using
+# https://doc.rust-lang.org/cargo/reference/profiles.html#strip or
+# https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo
+# in your cargo manifest will mean the debug symbols will no longer be present
+# in the `.so`
 strip = "default"
 
 # Folder containing extra shared libraries intended to be dynamically loaded at runtime.
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -56,7 +56,7 @@ apk_name = "myapp"
 #                              alongside the stripped shared libraries, with
 #                              a `.dwarf` extension.
 #
-# Note that the `strip` and `split` options will only have an effect if the
+# Note that the `strip` and `split` options will only have an effect if
 # debug symbols are present in the `.so` file(s) produced by your build, enabling
 # https://doc.rust-lang.org/cargo/reference/profiles.html#strip or
 # https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -45,6 +45,14 @@ assets = "path/to/assets_folder"
 # Defaults to package name.
 apk_name = "myapp"
 
+# `default` (or unspecified) - Debug symbols are not treated specially
+# `strip` - Debug symbols are stripped from the shared libraries before being
+#   copied into the APK
+# `split` - Functions the same as strip, except the debug symbols are written to
+#   the apk output directory alongside the stripped shared libraries, but with a
+#   `.dwarf` extension.
+strip = "default"
+
 # Folder containing extra shared libraries intended to be dynamically loaded at runtime.
 # Files matching `libs_folder/${android_abi}/*.so` are added to the apk
 # according to the specified build_targets.
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -53,7 +53,7 @@ apk_name = "myapp"
 #
 # `split`                    - Functions the same as strip, except the debug
 #                              symbols are written to the apk output directory
-#                              alongside the stripped shared libraries, but with
+#                              alongside the stripped shared libraries, with
 #                              a `.dwarf` extension.
 #
 # Note that the `strip` and `split` options will only have an effect if the
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -51,7 +51,7 @@ apk_name = "myapp"
 # `strip`                    - Debug symbols are stripped from the shared
 #                              libraries before being copied into the APK.
 #
-# `split`                    - Functions the same as strip, except the debug
+# `split`                    - Functions the same as `strip`, except the debug
 #                              symbols are written to the apk output directory
 #                              alongside the stripped shared libraries, with
 #                              a `.dwarf` extension.
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -57,7 +57,7 @@ apk_name = "myapp"
 #                              a `.dwarf` extension.
 #
 # Note that the `strip` and `split` options will only have an effect if the
-# debug symbols are present in the `.so` file(s) produced by your build, using
+# debug symbols are present in the `.so` file(s) produced by your build, enabling
 # https://doc.rust-lang.org/cargo/reference/profiles.html#strip or
 # https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo
 # in your cargo manifest will cause debug symbols will no longer be present
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -60,7 +60,7 @@ apk_name = "myapp"
 # debug symbols are present in the `.so` file(s) produced by your build, using
 # https://doc.rust-lang.org/cargo/reference/profiles.html#strip or
 # https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo
-# in your cargo manifest will mean the debug symbols will no longer be present
+# in your cargo manifest will cause debug symbols will no longer be present
 # in the `.so`.
 strip = "default"
 
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -61,7 +61,7 @@ apk_name = "myapp"
 # https://doc.rust-lang.org/cargo/reference/profiles.html#strip or
 # https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo
 # in your cargo manifest will mean the debug symbols will no longer be present
-# in the `.so`
+# in the `.so`.
 strip = "default"
 
 # Folder containing extra shared libraries intended to be dynamically loaded at runtime.
--- a/cargo-apk/README.md
+++ b/cargo-apk/README.md
@@ -46,10 +46,10 @@ assets = "path/to/assets_folder"
 apk_name = "myapp"
 
 # `default` (or unspecified) - Debug symbols, if they exist, are not treated
-#                              specially
+#                              specially.
 #
 # `strip`                    - Debug symbols are stripped from the shared
-#                              libraries before being copied into the APK
+#                              libraries before being copied into the APK.
 #
 # `split`                    - Functions the same as strip, except the debug
 #                              symbols are written to the apk output directory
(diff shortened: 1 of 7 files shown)


MarijnS95:
> But the `$HOME` is a lie

This is because the value defaults to `$HOME` (`dirs::home_dir()`):

https://github.com/rust-windowing/android-ndk-rs/blob/3ed8ecbb87cbe57715e3d0cedbf84b6c6682dfcf/ndk-build/src/ndk.rs#L55

And that was simply repeated in the example.

> There are also 2 other changes I made when doing this change since our build broke when testing.

Please don't make drive-by changes: that's terrible for review, bisecting, proper commit/PR titling (:thinking:) and changelog keeping (which is surprisingly also missing here).

MarijnS95:
As for the rest of the change (once separated out into logical PRs), can anyone else pick up review? I currently don't have any bandwidth remaining for `android-ndk-rs` besides figuring the latest `clap` upgrade breakage.

msiglreith (commented):
looks fine to me, an entry in our [example manifest](https://github.com/rust-windowing/android-ndk-rs/blob/master/cargo-apk/README.md) is missing

ndk-build/src/apk.rs
```diff
@@ -96,7 +120,56 @@ impl<'a> UnalignedApk<'a> {
...
+                        return Err(NdkError::CmdFailed(cmd));
+                    }
+                });
+
+                if self.config.strip == StripConfig::Split {
+                    let dwarf_path = {
+                        let mut dp = lib_path.clone();
+                        dp.set_extension("dwarf");
+                        dp
+                    };
```
MarijnS95, line 148: Doesn't [`.with_extension()`](https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.with_extension) exist for exactly this purpose?

ndk-build/src/apk.rs
```diff
@@ -96,7 +120,56 @@ impl<'a> UnalignedApk<'a> {
...
+            }
+            StripConfig::Strip | StripConfig::Split => {
+                let obj_copy = self.config.ndk.toolchain_bin("objcopy", target)?;
+
+                timeit!(format!("strip debuginfo from {}", lib_path.display()), {
+                    let mut cmd = Command::new(&obj_copy);
+                    cmd.current_dir(&self.config.build_dir);
+                    cmd.arg("--strip-debug");
+                    cmd.arg(path);
+                    cmd.arg(&lib_path);
```
MarijnS95, line 136: Does this need to run inside the build dir, can't it simply use the absolute `out` path instead, similar to what's passed to `std::fs::copy()` above?

Same question below.

MarijnS95 (changes requested):

ndk-build/src/apk.rs
```diff
@@ -96,7 +120,56 @@ impl<'a> UnalignedApk<'a> {
...
+                        cmd.arg(&dwarf_path);
+
+                        if !cmd.status()?.success() {
+                            return Err(NdkError::CmdFailed(cmd));
+                        }
+                    });
+
+                    let mut cmd = Command::new(obj_copy);
+                    cmd.current_dir(&self.config.build_dir);
+                    cmd.arg(format!("--add-gnu-debuglink={}", dwarf_path.display()));
```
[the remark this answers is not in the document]
Jake-Shadle: No clue, I can take it out if you want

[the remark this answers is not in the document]
MarijnS95: I'd just like to know if (still) works :)

ndk-build/CHANGELOG.md
```diff
@@ -32,7 +34,7 @@
 
 - **Breaking:** Default `target_sdk_version` to `30` or lower (instead of the highest supported SDK version by the detected NDK toolchain)
   for more consistent interaction with Android backwards compatibility handling and its increasingly strict usage rules:
-  https://developer.android.com/distribute/best-practices/develop/target-sdk
+  <https://developer.android.com/distribute/best-practices/develop/target-sdk>
```
MarijnS95, line 37: And this?

cargo-apk/README.md
```diff
@@ -45,6 +45,14 @@ assets = "path/to/assets_folder"
 # Defaults to package name.
 apk_name = "myapp"
 
+# `default` (or unspecified) - Debug symbols are not treated specially
+# `strip` - Debug symbols are stripped from the shared libraries before being
+#   copied into the APK
+# `split` - Functions the same as strip, except the debug symbols are written to
+#   the apk output directory alongside the stripped shared libraries, but with a
+#   `.dwarf` extension.
+strip = "default"
```
MarijnS95, line 54: I don't know what this looks like on your machine, but what's wrong with:

```suggestion
# `default` (or unspecified) - Debug symbols are not treated specially
# `strip` - Debug symbols are stripped from the shared libraries before being
#           copied into the APK
# `split` - Functions the same as strip, except the debug symbols are written to
#           the apk output directory alongside the stripped shared libraries, but
#           with a `.dwarf` extension.
strip = "default"
```

One-upping this, can we do something to align it with the `-` for `` `default` (or unspecified) - ``?

ndk-build/src/apk.rs
```diff
@@ -9,6 +9,24 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, serde::Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum StripConfig {
+    /// Does not treat debug symbols especially
```
MarijnS95, line 15: `especially` is an adverb...

MarijnS95: Maybe more clear to explain (here and for the other options, also in the README) that Rust's stripping may also have an affect (in the sense that `Split` won't have anything to work with anymore).

MarijnS95 (approved):
Looks good besides some wording issues in doc comments and indentation issues in the `README`. Feel free to document the effect/affect of Rust's `strip` option on this cargo-apk strip config.

ndk-build/src/apk.rs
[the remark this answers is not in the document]
Jake-Shadle: Presumably that's what debuglink is supposed to do, but I have no idea if adb respects it. I've literally never debugged on Android, if you're asking me to manually verify that debugging works if the debug symbols are split like this I guess I'll just make this PR a draft and get around to it eventually.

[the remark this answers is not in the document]
MarijnS95: Oh this shouldn't really be a blocker as `split` isn't the default, I was just interested. Maybe we need an extra line in the emitted `Android.mk`, for example.

It's not `adb` but a wrapper script in the NDK that implements this FWIW - but it's quite error-prone already.

msiglreith (approved):

ndk-build/src/apk.rs
```diff
@@ -9,6 +9,31 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
+/// The options for how to treat debug symbols that are present in any `.so`
+/// files that are added to the APK.
+///
+/// Using [strip](https://doc.rust-lang.org/cargo/reference/profiles.html#strip)
+/// or [split-debuginfo](https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo)
+/// in your cargo manifest(s) may cause debug symbols to not be present in an
+/// `.so`, which would cause these options to do nothing
```
MarijnS95, line 18:
```suggestion
/// in your cargo manifest(s) may cause debug symbols to not be present in a
/// `.so`, which would cause these options to do nothing.
```

cargo-apk/README.md
```diff
@@ -45,6 +45,25 @@ assets = "path/to/assets_folder"
 # Defaults to package name.
 apk_name = "myapp"
 
+# `default` (or unspecified) - Debug symbols, if they exist, are not treated
+#                              specially
+#
+# `strip`                    - Debug symbols are stripped from the shared
+#                              libraries before being copied into the APK
```
MarijnS95, line 52: Matching the sentence formatting below:

```suggestion
#                              specially.
#
# `strip`                    - Debug symbols are stripped from the shared
#                              libraries before being copied into the APK.
```

cargo-apk/README.md
```diff
@@ -45,6 +45,25 @@ assets = "path/to/assets_folder"
...
+#
+# `split`                    - Functions the same as strip, except the debug
+#                              symbols are written to the apk output directory
+#                              alongside the stripped shared libraries, but with
```
MarijnS95, line 56:
```suggestion
#                              alongside the stripped shared libraries, with
```

cargo-apk/README.md
```diff
@@ -45,6 +45,25 @@ assets = "path/to/assets_folder"
 # Defaults to package name.
 apk_name = "myapp"
 
+# `default` (or unspecified) - Debug symbols, if they exist, are not treated
+#                              specially
+#
+# `strip`                    - Debug symbols are stripped from the shared
+#                              libraries before being copied into the APK
+#
+# `split`                    - Functions the same as strip, except the debug
```
MarijnS95, line 54:
```suggestion
# `split`                    - Functions the same as `strip`, except the debug
```

cargo-apk/README.md
```diff
@@ -45,6 +45,25 @@ assets = "path/to/assets_folder"
...
+#                              symbols are written to the apk output directory
+#                              alongside the stripped shared libraries, but with
+#                              a `.dwarf` extension.
+#
+# Note that the `strip` and `split` options will only have an effect if the
```
MarijnS95, line 59:
```suggestion
# Note that the `strip` and `split` options will only have an effect if
```

cargo-apk/README.md
```diff
@@ -45,6 +45,25 @@ assets = "path/to/assets_folder"
...
+# debug symbols are present in the `.so` file(s) produced by your build, using
```
MarijnS95, line 60:
```suggestion
# debug symbols are present in the `.so` file(s) produced by your build, enabling
```

cargo-apk/README.md
```diff
@@ -45,6 +45,25 @@ assets = "path/to/assets_folder"
...
+# https://doc.rust-lang.org/cargo/reference/profiles.html#strip or
+# https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo
+# in your cargo manifest will mean the debug symbols will no longer be present
+# in the `.so`
```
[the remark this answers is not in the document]
MarijnS95: Nor `.dwarf` if you opt for `split`.

MarijnS95 (approved):
Just bike-shedding the wording now :)

ndk-build/src/cargo.rs
```diff
@@ -47,7 +47,7 @@ pub fn cargo_ndk(
             }
         }
         Err(std::env::VarError::NotUnicode(_)) => {
-            panic!("CARGO_ENCODED_RUSTFLAGS environment variable contains non-unicode characters")
+            panic!("RUSTFLAGS environment variable contains non-unicode characters")
```
MarijnS95, line 50: Whoops, leave this at `CARGO_ENCODED_RUSTFLAGS`, `RUSTFLAGS` is read above and has its own panic.

ndk-build/CHANGELOG.md
```diff
@@ -36,7 +37,7 @@
 
 - **Breaking:** Default `target_sdk_version` to `30` or lower (instead of the highest supported SDK version by the detected NDK toolchain)
   for more consistent interaction with Android backwards compatibility handling and its increasingly strict usage rules:
-  <https://developer.android.com/distribute/best-practices/develop/target-sdk>
+  https://developer.android.com/distribute/best-practices/develop/target-sdk
```
MarijnS95, line 40: We've already committed to these now, fine to leave them.

(As mentioned elsewhere in this / the other PR, vscode stops highlighting them in the code view without these, so it seems editors are expecting these markers too)

ndk-build/src/apk.rs
```diff
@@ -9,6 +9,31 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
+/// The options for how to treat debug symbols that are present in any `.so`
+/// files that are added to the APK.
+///
+/// Using [strip](https://doc.rust-lang.org/cargo/reference/profiles.html#strip)
+/// or [split-debuginfo](https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo)
```
MarijnS95, line 16:
```suggestion
/// Using [`strip`](https://doc.rust-lang.org/cargo/reference/profiles.html#strip)
/// or [`split-debuginfo`](https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo)
```

cargo-apk/README.md
```diff
@@ -45,6 +45,25 @@ assets = "path/to/assets_folder"
...
+# `split`                    - Functions the same as `strip`, except the debug
+#                              symbols are written to the apk output directory
+#                              alongside the stripped shared libraries, with
+#                              a `.dwarf` extension.
+#
+# Note that the `strip` and `split` options will only have an effect if
+# debug symbols are present in the `.so` file(s) produced by your build, enabling
+# https://doc.rust-lang.org/cargo/reference/profiles.html#strip or
+# https://doc.rust-lang.org/cargo/reference/profiles.html#split-debuginfo
+# in your cargo manifest will cause debug symbols will no longer be present
```
MarijnS95, line 63: will cause .. will?

MarijnS95 (approved):
Two more nits, and feel free to add back the `<>` around markdown links, then we're ready to merge!

PR #357, opened 2022-10-17
nkd-build: Propagate `RUSTFLAGS` into `CARGO_ENCODED_RUSTFLAGS`

Jake-Shadle:
When running the actual cargo build, the `CARGO_ENCODED_RUSTFLAGS` was being read so that the rustflags set by the user are not overwritten when cargo-apk is adding its flags. However, this environment variable will only be set if you are building the APK via a build script, if you are invoking `cargo apk build` cargo will not create that environment variable since it doesn't know what actual target/profile is going to be built. If `CARGO_ENCODED_RUSTFLAGS` is not found, this change now falls back to `RUSTFLAGS`, and interprets it the same as `cargo` does before writing it in the encoded form expected by `CARGO_ENCODED_RUSTFLAGS`

### DIFF
--- a/ndk-build/src/cargo.rs
+++ b/ndk-build/src/cargo.rs
@@ -14,12 +14,43 @@ pub fn cargo_ndk(
     let clang_target = format!("--target={}{}", target.ndk_llvm_triple(), sdk_version);
     let mut cargo = Command::new("cargo");
 
+    const SEP: char = '\x1f';
+
     // Read initial RUSTFLAGS
     let mut rustflags = match std::env::var("CARGO_ENCODED_RUSTFLAGS") {
         Ok(val) => val,
-        Err(std::env::VarError::NotPresent) => "".to_string(),
+        Err(std::env::VarError::NotPresent) => {
+            match std::env::var("RUSTFLAGS") {
+                Ok(val) => {
+                    // Note this is roughly equivalent to what cargo does
+                    // https://github.com/rust-lang/cargo/blob/f6de921a5d807746e972d9d10a4d8e1ca21e1b1f/src/cargo/core/compiler/build_context/target_info.rs#L682-L690
+                    let mut rf = String::with_capacity(val.len());
+                    if !val.is_empty() {
+                        for arg in val.split(' ').filter_map(|s| {
+                            let s = s.trim();
+                            if !s.is_empty() {
+                                Some(s)
+                            } else {
+                                None
+                            }
+                        }) {
+                            rf.push_str(arg);
+                            rf.push(SEP);
+                        }
+
+                        rf.pop();
+                    }
+
+                    rf
+                }
+                Err(std::env::VarError::NotPresent) => String::new(),
+                Err(std::env::VarError::NotUnicode(_)) => {
+                    panic!("RUSTFLAGS environment variable contains non-unicode characters")
+                }
+            }
+        }
         Err(std::env::VarError::NotUnicode(_)) => {
-            panic!("RUSTFLAGS environment variable contains non-unicode characters")
+            panic!("CARGO_ENCODED_RUSTFLAGS environment variable contains non-unicode characters")
         }
     };
 
@@ -36,7 +67,7 @@ pub fn cargo_ndk(
     // https://doc.rust-lang.org/beta/cargo/reference/environment-variables.html#configuration-environment-variables
     cargo.env(cargo_env_target_cfg("LINKER", triple), &clang);
     if !rustflags.is_empty() {
-        rustflags.push('\x1f');
+        rustflags.push(SEP);
     }
     rustflags.push_str("-Clink-arg=");
     rustflags.push_str(&clang_target);
@@ -67,7 +98,9 @@ pub fn cargo_ndk(
         // suggests to resort to RUSTFLAGS.
         // Note that `rustflags` will never be empty because of an unconditional `.push_str` above,
         // so we can safely start with appending \x1f here.
-        rustflags.push_str("\x1f-L\x1f");
+        rustflags.push(SEP);
+        rustflags.push_str("-L");
+        rustflags.push(SEP);
         rustflags.push_str(
             cargo_apk_link_dir
                 .to_str()
--- a/ndk-build/src/cargo.rs
+++ b/ndk-build/src/cargo.rs
@@ -16,7 +16,7 @@ pub fn cargo_ndk(
 
     const SEP: char = '\x1f';
 
-    // Read initial RUSTFLAGS
+    // Read initial CARGO_ENCODED_/RUSTFLAGS
     let mut rustflags = match std::env::var("CARGO_ENCODED_RUSTFLAGS") {
         Ok(val) => val,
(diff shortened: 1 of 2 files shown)


ndk-build/src/cargo.rs
```diff
@@ -14,12 +14,43 @@ pub fn cargo_ndk(
...
+
     // Read initial RUSTFLAGS
     let mut rustflags = match std::env::var("CARGO_ENCODED_RUSTFLAGS") {
         Ok(val) => val,
-        Err(std::env::VarError::NotPresent) => "".to_string(),
+        Err(std::env::VarError::NotPresent) => {
+            match std::env::var("RUSTFLAGS") {
+                Ok(val) => {
+                    // Note this is roughly equivalent to what cargo does
+                    // https://github.com/rust-lang/cargo/blob/f6de921a5d807746e972d9d10a4d8e1ca21e1b1f/src/cargo/core/compiler/build_context/target_info.rs#L682-L690
```
msiglreith, line 26: taking the cargo code and adding a `join(SEP)` at end should achieve the same I think while staying close to the original source

MarijnS95:
I think this PR simply refers to the _documented behaviour_ of `CARGO_ENCODED_RUSTFLAGS` being _mutually exclusive_ wrt `RUSTFLAGS`? https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags
If so, please mention that explicitly in the commit message and code implementation.

`CARGO_ENCODED_RUSTFLAGS` (just `RUSTFLAGS`, previously) wasn't really read because of `cargo` possibly setting it, but to update the user environment rather than blindly overwriting it like other tools.

Do you want to document this bugfix in the changelog?

ndk-build/src/cargo.rs
```diff
@@ -14,12 +14,43 @@ pub fn cargo_ndk(
...
+                    let mut rf = String::with_capacity(val.len());
+                    if !val.is_empty() {
+                        for arg in val.split(' ').filter_map(|s| {
+                            let s = s.trim();
+                            if !s.is_empty() {
+                                Some(s)
+                            } else {
+                                None
+                            }
+                        }) {
```
MarijnS95, line 36: I'd `.map(|s| s.trim()).filter(|s| !s.is_empty())` this :wink:

MarijnS95:
Suggestion for the commit/PR message:

```md
ndk-build uses `CARGO_ENCODED_RUSTFLAGS` to get around flags/arguments with spaces in `RUSTFLAGS` that are impossible to escape. For correctness this implementation reads `CARGO_ENCODED_RUSTFLAGS` instead of blindly overwriting its contents, but as per [`build.rustflags` precedence rules](https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags) the use of this environment variable would disregard and contents in the more common `RUSTFLAGS` environment variable. To make this more convenient for users, propagate the contents of `RUSTFLAGS` into `CARGO_ENCODED_RUSTFLAGS` to ensure it doesn't get ignored.
```

Couple followup questions:
- What do we do if both `RUSTFLAGS` and `CARGO_ENCODED_RUSTFLAGS` are set? Error? Merge them?
- Do we unset `RUSTFLAGS` after propagating its contents into `CARGO_ENCODED_RUSTFLAGS`, pre-empting any duplication issues that might otherwise possibly happen if things change up/downstream?

Jake-Shadle:
> What do we do if both RUSTFLAGS and CARGO_ENCODED_RUSTFLAGS are set? Error? Merge them?

I think the only case that would be relevant would be if there was _another_ non-cargo tool that invoked cargo-apk and set `CARGO_ENCODED_RUSTFLAGS` incorrectly based on the RUSTFLAGS/profile/etc which would then cause cargo-apk to propagate the incorrect ones, but that seems like a bit of a stretch.

> Do we unset RUSTFLAGS after propagating its contents into CARGO_ENCODED_RUSTFLAGS, pre-empting any duplication issues that might otherwise possibly happen if things change up/downstream?

I've cleared that env var now, since once cargo is invoked the ENCODED ones are the only ones that it cares about barring bugs, and there is already eg https://github.com/dtolnay/rustflags for parsing those if user's need to access them in build scripts.

MarijnS95:
> I think the only case that would be relevant would be if there was _another_ non-cargo tool that invoked cargo-apk and set `CARGO_ENCODED_RUSTFLAGS` incorrectly based on the RUSTFLAGS/profile/etc which would then cause cargo-apk to propagate the incorrect ones, but that seems like a bit of a stretch.

I wouldn't mind (prefer, actually) to be on the defensive side, and just disallow the case where `RUSTFLAGS` happens to be set when we just found `CARGO_ENCODED_RUSTFLAGS` to be set, too.

Then only @msiglreith's comment remains :)

Jake-Shadle:
I mean I could do that suggestion but would mean an unnecessary extra vector allocation since join is a slice method not an iterator method but if you're ok with that.

ndk-build/src/cargo.rs
MarijnS95: Heh I hadn't even opened the source link in `cargo` to see that they were doing the same, except for the `|s| s.trim()` -> `str::trim` :ok_hand:. This isn't only more concise, it's also closer to the reference implementation yet again.

ndk-build/src/cargo.rs
```diff
@@ -14,12 +14,43 @@ pub fn cargo_ndk(
...
+                );
+            }
+
+            val
+        }
+        Err(std::env::VarError::NotPresent) => {
+            match std::env::var("RUSTFLAGS") {
+                Ok(val) => {
+                    // Note cargo also does this very simplified parsing of only
+                    // splitting on spaces, ignoring things like quoting
```
MarijnS95, line 34: > ignoring things like quoting

As clearly stated in the documentation, and implied by my suggested improved commit/PR description (I'll apply that while merging if you don't beat me to it) this is by design / specification. This comment only causes confusion, implying that some form of escaping by quoting (and maybe other "_things_"?) is available.

Again a request to reference https://doc.rust-lang.org/cargo/reference/environment-variables.html, clearly quoting:

> This string is split by whitespace; for a more robust encoding of multiple arguments, see CARGO_ENCODED_RUSTFLAGS.

ndk-build/src/cargo.rs
```diff
@@ -14,12 +14,43 @@ pub fn cargo_ndk(
...
+                    // https://github.com/rust-lang/cargo/blob/f6de921a5d807746e972d9d10a4d8e1ca21e1b1f/src/cargo/core/compiler/build_context/target_info.rs#L682-L690
+                    let to_join: Vec<_> = val
+                        .split(' ')
+                        .map(|s| s.trim())
+                        .filter(|s| !s.is_empty())
+                        .collect();
+
+                    cargo.env_remove("RUSTFLAGS");
+
+                    to_join.join(SEP)
```
MarijnS95, line 44: Feel free to swap these and get rid of `to_join`, you already have the value of `RUSTFLAGS` so it can be deleted from the (nested, anyway) `cargo` env right after `Ok(val)`.

MarijnS95 (approved):

PR #358, opened 2022-10-17
Read keystore path/password from env vars prior to package metadata

Jake-Shadle:
In the cargo-apk docs, it has

```ini
[package.metadata.android.signing.<profile>]
path = "$HOME/.android/debug.keystore"
keystore_password = "android"
```

But the `$HOME` is a lie, cargo-apk doesn't actually expand environment variables, this change just adds a _very_ simple and dumb expansion, only if an environment variable is present at the beginning, and it actually exists.

### DIFF
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- Profile signing information can now be specified via the `CARGO_APK_<PROFILE>_KEYSTORE` and `CARGO_APK_<PROFILE>_KEYSTORE_PASSWORD` environment variables. The environment variables take precedence over signing information in the cargo manifest. Both environment variables are required except in the case of the `dev` profile, which will fall back to the default password if the `CARGO_APK_DEV_KEYSTORE_PASSWORD` is not set. ([#358](https://github.com/rust-windowing/android-ndk-rs/pull/358))
+- Profile signing information can now be specified via the `CARGO_APK_<PROFILE>_KEYSTORE` and `CARGO_APK_<PROFILE>_KEYSTORE_PASSWORD` environment variables. The environment variables take precedence over signing information in the cargo manifest. Both environment variables are required except in the case of the `dev` profile, which will fall back to the default password if `CARGO_APK_DEV_KEYSTORE_PASSWORD` is not set. ([#358](https://github.com/rust-windowing/android-ndk-rs/pull/358))
 
 (0.9.5, released on 2022-10-14, was yanked due to unintentionally bumping MSRV through the `quick-xml` crate, and breaking `cargo apk --` parsing after switching to `clap`.)
 
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- Profile signing information can now be specified via the `CARGO_APK_<PROFILE>_KEYSTORE` and `CARGO_APK_<PROFILE>_KEYSTORE_PASSWORD` environment variables if the signing information isn't present in the cargo manifest. ([#358](https://github.com/rust-windowing/android-ndk-rs/pull/358))
+- Profile signing information can now be specified via the `CARGO_APK_<PROFILE>_KEYSTORE` and `CARGO_APK_<PROFILE>_KEYSTORE_PASSWORD` environment variables. The environment variables take precedence over signing information the cargo manifest. Both environment variables are required except in the case of the `dev` profile, which will fallback to the default password if the `CARGO_APK_DEV_KEYSTORE_PASSWORD` is not set. ([#358](https://github.com/rust-windowing/android-ndk-rs/pull/358))
 
 (0.9.5, released on 2022-10-14, was yanked due to unintentionally bumping MSRV through the `quick-xml` crate, and breaking `cargo apk --` parsing after switching to `clap`.)
 
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- The `path` component of the profile signing is now optional, if not specified the `CARGO_APK_<PROFILE>_KEYSTORE` environment variable is read to determine the path to the keystore. ([#358](https://github.com/rust-windowing/android-ndk-rs/pull/358))
+
 (0.9.5, released on 2022-10-14, was yanked due to unintentionally bumping MSRV through the `quick-xml` crate, and breaking `cargo apk --` parsing after switching to `clap`.)
 
 - Automate `adb reverse` port forwarding through `Cargo.toml` metadata ([#348](https://github.com/rust-windowing/android-ndk-rs/pull/348))
@@ -29,10 +31,10 @@
 # 0.9.0 (2022-05-07)
 
 - **Breaking:** Use `min_sdk_version` to select compiler target instead of `target_sdk_version`. ([#197](https://github.com/rust-windowing/android-ndk-rs/pull/197))
-  See https://developer.android.com/ndk/guides/sdk-versions#minsdkversion for more details.
+  See <https://developer.android.com/ndk/guides/sdk-versions#minsdkversion> for more details.
 - **Breaking:** Default `target_sdk_version` to `30` or lower (instead of the highest supported SDK version by the detected NDK toolchain)
   for more consistent interaction with Android backwards compatibility handling and its increasingly strict usage rules:
-  https://developer.android.com/distribute/best-practices/develop/target-sdk
+  <https://developer.android.com/distribute/best-practices/develop/target-sdk>
   ([#203](https://github.com/rust-windowing/android-ndk-rs/pull/203))
 - Allow manifest `package` property to be provided in `Cargo.toml`. ([#236](https://github.com/rust-windowing/android-ndk-rs/pull/236))
 - Add `MAIN` intent filter in `from_subcommand` instead of relying on a custom serialization function in `ndk-build`. ([#241](https://github.com/rust-windowing/android-ndk-rs/pull/241))
@@ -44,7 +46,7 @@
 
 - Fixed the library name in case of multiple build artifacts in the Android manifest.
 - Work around missing `libgcc` on NDK r23 beta 3 and above, by providing linker script that "redirects" to `libunwind`.
-  See https://github.com/rust-windowing/android-ndk-rs/issues/149 and https://github.com/rust-lang/rust/pull/85806 for more details.
+  See <https://github.com/rust-windowing/android-ndk-rs/issues/149> and <https://github.com/rust-lang/rust/pull/85806> for more details.
 
 # 0.8.1 (2021-08-06)
 
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- Non-dev profile signing information can now be specified via the `CARGO_APK_<PROFILE>_KEYSTORE` and `CARGO_APK_<PROFILE>_KEYSTORE_PASSWORD` environment variables if the signing information isn't present in the cargo manifest. ([#358](https://github.com/rust-windowing/android-ndk-rs/pull/358))
+- Profile signing information can now be specified via the `CARGO_APK_<PROFILE>_KEYSTORE` and `CARGO_APK_<PROFILE>_KEYSTORE_PASSWORD` environment variables if the signing information isn't present in the cargo manifest. ([#358](https://github.com/rust-windowing/android-ndk-rs/pull/358))
 
 (0.9.5, released on 2022-10-14, was yanked due to unintentionally bumping MSRV through the `quick-xml` crate, and breaking `cargo apk --` parsing after switching to `clap`.)
 
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-- Profile signing information can now be specified via the `CARGO_APK_<PROFILE>_KEYSTORE` and `CARGO_APK_<PROFILE>_KEYSTORE_PASSWORD` environment variables. The environment variables take precedence over signing information the cargo manifest. Both environment variables are required except in the case of the `dev` profile, which will fallback to the default password if the `CARGO_APK_DEV_KEYSTORE_PASSWORD` is not set. ([#358](https://github.com/rust-windowing/android-ndk-rs/pull/358))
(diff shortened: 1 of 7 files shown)


MarijnS95:
Fwiw I disagree with supporting the base case, we should either support env vars properly (on Linux _and_ Windows with their respective `$` / `%FOO%` formattings) or document this limiting implementation clearly in the readme.

At that point I much rather update the example to state that absolute paths should be used, and that _"the default is `.android/...` inside the users ""home"" directory"_.

Jake-Shadle:
The default only applies to the dev profile, building for any other profile requires a valid key path. This used to not be the case, so I'm totally fine removing this pr if the old path was returned.

Absolute paths are completely unacceptable as it breaks running apk on whatever platform doesn't support the absolute path format, there needs to be some way for the user to specify a path not relative to the manifest in a host platform agnostic way. 

The choice of $ or ‰‰ is irrelevant since this is not shell expansion and it does not matter what platform/shell the user is using.

MarijnS95:
> The default only applies to the dev profile, building for any other profile requires a valid key path. This used to not be the case, so I'm totally fine removing this pr if the old path was returned.

I wouldn't recommend using the debug key for release builds, but if you insist...

> Absolute paths are completely unacceptable as it breaks running apk on whatever platform doesn't support the absolute path format, there needs to be some way for the user to specify a path not relative to the manifest in a host platform agnostic way.

Relative is fine by me as well, have a look at [how `[env]` treats this](https://doc.rust-lang.org/cargo/reference/config.html#env).

> The choice of $ or ‰‰ is irrelevant since this is not shell expansion and it does not matter what platform/shell the user is using.

Anything is arbitrary at this point. I don't mind having a `CARGO_APK_RELEASE_KEYSTORE` env var otherwise, instead of limited env-var parsing support.

cargo-apk/README.md
```diff
@@ -52,9 +52,12 @@ runtime_libs = "path/to/libs_folder"
...
 # Defaults to `$HOME/.android/debug.keystore` for the `dev` profile. Will ONLY generate a new
 # debug.keystore if this file does NOT exist.
-# A keystore path is always required on the `release` profile.
+# If specified, the keystore path is a path relative to the Cargo.toml file, if
+# not specified, the environment variable `CARGO_APK_<PROFILE>_KEYSTORE` env
+# variable is read. The profile portion follows the same rules as <cfg>, it is
+# the uppercased profile name with `-` replaced with `_`.
 [package.metadata.android.signing.<profile>]
-path = "$HOME/.android/debug.keystore"
+path = "relative/path/my.keystore"
```
MarijnS95, line 60: The path can also be absolute, I made extra care to check that when submitting this:

https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.push

> If `path` is absolute, it replaces the current path.

cargo-apk/CHANGELOG.md
```diff
@@ -29,10 +31,10 @@
 # 0.9.0 (2022-05-07)
 
 - **Breaking:** Use `min_sdk_version` to select compiler target instead of `target_sdk_version`. ([#197](https://github.com/rust-windowing/android-ndk-rs/pull/197))
-  See https://developer.android.com/ndk/guides/sdk-versions#minsdkversion for more details.
+  See <https://developer.android.com/ndk/guides/sdk-versions#minsdkversion> for more details.
```
MarijnS95, line 34: Is it an autoformatter doing this?

cargo-apk/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- The `path` component of the profile signing is now optional, if not specified the `CARGO_APK_<PROFILE>_KEYSTORE` environment variable is read to determine the path to the keystore. ([#358](https://github.com/rust-windowing/android-ndk-rs/pull/358))
```
MarijnS95, line 3: Still a bit confused by the `path` being made optional, but the keystore password not?

MarijnS95 (commented):
A step in the right direction, but we should apply the same treatment to the password. Otherwise one might as well ask the user to hardcode the relative or absolute (yes, `.join()` supports an absolute RHS) path to the keystore, when they are already forced to provide the keystore _password_ for that specific file _right next to it anyway_.

cargo-apk/CHANGELOG.md
Jake-Shadle: Yes.

cargo-apk/CHANGELOG.md
Jake-Shadle: I guess I'll another environment variable.

MarijnS95: Thanks, I'd just make the path not optional again so that the user has to specify both or none in `Cargo.toml`.

cargo-apk/README.md
```diff
@@ -51,10 +51,15 @@ apk_name = "myapp"
...
 
 # Defaults to `$HOME/.android/debug.keystore` for the `dev` profile. Will ONLY generate a new
-# debug.keystore if this file does NOT exist.
-# A keystore path is always required on the `release` profile.
+# debug.keystore if this file does NOT exist. A keystore is never auto-generated
+# for other profiles.
+#
+# If specified, the keystore path can be absolute, or relative to the Cargo.toml
+# file. If not specified, the environment variable `CARGO_APK_<PROFILE>_KEYSTORE`
+# environment variable is read. The profile portion follows the same rules as
```
MarijnS95, line 59: Little `environment variable ... environment variable` repeat, and be sure to mention `_KEYSTORE_PASSWORD` :)

cargo-apk/CHANGELOG.md
MarijnS95: Bad autoformatter, bad :grimacing:

cargo-apk/src/apk.rs
```diff
@@ -231,7 +231,30 @@ impl<'a> ApkBuilder<'a> {
                 password: signing.keystore_password.clone(),
             },
             (None, true) => self.ndk.debug_key()?,
-            (None, false) => return Err(Error::MissingReleaseKey(profile_name.to_owned())),
+            (None, false) => {
```
MarijnS95, line 234: Tbh I don't mind supporting this for `dev` as well (after all, the documentation seems to imply that this should work OOTB), and I also think env vars should get precedence over config files - as is done with many other configurations like `RUSTFLAGS`? https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags

cargo-apk/README.md
```diff
@@ -51,10 +51,15 @@ apk_name = "myapp"
...
+# `<cfg>`, it is the uppercased profile name with `-` replaced with `_`.
 [package.metadata.android.signing.<profile>]
-path = "$HOME/.android/debug.keystore"
+path = "relative/or/absolute/path/my.keystore"
```
MarijnS95, line 62:
```suggestion
path = "relative/or/absolute/path/to/my.keystore"
```

MarijnS95 (commented):
Thanks a ton, sorry for pushing you through all this churn but I really like this direction and it's just bikeshedding things like documentation and the precedence of `Cargo.toml` `metadata` vs environment variables at this stage!

Don't forget to update the title/description of this PR to match the new direction, that makes it easier for me to find the PR and finally merge it :)

cargo-apk/src/apk.rs
```diff
@@ -225,13 +225,41 @@ impl<'a> ApkBuilder<'a> {
...
+            let path = {
+                let profile_env = format!("CARGO_APK_{}_KEYSTORE", env_profile_name);
+                std::env::var_os(&profile_env).map(PathBuf::from)
+            };
+
+            let password = {
+                let profile_env = format!(
+                    "CARGO_APK_{}_KEYSTORE_PASSWORD",
+                    profile_name.to_uppercase().replace('-', "_")
+                );
```
MarijnS95, line 245: Nit: `env_profile_name + "_PASSWORD"`?

cargo-apk/src/apk.rs
```diff
@@ -225,13 +225,41 @@ impl<'a> ApkBuilder<'a> {
 
         let signing_key = self.manifest.signing.get(profile_name);
 
-        let signing_key = match (signing_key, is_debug_profile) {
-            (Some(signing), _) => Key {
+        let signing_key = if let Some(signing) = signing_key {
```
MarijnS95, line 228: What's your thought on letting the env vars take precedence over what was configured in `self.manifest.signing`?

cargo-apk/src/apk.rs
```diff
@@ -225,13 +225,41 @@ impl<'a> ApkBuilder<'a> {
...
+            } else {
+                Key {
+                    path: path.ok_or_else(|| Error::MissingReleaseKey(profile_name.to_owned()))?,
+                    password: password
+                        .ok_or_else(|| Error::MissingReleaseKey(profile_name.to_owned()))?,
+                }
+            }
         };
 
         Ok(apk.add_pending_libs_and_align()?.sign(signing_key)?)
```
MarijnS95, line 265: How do you feel about using a little `println!("Signing APK with {}", signing_key.path);` here, just to at least help the user validate what is going on, in case they don't get the environment variables right?

cargo-apk/src/apk.rs
```diff
@@ -225,13 +225,41 @@ impl<'a> ApkBuilder<'a> {
...
+                let profile_env = format!(
+                    "CARGO_APK_{}_KEYSTORE_PASSWORD",
+                    profile_name.to_uppercase().replace('-', "_")
+                );
+
+                std::env::var(&profile_env).ok()
+            };
+
+            if is_debug_profile {
+                if let (Some(path), Some(password)) = (path, password) {
```
MarijnS95, line 251: Thinking about this some, both the path _and_ password are required. Do we silently ignore the `_KEYSTORE` env var when `_KEYSTORE_PASSWORD` isn't set, or should we `eprintln` a `warn` in that case? I can imagine users forgetting about the ""default"" `"android"` password for their debug keystore... :smile:

Jake-Shadle: Agreed, think it's reasonable for the debug store to fallback to the default password.

cargo-apk/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Profile signing information can now be specified via the `CARGO_APK_<PROFILE>_KEYSTORE` and `CARGO_APK_<PROFILE>_KEYSTORE_PASSWORD` environment variables. The environment variables take precedence over signing information the cargo manifest. Both environment variables are required except in the case of the `dev` profile, which will fallback to the default password if the `CARGO_APK_DEV_KEYSTORE_PASSWORD` is not set. ([#358](https://github.com/rust-windowing/android-ndk-rs/pull/358))
```
MarijnS95, line 3:
```suggestion
- Profile signing information can now be specified via the `CARGO_APK_<PROFILE>_KEYSTORE` and `CARGO_APK_<PROFILE>_KEYSTORE_PASSWORD` environment variables. The environment variables take precedence over signing information in the cargo manifest. Both environment variables are required except in the case of the `dev` profile, which will fall back to the default password if `CARGO_APK_DEV_KEYSTORE_PASSWORD` is not set. ([#358](https://github.com/rust-windowing/android-ndk-rs/pull/358))
```

cargo-apk/README.md
```diff
@@ -50,11 +50,20 @@ apk_name = "myapp"
...
-# A keystore path is always required on the `release` profile.
+# Defaults to `$HOME/.android/debug.keystore` for the `dev` profile. Will ONLY
+# generate a new debug.keystore if this file does NOT exist. A keystore is never
+# auto-generated for other profiles.
+#
+# The keystore path can be absolute, or relative to the Cargo.toml file.
+#
+# The environment variables `CARGO_APK_<PROFILE>_KEYSTORE` and
+# `CARGO_APK_<PROFILE>_KEYSTORE_PASSWORD` can be read to retrieve the keystore path
+# and keystore password respectively. The profile portion follows the same rules
```
MarijnS95, line 61:
```suggestion
# `CARGO_APK_<PROFILE>_KEYSTORE_PASSWORD` can be configured/set to contain the keystore path
# and keystore password respectively. The profile portion follows the same rules
```

(They are _read_ by `cargo-apk`, but the user has to _set_ them)

cargo-apk/src/apk.rs
```diff
@@ -223,18 +223,49 @@ impl<'a> ApkBuilder<'a> {
...
+                    return Err(Error::MissingReleaseKey(profile_name.to_owned()));
+                }
+            }
         };
 
-        Ok(apk.add_pending_libs_and_align()?.sign(signing_key)?)
+        let unsigned = apk.add_pending_libs_and_align()?;
+
+        println!(
+            "Signing '{}' with keystore '{}'",
```
MarijnS95, line 264: Nit: We typically use backticks here.

cargo-apk/src/apk.rs
```diff
@@ -223,18 +223,49 @@ impl<'a> ApkBuilder<'a> {
...
+                eprintln!(
+                    "{} not specified, falling back to default password",
+                    password_env
+                );
+                Key {
+                    path,
+                    password: ndk_build::ndk::DEFAULT_DEV_PASSWORD.to_owned(),
+                }
+            }
+            (_, _) => {
```
MarijnS95, line 247: `(None, _)` to be a bit more explicit?

ndk-build/src/apk.rs
```diff
@@ -32,7 +32,7 @@ impl ApkConfig {
             .join(format!("{}-unaligned.apk", self.apk_name))
     }
 
-    fn apk(&self) -> PathBuf {
+    pub fn apk(&self) -> PathBuf {
```
MarijnS95, line 35: Would be nice to add a doc comment here, and mention this and `DEFAULT_DEV_PASSWORD` in the changelog for `ndk-build` :)

ndk-build/src/ndk.rs
```diff
@@ -4,6 +4,8 @@ use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
+pub const DEFAULT_DEV_PASSWORD: &str = "android";
```
MarijnS95, line 7:
```suggestion
pub const DEFAULT_DEV_KEYSTORE_PASSWORD: &str = "android";
```

MarijnS95 (approved):
Logic looks solid now, just wording/naming nits :)

cargo-apk/src/apk.rs
Jake-Shadle: That makes it more complicated as there is no (Some, None) case then.

MarijnS95: That's actually a good catch, as it would be the perfect case for a warning/error explaining that if the user sets a  (non-dev) keystore path, they _must also_ provide the password for it - more reason to have it don't you think? :grin:

MarijnS95 (approved):

MarijnS95:
Seems like I have no rights to apply the final cleanup and merge this :)

PR #360, opened 2022-11-05
cargo-apk: Inherit package configuration (`version`) from workspace root

### LINKED ISSUE #359: Cargo apk returns `Error: Failed to parse config` when Cargo version inherits from workspace
Since Rust 1.64.0, it is possible to [inherit some package properties from the workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-workspacepackage-table).
This feature seems not yet supported by Cargo apk for the `package.version` property.

For example, running the command `cargo apk run --package my_package` in a workspace returns the following error:
```text
Error: Failed to parse config.

Caused by:
    invalid type: map, expected a string for key `package.version` at line 3 column 21
```
Where the `Cargo.toml` file of `my_package` is:
```toml
[package]
name = "my_package"
version.workspace = true
```

MarijnS95:
Depends on https://github.com/dvc94ch/cargo-subcommand/pull/23, https://github.com/dvc94ch/cargo-subcommand/pull/24
Fixes #359

Starting with [Rust 1.64] common `[package]` parameters (and dependencies) can now be specified in the workspace root manifest by setting `<field>.workspace=true` in a `[package]` and specifying its value in the workspace root manifest under [`[workspace.package]`][1].

Since `cargo-apk` reads the `version` field from `[package]` it has to support this new format and pull the value from the root manifest instead, in order to support projects utilizing this new format.

This is currently implemented ad-hoc for the version field, but could be done in a cleaner way if/when more fields are needed.

[Rust 1.64]: https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds
[1]: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-workspacepackage-table


msiglreith (approved):

MarijnS95:
I'll merge this in when closer to a `cargo-apk` release - which is basically exclusively blocking on #363 or a similar solution - at which point I'll also make a new breaking `cargo-subcommand` release.

Issue #368, opened 2022-11-21
error: failed to verify the checksum of `ndk-sys v0.4.0`


Fuuzetsu:
I suspect it's because of https://github.com/rust-lang/cargo/issues/7180

linked to upload to crates.io from couple of hours ago

![w](https://user-images.githubusercontent.com/893115/203178358-952a0f0c-50be-4232-8cbe-d5ce329a8c3a.png)

I guess maybe it's best to refrain from uploading such versions and instead doing 0.4.1, 0.4.2 and so on...

MarijnS95:
I suspect that's because of bad CI tuning: publishes should only happen on pushes to _TAGS_, but it seems we run it on `master` so it published the result of #324 before we bumped the version and made a formal release.

I've yanked the offending version that should not have been released ~apologies~ _blame our CI_ for the inconvenience.

PR #370, opened 2022-11-23
ndk-sys: Regenerate against 25.1.8937393 and release ndk-sys-0.4.2

MarijnS95:
Preempting some of the version-metadata related issues (#368/#369, our CI published ), release `ndk-sys-0.4.2` immediately.

Not sure if we should opt for `0.4.2` or `0.5.0`; bindgen did introduce some strictly semver breaking changes such as removing (and inlining) the `pub type size_t` alias, nothing breaking occurred in the `25.0.8775105` regen though.  Alternatively I can downgrade bindgen since such a breaking bump should otherwise go paired with a fresh (also breaking ☹️) `ndk` release so that folks don't get broken `dependabot` upgrade suggestions.  If we have to I might as well make some breaking changes in light of https://github.com/rust-windowing/android-ndk-rs/pull/353#discussion_r991048968, though.


ndk-sys/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Unreleased
 
+# 0.4.2 (2022-11-23)
+
+- Regenerate against NDK 25.1.8937393. (#370)
+- Regenerate against NDK 25.0.8775105. (#324)
```
MarijnS95, line 6: In other repos (`ash`) I just fold these together, perhaps this should just be:

```suggestion
- Regenerate against NDK 25.1.8937393 with `bindgen 0.63`. (#324, #370)
```

(and imply a bindgen update)

as nobody cares about an intermediate update that wasn't even released.

MarijnS95:
Okay, I've set everything up towards a release, but I don't think that I'll actually release this today nor (any time) soon unless someone is actively in need of an `ndk-sys` regen bump, in which case we should see towards providing `ndk` bindings as well. I'll just leave this PR open to take these changes into account whenever we're ready to perform the next breaking release (and allow doing nonbreaking smaller releases in the meantime), in hopes of picking up some more changes first.

Alternatively we can bump down `bindgen` too, if there's a need for NDK 25 bindings.

ndk/CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Unreleased
 
+# 0.8.0 (TODO)
+
 - event: Add `tool_type` getter for `Pointer`. (#323)
 - input_queue: Allow any non-zero return code from `pre_dispatch()` again, as per documentation. (#325)
+- **Breaking:** Upgrade to [`ndk-sys 0.5.0`](../ndk-sys/CHANGELOG.md#050-2022-11-23). (#370)
```
MarijnS95, line 7: Shouldn't forget to update the date here.

Issue #372, opened 2022-12-06
Modularize and decouple from rust-windowing


rib:
Following up from https://github.com/rust-windowing/android-ndk-rs/issues/365 here is a specific plan for us to modularize this repo

## Why?
Several of the crates are very orthogonal from each other:
- `ndk-glue` in particular is unlikely to see continued development / maintenance so it would make sense to clearly sign post from here to `android-activity` and provide an archive for posterity
- `cargo apk` may eventually be superseded by something like `xbuild` and it's also not inherently coupled to the the other `ndk` components
- Eventually we may aim to make `ndk-context` redundant if we can instead encourage crates that support Android to have explicit initialization APIs that may be given a jvm + Context/Activity (and therefore enabling multiple Activities in a single process)
- None of these crates are inherently `rust-windowing` specific and there is probably more opportunity to foster collaboration around an organisation that's more broadly related to mobile OSs in general.

## Where to?

Since it's anticipated that most people interested in Rust + Android are likely also interested cross-platform mobile support, we have created a general [rust-mobile](https://github.com/rust-mobile) organization, initially owned by the contributors to `android-ndk-rs` and `android-activity`.

The idea is that I will also move `android-activity` here, along with splitting the `android-activity/examples` out into a new repo.

I believe @dvc94ch was also open to moving `xbuild` here too

Other Rust utilities and build tools etc catering to supporting Rust development for mobile operating systems would also be welcome to create repos here

## How?

These are key repos that would be split out of `android-ndk-rs`:
- `ndk-glue` - A filter of the `ndk-glue` and `ndk-macro` subdirectories, updated with a notice to recommend `android-activity` for new projects and then archived for posterity.
- `cargo-apk` - would be a filter of the `cargo-apk` + `ndk-build` subdirectories and `ndk-examples` (renamed to `examples) and outstanding PRs + issues would be transferred to the new repo
- `ndk-context` - A filter of the `ndk-context` subdirectory
- `ndk` - After removing all the subdirectories that have been filtered into standalone repos this repo would then be transferred to `rust-mobile` (so all issues / PRs are preserved)

### How to filter the repos:

```
git clone https://github.com/newren/git-filter-repo
```

#### ndk-glue
```
git clone https://github.com/rust-windowing/android-ndk-rs.git ndk-glue
cd ndk-glue
[...]

PR #373, opened 2022-12-06
Removes ndk-glue + ndk-macro as part of #372 modularization

rib:
The ndk-glue and ndk-macro subdirectores have been split into a stand-alone repository at: https://github.com/rust-mobile/ndk-glue

This is being done as part of the modularization effort described in issue #372

The stand-alone repository was filtered like this:

```
git clone https://github.com/newren/git-filter-repo
git clone https://github.com/rust-windowing/android-ndk-rs.git ndk-glue
cd ndk-glue
py ../git-filter-repo/git-filter-repo \
    --path ndk-glue --path ndk-macro \
    --path ndk-examples --path-rename ndk-examples:examples \
    --path LICENSE-APACHE --path LICENSE-MIT \
    --path Cargo.toml --path rustfmt.toml \
    --path .github --path .gitignore
```

The last commit in this repo that is included within the stand-alone repository is 107f03e3858bdced3a7a898e43b339b73d7fc1af

Although ndk-glue is also being deprecated at the same time as splitting it out this commit doesn't make any README changes for this repo yet which will be addressed separately.

For now the ndk-examples continue to use the ndk-glue crate (the 0.7 release) and the current plan is to move these examples to the cargo-apk crate where they will be ported to use `android-activity`


dvc94ch (approved):
Thanks!

rib:
In addressing the last CI failure I've added a note that the ndk-examples are now using the 0.7 release of ndk-glue instead of the in-tree version. This requires a `[path.crates-io]` to ensure a consistent choice over the `ndk` crate implementation.

This doesn't look to port the examples away from ndk-glue at this point since the current idea is that the examples will actually be removed from this repo and instead maintained within the `cargo apk` repo (it's more compelling to have examples in the `cargo apk` repo for testing that the tool can package and deploy those examples as part of its CI)

Even if we decide we want to keep a copy of these examples in this repo too, it seems reasonable to handle porting to android-activity separately.

rib:
I'm closing this PR since I think it's going to be more practical to track all the removal changes related to #372 together

Please see https://github.com/rust-windowing/android-ndk-rs/pull/374 instead (currently a draft until I have made a patch to remove `cargo-apk`, `ndk-build` and `ndk-examples`)

PR #374, opened 2022-12-07
Remove split subdirectories as part of #372 modularization

rib:
This PR removes `ndk-glue`, `ndk-macro`, `cargo-apk`, `ndk-build`, `ndk-examples` and `ndk-context` from this repo as part of the modularization effort described in issue #372

- `ndk-glue` and `ndk-macro` have been moved to https://github.com/rust-mobile/ndk-glue
- `cargo-apk`, `ndk-build` and `ndk-examples` have been moved to https://github.com/rust-mobile/cargo-apk
- `ndk-context` has moved to https://github.com/rust-mobile/ndk-context

Supersedes: #373

PR #376, opened 2022-12-09
Release cargo-apk-0.9.7

### LINKED ISSUE #375: `cargo apk -- build --release` doesn't build in `release` profile anymore
I'm trying to use `cargo-apk` to only build a shared library from Rust which then I'll load and use from a "regular" (Java+Kotlin) Android project - so I don't need any APK generation, code signing, etc. from `cargo-apk`.

I was using `cargo apk -- build --release` for this, as suggested here: https://github.com/rust-windowing/android-ndk-rs/issues/326#issuecomment-1214638774

This used to work well, however, since then, it has stopped building in `release` mode for some reason, as can be seen here:
https://github.com/torokati44/ruffle-android/actions/runs/3623790481/jobs/6110051919

image

image

MarijnS95:
Fixes #375

### DIFF
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Removed known-arg parsing from `cargo apk --` to not make argument flags/values get lost, see also [#375](https://github.com/rust-windowing/android-ndk-rs/issues/375). ([#377](https://github.com/rust-windowing/android-ndk-rs/pull/377))
+
 # 0.9.6 (2022-11-23)
 
 - Profile signing information can now be specified via the `CARGO_APK_<PROFILE>_KEYSTORE` and `CARGO_APK_<PROFILE>_KEYSTORE_PASSWORD` environment variables. The environment variables take precedence over signing information in the cargo manifest. Both environment variables are required except in the case of the `dev` profile, which will fall back to the default password if `CARGO_APK_DEV_KEYSTORE_PASSWORD` is not set. ([#358](https://github.com/rust-windowing/android-ndk-rs/pull/358))
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+# 0.9.7 (2022-12-09)
+
+- Reimplement default package selection based on `$PWD` or `--manifest-path` by upgrading to [`cargo-subcommand 0.11.0`](https://github.com/rust-mobile/cargo-subcommand/releases/tag/0.11.0).
 - Removed known-arg parsing from `cargo apk --` to not make argument flags/values get lost, see also [#375](https://github.com/rust-windowing/android-ndk-rs/issues/375). ([#377](https://github.com/rust-windowing/android-ndk-rs/pull/377))
 
 # 0.9.6 (2022-11-23)
--- a/cargo-apk/Cargo.toml
+++ b/cargo-apk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cargo-apk"
-version = "0.9.6"
+version = "0.9.7"
 authors = ["The Rust Windowing contributors"]
 edition = "2018"
 description = "Helps cargo build APKs"
--- a/cargo-apk/Cargo.toml
+++ b/cargo-apk/Cargo.toml
@@ -13,7 +13,7 @@ rust-version = "1.60"
 
 [dependencies]
 anyhow = "1.0.57"
-cargo-subcommand = "0.10"
+cargo-subcommand = "0.11"
 clap = { version = "4", features = ["derive"] }
 dunce = "1"
 env_logger = "0.9"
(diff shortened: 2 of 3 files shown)


dvc94ch (changes requested):
wrong repo

MarijnS95:
Like I said I don't have much time for all this shuffling around. It's okay with me as long as it's not disruptive at this moment.

dvc94ch:
you're saying it's too inconvenient to open a PR in the correct repo?

MarijnS95:
It would be a bit inconvenient to transplant the commits to the other repo (but that has to happen regardless).

@rib mentioned in emails to owners for these crates that I won't be allowed to make releases, but I can since I am part of @rust-windowing/publishers, and `cargo` doesn't care which git repositories a crate is hosted at for that to work. Regardless, it'd be great if I could have direct control over the crates (and in turn grant publishing rights to a new group over at @rust-mobile).

dvc94ch:
Mmh you won't be allowed to make releases? I'm missing some context here. Are you and @rib still not getting along?

MarijnS95:
I would, unless I missed some context when @rib emailed the current owners of these crates (on crates.io) where he mentioned that. Which I thank him for sending that by the way, saves me some time and effort!

rib:
> @rib mentioned in emails to owners for these crates that I won't be allowed to make releases, but I can since I am part of @rust-windowing/publishers

aah, right, good point, for some reason I was thinking that once the repo is split then you'd end up loosing permission to publish because I was thinking that you were reliant on credentials that were associated with the android-ndk-rs repo.

If you're also a member of @rust-windowing/publishers then there's hopefully nothing to worry about in the short term, and we can follow up on any ownership changes for crates asynchronously

rib:
> Mmh you won't be allowed to make releases? I'm missing some context here. Are you and @rib still not getting along?

:) no worries here; @MarijnS95 isn't saying that I was asserting that they would be _disallowed_. I sent some emails to tomaka and Osspial asking them to give @MarijnS95 co-ownership of the cargo-apk and ndk-build crates on crates.io because I was worried that @MarijnS95 wouldn't be able to make releases from the standalone repo.

rib:
> It would be a bit inconvenient to transplant the commits to the other repo (but that has to happen regardless).

btw @MarijnS95 I was thinking we may be able to directly transfer any pending PRs/Issues for cargo-apk directly from this repo - though that requires that you have write permission to this repo to do that: https://docs.github.com/en/issues/tracking-your-work-with-issues/transferring-an-issue-to-another-repository

It would be rather awkward to land anything here because it means I have to re-run the filtering etc for the standalone repo.

If someone gives me write access to this repo I could look at transferring cargo-apk issues/PRs (probably best to wait until we merge PR #374 ideally)

Alternatively for smalls PRs like this then I can cherry pick the patches and create an equivalent PR

rib:
I've created an equivalent PR against the standalone repo here: https://github.com/rust-mobile/cargo-apk/pull/1

MarijnS95:
Disclaimer: What I ended up replying here should probably all have gone into #372...

---

> aah, right, good point, for some reason I was thinking that once the repo is split then you'd end up loosing permission to publish because I was thinking that you were reliant on credentials that were associated with the android-ndk-rs repo.
> 
> If you're also a member of @rust-windowing/publishers then there's hopefully nothing to worry about in the short term, and we can follow up on any ownership changes for crates asynchronously

Yes I think it's fine for now, but we have to make sure that @rust-windowing/publishers doesn't kick me out after effectively all repos I need publish rights for have moved out. Until I receive the appropriate rights of course.

I will also see about updating the GitHub secret to use a new token from my account, instead of publishing every release (performed by me for the past year-or-so) via Francesca's account :grimacing: 

> btw @MarijnS95 I was thinking we may be able to directly transfer any pending PRs/Issues for cargo-apk directly from this repo - though that requires that you have write permission to this repo to do that: https://docs.github.com/en/issues/tracking-your-work-with-issues/transferring-an-issue-to-another-repository

I definitely wanted to transfer relevant issues to the other repo before opening PRs against it. That doesn't seem possible for PRs though, since they're heavily dependent on the available git history (I don't even think you can do it between direct forks).

However, it seems I'm only allowed to transfer to other repos _under the @rust-windowing organization_:

image


> It would be rather awkward to land anything here because it means I have to re-run the filtering etc for the standalone repo.

That is true, and mentioned above by me; I just wasn't sure how far along your filtering was in terms of demonstrating "this is how it will end up looking like" versus "everything is moved now, please use the new system". Looks like the latter, I'll try and review your PRs soon to make sure the CI publishing keeps working when we land (the "transferred" version of) this PR.

> If someone gives me write access to this repo I could look at transferring cargo-apk issues/PRs (probably best to wait until we merge PR #374 ideally)

I don't have enough rights for that, unfortunately. @rust-windowing/publishers, any help here? Can you also give me enough rights to access the settings menu and initiate the transfer to @rust-mobile? @madsmtm?

> Alternatively for smalls PRs like this then I can cherry pick the patches and create an equivalent PR

It's weekend now, I can easily take care of transplanting my own PRs on Monday...

> I've created an equivalent PR against the standalone repo here: [rust-mobile/cargo-apk#1](https://github.com/rust-mobile/cargo-apk/pull/1)

But you already took care of that :no_mouth:

PR #377, opened 2022-12-09
Revert "Add `args: Args` back to `Ndk` subcommand to see them in `-h`"

### LINKED ISSUE #375: `cargo apk -- build --release` doesn't build in `release` profile anymore

MarijnS95:
Fixes #375

This reverts commit e2b9db4235138ee7e24d2c57c797a6c85698e0b2, merged as part of 419df14 ("cargo-apk: Reimplement "default" (`--`) subcommand trailing args for `cargo` (#363)").

### DIFF
--- a/cargo-apk/CHANGELOG.md
+++ b/cargo-apk/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Removed known-arg parsing from `cargo apk --` to not make argument flags/values get lost, see also [#375](https://github.com/rust-windowing/android-ndk-rs/issues/375). ([#377](https://github.com/rust-windowing/android-ndk-rs/pull/377))
+
 # 0.9.6 (2022-11-23)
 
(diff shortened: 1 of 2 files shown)


rib:
Please see the equivalent PR against the standalone repo here: https://github.com/rust-mobile/cargo-apk/pull/2

PR #378, opened 2022-12-22
Naming cleanup

MarijnS95:
- CI/publish: Replace owning organization with `rust-mobile`
- Replace links with new rust-mobile/ndk organization/repository
- Strip `android-` suffix from documentation links and ndk/ndk-sys names


rib (approved):
Looks good to me

PR #385, opened 2023-02-15
Update README to reflect modularization changes

rib:
This is fairly ruthless with the previous guidance that was in the README but now that this repo has a much more narrow scope this strips the README back to just the essentials that relate to the NDK bindings.

This also adds sign posting for all the crates that were split out from this repo, considering that the original `android-ndk-rs` repo URLs will redirect here, and people may wonder where these crates have moved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,175 +1,22 @@
-# Rust on Android
+> **Note**
+> This repository was recently [modularized](https://github.com/rust-mobile/ndk/issues/372) and the following crates were split into separate repositories:
+>
+> Crate | New Location | Notes
+> ------|--------------|------
+> ndk-context | https://github.com/rust-mobile/ndk-context |
+> ndk-glue | https://github.com/rust-mobile/ndk-glue | _deprecated_ - see [android-activity](https://github.com/rust-mobile/android-activity)
+> ndk-macro | https://github.com/rust-mobile/ndk-glue | _deprecated_ - see [android-activity](https://github.com/rust-mobile/android-activity)
+> ndk-build | https://github.com/rust-mobile/cargo-apk |
+> cargo-apk | https://github.com/rust-mobile/cargo-apk |
 
 [![Rust](https://github.com/rust-mobile/ndk/workflows/Rust/badge.svg)](https://github.com/rust-mobile/ndk/actions) ![MIT license](https://img.shields.io/badge/License-MIT-green.svg) ![APACHE2 license](https://img.shields.io/badge/License-APACHE2-green.svg) [![MSRV](https://img.shields.io/badge/rustc-1.60.0+-ab6000.svg)](https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html)
 
-
-Libraries and tools for Rust programming on Android targets:
+Rust bindings to the [Android NDK](https://developer.android.com/ndk)
 
 Name | Description | Badges
 --- | --- | ---
 [`ndk-sys`](./ndk-sys) | Raw FFI bindings to the NDK | [![crates.io](https://img.shields.io/crates/v/ndk-sys.svg)](https://crates.io/crates/ndk-sys) [![crates.io](https://docs.rs/ndk-sys/badge.svg)](https://docs.rs/ndk-sys)
 [`ndk`](./ndk) | Safe abstraction of the bindings | [![crates.io](https://img.shields.io/crates/v/ndk.svg)](https://crates.io/crates/ndk) [![crates.io](https://docs.rs/ndk/badge.svg)](https://docs.rs/ndk)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,20 +1,25 @@
-# Rust on Android
+> ⚠️ **Can't find what you're looking for?** ⚠️
```
MarijnS95, line 1: I don't like quotes for this... but learned last week that there's a way to get a specially highlighted block:

https://github.com/community/community/discussions/16925

README.md
```diff
@@ -1,20 +1,25 @@
...
+>
+> This repository was recently [modularized](https://github.com/rust-mobile/ndk/issues/372) and the following crates were split into separate repositories:
+>
+> Crate | New Location | Notes
+> ------|--------------|------
+> ndk-context | https://github.com/rust-mobile/ndk-context |
+> ndk-glue | https://github.com/rust-mobile/ndk-glue | _deprecated_ - see [android-activity](https://github.com/rust-mobile/android-activity)
+> ndk-build | https://github.com/rust-mobile/cargo-apk |
+> cargo-apk | https://github.com/rust-mobile/cargo-apk |
```
MarijnS95, line 10: Don't think anyone uses `ndk-macro` but we could mention it... Otherwise drop the pointer to `ndk-build` as well?

README.md
```diff
@@ -1,20 +1,25 @@
...
-[`ndk-context`](./ndk-context) | Android handles | [![crates.io](https://img.shields.io/crates/v/ndk-context.svg)](https://crates.io/crates/ndk-context)
-[`ndk-glue`](./ndk-glue) | Startup code | [![crates.io](https://img.shields.io/crates/v/ndk-glue.svg)](https://crates.io/crates/ndk-glue) [![crates.io](https://docs.rs/ndk-glue/badge.svg)](https://docs.rs/ndk-glue)
-[`ndk-build`](./ndk-build) | Everything for building apk's | [![crates.io](https://img.shields.io/crates/v/ndk-build.svg)](https://crates.io/crates/ndk-build) [![crates.io](https://docs.rs/ndk-build/badge.svg)](https://docs.rs/ndk-build)
-[`cargo-apk`](./cargo-apk) | Build tool | [![crates.io](https://img.shields.io/crates/v/cargo-apk.svg)](https://crates.io/crates/cargo-apk) [![crates.io](https://docs.rs/cargo-apk/badge.svg)](https://docs.rs/cargo-apk)
 
-See [`ndk-examples`](./ndk-examples) for examples using the NDK and the README files of the crates for more details.
+
+See these [`ndk-examples`](https://github.com/rust-mobile/cargo-apk/tree/main/examples/examples) and these [`rust-android-examples`](https://github.com/rust-mobile/rust-android-examples) for examples using the NDK.
 
 ## Supported NDK versions
```
MarijnS95, line 24: Let's move this to `cargo-apk`; the table only concerns itself with "can we find the right compiler" etc. The NDK code itself backwards-compatible after all and this crate only defines Rust'y wrappers for a subset of it. `ndk-sys` OTOH contains the target NDK version as metadata since it is generated to represent the entire API (and "supports" there is the wrong word too IMO).

README.md
rib: ah yep, forgot about ndk-macros - I think I just went by the list of crates that were removed from the previous table in the README. I'll add a line for ndk-macros.

README.md
rib: Yeah, it's mainly just so it gets formatted distinctly like a notice. Using the `> **Note**` markdown extension is probably better though.

MarijnS95: We could use either:

> **Note**

> **Warning**

I like the new format regardless :)

MarijnS95 (approved):

README.md
```diff
@@ -1,175 +1,22 @@
...
-)]
-fn main() {}
-```
 
-## JNI
-Java Native Interface (JNI) allows executing Java code in a VM from native applications. To access
-the JNI use the `AndroidContext` from the `ndk-context` crate. `ndk-examples` contains a `jni_audio`
-example which will print out all output audio devices in the log.
 
-- [`jni`](https://crates.io/crates/jni), JNI bindings for Rust
```
MarijnS95, line 175: Since there's _some_ tie-in between JNI and NDK it might be useful to retain minimal documentation around this? Not a hard requirement, probably in-line with your idea/comment to spend some more time later to write concise guides how to use this (these repos) in conjunction with each other?

MarijnS95: Let's tackle this at a later point.

MarijnS95:
@autumnontape the table has no meaning for the `ndk` crate, as explained in https://github.com/rust-mobile/ndk/pull/385#discussion_r1107293617. It isn't mentioned in `cargo-apk` yet because it's deprecated, and `xbuild` internally bundles toolchain bits from a known-working NDK.

PR #389, opened 2023-03-08
ndk: Drop optional `jni-glue` dependency

MarijnS95:
This crate doesn't seem to have taken off (3 years since last update) against the `jni` and `jni-sys` crate, and the `ndk` only uses it for a single (compile-testable) documentation example.


rib (approved):
yeah. I can understand the original sentiment since the `jni` crate wasn't being actively maintained either for years, but at least for now it seems reasonable to just support the `jni` and `jni-sys` crates

MarijnS95:
@rib this is only a documentation example after all, and a rather trivial one at that which would be super easy to repro for the few that might still be aware of and using `jni_glue`.

PR #390, opened 2023-03-08
ndk: Bump optional `jni` dependency for doctest-example to `0.21`

MarijnS95:
I can't tell why I initially called this "breaking because it is used in the public API" in #300 (bump to `0.19`) as `jni` is clearly only used for demonstration purposes in the doctestable documentation example.


ndk/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 - event: Add `tool_type` getter for `Pointer`. (#323)
 - input_queue: Allow any non-zero return code from `pre_dispatch()` again, as per documentation. (#325)
+- Bump optional `jni` dependency for doctest-example from `0.19` to `0.21`. (#390)
```
MarijnS95, line 5: `doctest-example` or `(doctest'able-) example`? Or do we have a clearer/concise way to refer to / name (testable!) code snippets/examples in documentation? Suggestions welcome!

rib: I probably wouldn't hyphenate and would write "doctest example". Hyphens are more often used to create compound adjectives. E.g. you could have a "doc-test example" and in that case "doc-test" serves as an adjective (but "doctest" without a hyphen is more appropriate here since that's an established term in this context). But also wouldn't too much about this - what you have written is perfectly readable / understandable I think.

rib (approved):

ndk/CHANGELOG.md
MarijnS95: Good call, I think I had the hyphen because of originally using `doctestable` as adjective, but then dropped `able`.

PR #391, opened 2023-04-10
Add dependabot support

notgull:
This PR adds dependabot support in order to make sure dependencies are automatically updated.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: weekly
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,7 +4,11 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    ignore:
+      - update-types: ["version-update:semver-patch", "version-update:semver-minor"]
   - package-ecosystem: github-actions
     directory: /
     schedule:
       interval: weekly
+    ignore:
+      - update-types: ["version-update:semver-patch", "version-update:semver-minor"]


notgull:
> On that note it better doesn't start spamming semver-compatible bumps: in that case we might better remove the minor version from `Cargo.toml` instead unless actively testing `cargo -Zminimal-versions` (which I recall doing on the `ndk` crate at some point).

I can configure it so that it only files PRs for major version bumps.

MarijnS95:
If major means "most significant", or semver incompatible (so _including_ `0.2` -> `0.3`) that'd be useful.

MarijnS95:
Thanks, really cool to know this config now but I'm still very curious as to whether this blocks strictly minor version which are actually semver-incompatible "major" bumps like the `0.2.0` -> `0.3.0` example above.

Will definitely start using this in an internal codebase where we're currently limiting dependabot spam in different ways.

.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
...
+    schedule:
+      interval: weekly
+    ignore:
+      - update-types: ["version-update:semver-patch", "version-update:semver-minor"]
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly
+    ignore:
+      - update-types: ["version-update:semver-patch", "version-update:semver-minor"]
```
MarijnS95, line 14: Nit: fine to leave the minor patches in for GH actions I think, the config only ever contains `@vX` with just the major version.

.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
+updates:
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: weekly
+    ignore:
+      - update-types: ["version-update:semver-patch", "version-update:semver-minor"]
```
MarijnS95, line 8: I just integrated this in an internal repo, and used to get an implicit `dependabot config OK` CI run way back in the day, but it's not there anymore.

Unfortunately now after merging it, the dependabot view complains:

```
The property '#/updates/0/ignore/0' did not contain a required property of 'dependency-name'
```

You'll have to fix the same here, and it's [documented](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#specifying-dependencies-and-versions-to-ignore):

> You can combine this with `dependency-name: "*"` to ignore particular `update-types` for all dependencies.

notgull:
Closing as I am now less enthused about this

MarijnS95:
@notgull because?

I'm personally leaning towards Renovate lately, it looks really promising in some repos I'm tracking and is much more featureful and fine-grained when it comes to handling Rust projects.

PR #394, opened 2023-05-27
Bump `bitflags` to `2.0.0`

daxpedda:
Changelog: https://github.com/bitflags/bitflags/blob/2.0.0/CHANGELOG.md#200.

This is a breaking change. The API that is implemented by the macro is now different.

### DIFF
--- a/ndk/Cargo.toml
+++ b/ndk/Cargo.toml
@@ -31,7 +31,7 @@ api-level-30 = ["api-level-29"]
 test = ["ffi/test", "jni", "all"]
 
 [dependencies]
-bitflags = "1.2.1"
+bitflags = "2.0.0"
 jni-sys = "0.3.0"
 num_enum = "0.5.1"
 raw-window-handle = "0.5"
--- a/ndk/src/looper.rs
+++ b/ndk/src/looper.rs
@@ -31,6 +31,7 @@ pub struct ThreadLooper {
 
 bitflags! {
     /// Flags for file descriptor events that a looper can monitor.
+    #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
     pub struct FdEvent: u32 {
         const INPUT = ffi::ALOOPER_EVENT_INPUT;
         const OUTPUT = ffi::ALOOPER_EVENT_OUTPUT;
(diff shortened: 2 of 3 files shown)


ndk/src/looper.rs
```diff
@@ -31,6 +31,7 @@ pub struct ThreadLooper {
 
 bitflags! {
     /// Flags for file descriptor events that a looper can monitor.
+    #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
MarijnS95, line 34: Let's keep these in a somewhat consistent order (yes it's messy like this in `native_window.rs`/`surface_texture.rs` too):

```suggestion
    #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
```

ndk/src/native_activity.rs
```diff
@@ -17,6 +17,7 @@ bitflags! {
     /// <https://developer.android.com/ndk/reference/group/native-activity#group___native_activity_1ga2f1398dba5e4a5616b83437528bdb28e>
     ///
     /// [`android.view.WindowManager.LayoutParams`]: https://developer.android.com/reference/android/view/WindowManager.LayoutParams
+    #[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
MarijnS95, line 20:
```suggestion
    #[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
```

MarijnS95 (approved):
Thanks! Looks like my request to fix the `bitflags` changelog is paying off :grinning:

daxpedda:
Done.

MarijnS95 (approved):

ndk/Cargo.toml
```diff
@@ -31,7 +31,7 @@ api-level-30 = ["api-level-29"]
 test = ["ffi/test", "jni", "all"]
 
 [dependencies]
-bitflags = "1.2.1"
+bitflags = "2.0.0"
```
MarijnS95, line 34: Hope folks don't get confused with this simply meaning `>= 2` :)

PR #395, opened 2023-06-07
Add fallible version of format(): try_format()

dcsommer:
We have observed panics in practice for this API due to unexpected formats. We need to compare the format without panicking.


ndk/src/bitmap.rs
```diff
@@ -136,9 +136,13 @@ impl AndroidBitmapInfo {
...
 
-    pub fn format(&self) -> BitmapFormat {
+    pub fn try_format(&self) -> Option<BitmapFormat> {
         let format = self.inner.format as u32;
-        format.try_into().unwrap()
+        format.try_into().ok()
+    }
+
+    pub fn format(&self) -> BitmapFormat {
+        self.try_format().unwrap()
```
MarijnS95, line 145: The downside is that this now panics on `None` rather than the descriptive error from `TryFromPrimitive` (which hopefully includes the `u32` number that failed to parse).

dcsommer: Fair enough. Updated to return the `Result`.

MarijnS95 (approved):

MarijnS95:
A doc-comment might be nice too, pointing the user in the right direction.

MarijnS95:
@dcsommer can you still add a doc-comment and changelog entry?

dcsommer:
Done. Let me know if it works for you.

ndk/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - hardware_buffer_format: Add `YCbCr_P010` and `R8_UNORM` variants. (#405)
 - **Breaking:** hardware_buffer_format: Add catch-all variant. (#407)
 - Add panic guards to callbacks. (#412)
+- Add `try_format()` to `AndroidBitmapInfo` to handle unexpected formats without panicking.
```
MarijnS95, line 18:
```suggestion
- bitmap: Add `try_format()` to `AndroidBitmapInfo` to handle unexpected formats without panicking. (#395)
```

And sort it in between entry `#394` and `#398`.

ndk/src/bitmap.rs
```diff
@@ -136,9 +136,23 @@ impl AndroidBitmapInfo {
...
+        format.try_into()
+    }
+
+    /// Attempt to convert the internal, native `AndroidBitmapInfo` into
+    /// this crate's [`BitmapFormat`] type.
+    ///
+    /// # Panics
+    ///
+    /// This function panics if the underlying value does not have a
+    /// corresponding variant in [`BitmapFormat`].
```
MarijnS95, line 153: Add something like: ``Use [`try_format()`][AndroidBitmapInfo::try_format() for an infallible version of this function``.

MarijnS95 (approved):

PR #397, opened 2023-06-15
ndk: Add bindings for AFont

paxbun:
This PR provides bindings for `AFont`, `AFontMatcher`, and `ASystemFontIterator` and some helper structs for the better abstraction of some parts of the OpenType specification.

Example:
```rust
for font in ndk::font::SystemFontIterator::new().unwrap() {
  log::debug!("Font: {}", font.path().display());
  for idx in 0..font.axis_count() {
    log::debug!("  Axis {idx} ({}): {}", font.axis_tag_at(idx), font.axis_value_at(idx));
  }
  log::debug!("  Collection Idx: {}", font.collection_index());
  log::debug!("  Locale: {:?}", font.locale());
  log::debug!("  Weight: {}", font.weight());
  log::debug!("  Is Italic: {}", font.is_italic());
}
```

Result (Tested on Galaxy Tab S8):
image


paxbun:
TODOs

- [ ] Reflect PR number to CHANGELOG.md
- [ ] Reflect NDK version (I used `25.1.8937393`) to CHANGELOG.md and Cargo.toml or re-generate `ndk-sys` bindings

MarijnS95:
Nice! I'll pull in the bindings regeneration via #370, and then have a look at this.

ndk-sys/generate_bindings.sh
```diff
@@ -1,10 +1,21 @@
...
+else
+    host_tag="linux-x86_64"
+fi
+
+sysroot="${ANDROID_NDK_ROOT}"/toolchains/llvm/prebuilt/${host_tag}/sysroot/
 [ ! -d "$sysroot" ] && echo "Android sysroot $sysroot does not exist!" && exit 1
 
 while read ARCH && read TARGET ; do
     bindgen wrapper.h -o src/ffi_$ARCH.rs \
+        --no-size_t-is-usize \
```
MarijnS95, line 18: I think I _intentionally_ got rid of that in #370, any reason to keep it?

paxbun: I found `event.rs` is using `size_t`, so I thought `size_t` is needed for some reason, and I didn't know the existence of #370. The last version of `ffi_*.rs` still had `size_t` before I modified it without `--no-size_t-is-usize`, so I encountered compile errors due to lack of `size_t`. So I thought it is something like a platform-specific behavior of `bindgen`, because I'm using mac now.

I also think removing `size_t` is better, but without `--no-size_t-is-usize`, the current code still makes compile errors due to other modules.

MarijnS95: The change is now merged in as I think the next release will be breaking anyway (lots of things queued up), so you should be safe to rebase on `main` and drop most of the regen changes here except adding the new API.

paxbun: Got it. Font bindings are not depending on `size_t`, so `font.rs` will be OK even without `--no-size_t-is-usize`.

paxbun: I merged (I'm sorry, but I have a lot of bad memories of `rebase`, it always broke the whole branch) the newly updated `master` branch and removed `--no-size_t-is-usize` from `generate_bindings.sh`.

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+impl FontWeight {
+    pub const fn new(value: u16) -> std::result::Result<Self, TryFromU16Error> {
+        if 0 < value && value <= 1000 {
+            Ok(Self(value))
+        } else {
+            Err(TryFromU16Error(()))
+        }
+    }
+
+    pub const fn value(&self) -> u16 {
```
MarijnS95, line 33:
```suggestion
    pub const fn value(self) -> u16 {
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+/// * [`AFONT_WEIGHT_*`]
+/// * [`Font::weight`]
+///
+/// [`AFONT_WEIGHT_*`]: https://developer.android.com/ndk/reference/group/font#anonymous-enum-33
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub struct FontWeight(u16);
+
+impl FontWeight {
+    pub const fn new(value: u16) -> std::result::Result<Self, TryFromU16Error> {
+        if 0 < value && value <= 1000 {
```
MarijnS95, line 26: Use existing `MIN` and `MAX` consts instead of open-coding this.

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    pub const THIN: FontWeight = FontWeight(ffi::AFONT_WEIGHT_THIN as u16);
+    pub const EXTRA_LIGHT: FontWeight = FontWeight(ffi::AFONT_WEIGHT_EXTRA_LIGHT as u16);
+    pub const LIGHT: FontWeight = FontWeight(ffi::AFONT_WEIGHT_LIGHT as u16);
+    pub const NORMAL: FontWeight = FontWeight(ffi::AFONT_WEIGHT_NORMAL as u16);
+    pub const MEDIUM: FontWeight = FontWeight(ffi::AFONT_WEIGHT_MEDIUM as u16);
+    pub const SEMI_BOLD: FontWeight = FontWeight(ffi::AFONT_WEIGHT_SEMI_BOLD as u16);
+    pub const BOLD: FontWeight = FontWeight(ffi::AFONT_WEIGHT_BOLD as u16);
+    pub const EXTRA_BOLD: FontWeight = FontWeight(ffi::AFONT_WEIGHT_EXTRA_BOLD as u16);
+    pub const BLACK: FontWeight = FontWeight(ffi::AFONT_WEIGHT_BLACK as u16);
+    pub const MAX: FontWeight = FontWeight(ffi::AFONT_WEIGHT_MAX as u16);
```
MarijnS95, line 46: `AFONT_WEIGHT_MIN` exists in the bindings too? Either have both or drop both :)

ndk-sys/generate_bindings.sh
```diff
@@ -1,6 +1,16 @@
...
+
+if [[ "$os" == "Darwin" ]]; then
+    host_tag="darwin-x86_64"
+elif [[ "$os" == "CYGWIN"* ]]; then
+    host_tag="windows-x86_64"
+else
+    host_tag="linux-x86_64"
+fi
+
+sysroot="${ANDROID_NDK_ROOT}"/toolchains/llvm/prebuilt/${host_tag}/sysroot/
```
MarijnS95, line 13: I'll probably push a regen to this PR later before merging it, since while it seems this unblocks you from writing the initial bindings a different host OS causes a lot of unnecessary code-motion and bloated diff in the generated files (even if they may in the end compile to roughly? the same crate).

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+
+/// The error type returned when an invalie font weight value is passed.
+#[derive(Debug)]
+pub struct TryFromU16Error(());
+
+impl fmt::Display for TryFromU16Error {
+    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            fmt,
+            "font weight must be positive and less than or equal to 1000"
```
MarijnS95, line 74: I tend to prefer `f.write_str()` in these cases.

And:

```suggestion
            "Font weight must be positive and less than or equal to 1000"
```

MarijnS95: Otherwise, how about printing the faulty value after storing it inside `self`?

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+        FontWeight::new(value)
+    }
+}
+
+/// A 4-byte integer representing an OpenType axis tag.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub struct AxisTag(u32);
+
+impl AxisTag {
+    pub const fn from_be(value: u32) -> std::result::Result<Self, TryFromU32Error> {
```
MarijnS95, line 94:
```suggestion
    pub const fn from_be(value: u32) -> Result<Self, TryFromU32Error> {
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "AxisTag(")?;
+        fmt::Display::fmt(self, f)?;
+        write!(f, " {:#x})", self.0)
+    }
+}
+
+/// The error type returned when an invalie font weight value is passed.
+#[derive(Debug)]
+pub struct TryFromU32Error(());
```
MarijnS95, line 135:
```suggestion
pub struct TryFromU32Error;
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+
+impl fmt::Debug for AxisTag {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "AxisTag(")?;
+        fmt::Display::fmt(self, f)?;
+        write!(f, " {:#x})", self.0)
+    }
+}
+
+/// The error type returned when an invalie font weight value is passed.
```
MarijnS95, line 133:
```suggestion
/// The error type returned when an invalid axis tag value is passed.
```

MarijnS95: And this is not the weight... but the axis tag?

Fwiw this and the comment on `TryFromU16Error` (and the debug prints) are nonsensical unless the error type is renamed to `AxisTagFromU32` or something.

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+            bytes[0] as char, bytes[1] as char, bytes[2] as char, bytes[3] as char
+        )
+    }
+}
+
+impl fmt::Debug for AxisTag {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "AxisTag(")?;
+        fmt::Display::fmt(self, f)?;
+        write!(f, " {:#x})", self.0)
```
MarijnS95, line 129: What is wrong with using:
```suggestion
        write!(f, "AxisTag({} {:x})", self, self.0)
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+            FontWeight::SEMI_BOLD => writeln!(f, "Semi Bold (Demi Bold)"),
+            FontWeight::BOLD => writeln!(f, "Bold"),
+            FontWeight::EXTRA_BOLD => writeln!(f, "Extra Bold (Ultra Bold)"),
+            FontWeight::BLACK => writeln!(f, "Black (Heavy)"),
+            _ => writeln!(f, "{}", self.0),
+        }
+    }
+}
+
+/// The error type returned when an invalie font weight value is passed.
```
MarijnS95, line 66:
```suggestion
/// The error type returned when an invalid font weight value is passed.
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+pub struct AxisTag(u32);
+
+impl AxisTag {
+    pub const fn from_be(value: u32) -> std::result::Result<Self, TryFromU32Error> {
+        // Each byte in a tag must be in the range 0x20 to 0x7E.
+        // See https://learn.microsoft.com/en-us/typography/opentype/spec/otff#data-types for details.
+        macro_rules! check_in_valid_range {
+            ($($byte: expr),+) => {
+                $(
+                    if !(0x20 <= ($byte) && ($byte) <= 0x7E) {
```
MarijnS95, line 100: Hmm `char::is_ascii()` only checks the upper-bound, not the lower one.

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+impl Iterator for SystemFontIterator {
+    type Item = Font;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        NonNull::new(unsafe { ffi::ASystemFontIterator_next(self.ptr.as_ptr()) })
+            .map(|p| unsafe { Font::from_ptr(p) })
+    }
+}
+
+#[cfg(feature = "api-level-29")]
```
MarijnS95, line 396: So far seems that everything useful is `api-level-29`... Can we just wrap the module in it with a `#!` at the top of the file?

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+/// [`AFont *`]: https://developer.android.com/ndk/reference/group/font#afont_close
+#[cfg(feature = "api-level-29")]
+#[derive(Debug)]
+pub struct Font {
+    ptr: NonNull<ffi::AFont>,
+}
+
+#[cfg(feature = "api-level-29")]
+impl Font {
+    /// Create an `Font` from a pointer
```
MarijnS95, line 159:
```suggestion
    /// Create a `Font` from a pointer
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    ///
+    /// # Safety
+    /// By calling this function, you assert that the pointer is a valid pointer to a native
```
MarijnS95, line 162:
```suggestion
    /// By calling this function, you assert that it is a valid pointer to a native
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+pub struct FontMatcher {
+    ptr: NonNull<ffi::AFontMatcher>,
+}
+
+#[cfg(feature = "api-level-29")]
+impl FontMatcher {
+    /// Create an `FontMatcher` from a pointer
+    ///
+    /// # Safety
+    /// By calling this function, you assert that the pointer is a valid pointer to a native
```
MarijnS95, line 302:
```suggestion
    /// By calling this function, you assert that it is a valid pointer to a native
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    ptr: NonNull<ffi::AFont>,
+}
+
+#[cfg(feature = "api-level-29")]
+impl Font {
+    /// Create an `Font` from a pointer
+    ///
+    /// # Safety
+    /// By calling this function, you assert that the pointer is a valid pointer to a native
+    /// `AFont`.
```
MarijnS95, line 163: I did my best to make all these intraodc links to the FFI type... Seems these were inherited from old docs that have not had such a cleanup pass :/

ndk/src/font.rs
MarijnS95: And this should mention that it assumes ownership of `ptr` (since you drop `_close` it later). `NativeWindow` docs are probably the most up-to-date as a reference for how the docs should look.

ndk/src/font.rs
MarijnS95: E.g. a good example from `NativeWindow`:

```rust
    /// `ptr` must be a valid pointer to an Android [`ffi::ANativeWindow`].
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    /// Returns the pointer to the native `AFont`.
+    pub fn ptr(&self) -> NonNull<ffi::AFont> {
+        self.ptr
+    }
+
+    /// Return a count of font variation settings associated with the current font.
+    ///
+    /// The font variation settings are provided as multiple tag-values pairs.
```
MarijnS95, line 175:
```suggestion
    /// The font variation settings are provided as multiple tag-value pairs.
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    ///
+    /// For example, bold italic font may have following font variation settings: 'wght' 700,
+    /// 'slnt' -12. In this case, [`Font::axis_count`] returns 2 and [`Font::axis_tag_at`] and
```
MarijnS95, line 178:
```suggestion
    /// 'slnt' -12. In this case, [`Font::axis_count()`] returns 2 and [`Font::axis_tag_at()`] and
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// [`Font::axis_value_at`] will return following values.
```
MarijnS95, line 179:
```suggestion
    /// [`Font::axis_value_at()`] will return following values.
```

But where are these "following" values? The sentence stops there. Upstream docs have a full example here!

I know this is copied verbatim from Android docs, but they keep giving me tears for how poor the wording is :disappointed:

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    pub fn axis_count(&self) -> usize {
+        unsafe { ffi::AFont_getAxisCount(self.ptr.as_ptr()) }
+    }
+
+    /// Return an OpenType axis tag associated with the current font.
```
MarijnS95, line 184:
```suggestion
    /// Returns an OpenType axis tag associated with the current font.
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    ///
+    /// See [`Font::axis_count`] for more details.
```
MarijnS95, line 186:
```suggestion
    /// See [`Font::axis_count()`] for more details.
```

Are there any more such occurrences?

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    pub fn axis_tag_at(&self, idx: usize) -> AxisTag {
+        // Android returns Axis Tag in big-endian.
+        // See https://cs.android.com/android/platform/superproject/+/refs/heads/master:frameworks/base/native/android/system_fonts.cpp;l=197 for details
+        AxisTag(unsafe { ffi::AFont_getAxisTag(self.ptr.as_ptr(), idx as u32) })
```
MarijnS95, line 190: Any reason to not self-document and validate this via `AxisTag::from_be()`?

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    /// Return an OpenType axis value associated with the current font.
```
MarijnS95, line 193:
```suggestion
    /// Returns an OpenType axis value associated with the current font.
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    ///
+    /// See [`Font::axis_count`] for more details.
```
MarijnS95, line 195:
```suggestion
    /// See [`Font::axis_count()`] for more details.
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    pub fn axis_value_at(&self, idx: usize) -> f32 {
+        unsafe { ffi::AFont_getAxisValue(self.ptr.as_ptr(), idx as u32) }
+    }
+
+    /// Return a font collection index value associated with the current font.
```
MarijnS95, line 200:
```suggestion
    /// Returns a font collection index value associated with the current font.
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    ///
+    /// In case the target font file is a font collection (e.g. .ttc or .otc), this returns a non
```
MarijnS95, line 202:
```suggestion
    /// In case the target font file is a font collection (e.g. `.ttc` or `.otc`), this returns a non
```
And isn't it `non-negative` with a hyphen?

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// negative value as an font offset in the collection. This always returns 0 if the target font
```
MarijnS95, line 203:
```suggestion
    /// negative value as a font offset in the collection. This always returns 0 if the target font
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// Return an absolute path to the current font file.
+    ///
+    /// Here is a list of font formats returned by this method:
+    ///
+    /// * OpenType
+    /// * OpenType Font Collection
+    /// * TrueType
+    /// * TrueType Collection
+    ///
+    /// The file extension could be one of *.otf, *.ttf, *.otc or *.ttc.
```
MarijnS95, line 218: And monospace (backticks) these all, to prevent accidentally creating bold sections.

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// The font file returned is guaranteed to be opened with `O_RDONLY`.
```
MarijnS95, line 219: Is it opened, or open**able**? I don't see a file descriptor (of an open file) here... only a path.

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// Return a font collection index value associated with the current font.
+    ///
+    /// In case the target font file is a font collection (e.g. .ttc or .otc), this returns a non
+    /// negative value as an font offset in the collection. This always returns 0 if the target font
+    /// file is a regular font.
+    pub fn collection_index(&self) -> usize {
+        unsafe { ffi::AFont_getCollectionIndex(self.ptr.as_ptr()) }
+    }
+
+    /// Return an absolute path to the current font file.
```
MarijnS95, line 209:
```suggestion
    /// Returns an absolute path to the current font file.
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// * TrueType Collection
+    ///
+    /// The file extension could be one of *.otf, *.ttf, *.otc or *.ttc.
+    /// The font file returned is guaranteed to be opened with `O_RDONLY`.
+    pub fn path(&self) -> &Path {
+        let path = unsafe { CStr::from_ptr(ffi::AFont_getFontFilePath(self.ptr.as_ptr())) };
+        OsStr::from_bytes(path.to_bytes()).as_ref()
+    }
+
+    /// Return a IETF BCP47 compliant language tag associated with the current font.
```
MarijnS95, line 225:
```suggestion
    /// Returns an IETF BCP47 compliant language tag associated with the current font.
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// For information about IETF BCP47, read [`Locale.forLanguageTag(java.lang.String)`].
+    ///
+    /// [`Locale.forLanguageTag(java.lang.String)`]: https://developer.android.com/reference/java/util/Locale.html#forLanguageTag(java.lang.String)
+    pub fn locale(&self) -> Option<&CStr> {
+        let ptr = unsafe { ffi::AFont_getLocale(self.ptr.as_ptr()) };
+        if ptr.is_null() {
+            None
+        } else {
+            Some(unsafe { CStr::from_ptr(ptr) })
+        }
```
MarijnS95, line 236: Could `let ptr = NonNull::new(ptr)?;` help us out here?

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    /// Return a weight value associated with the current font.
```
MarijnS95, line 239:
```suggestion
    /// Returns a weight value associated with the current font.
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    /// | 500   | Medium                    | [`FontWeight::MEDIUM`]      |
+    /// | 600   | Semi Bold (Demi Bold)     | [`FontWeight::SEMI_BOLD`]   |
+    /// | 700   | Bold                      | [`FontWeight::BOLD`]        |
+    /// | 800   | Extra Bold (Ultra Bold)   | [`FontWeight::EXTRA_BOLD`]  |
+    /// | 900   | Black (Heavy)             | [`FontWeight::BLACK`]       |
+    pub fn weight(&self) -> FontWeight {
+        FontWeight(unsafe { ffi::AFont_getWeight(self.ptr.as_ptr()) })
+    }
+
+    /// Return true if the current font is italic, otherwise returns false.
```
MarijnS95, line 259:
```suggestion
    /// Returns [`true`] if the current font is italic, otherwise returns [`false`].
```

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    ptr: NonNull<ffi::AFontMatcher>,
+}
+
+#[cfg(feature = "api-level-29")]
+impl FontMatcher {
+    /// Create an `FontMatcher` from a pointer
+    ///
+    /// # Safety
+    /// By calling this function, you assert that the pointer is a valid pointer to a native
+    /// `AFontMatcher`.
```
MarijnS95, line 303: Same comments about generalizing this.

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    pub unsafe fn from_ptr(ptr: NonNull<ffi::AFontMatcher>) -> Self {
+        Self { ptr }
+    }
+
+    /// Returns the pointer to the native `AFontMatcher`.
+    pub fn ptr(&self) -> NonNull<ffi::AFontMatcher> {
+        self.ptr
+    }
+
+    /// Select the best font from given parameters.
```
MarijnS95, line 313: No parameters are given.

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    ///
+    /// Creates a new [`FontMatcher`] object.
+    pub fn new() -> Self {
+        NonNull::new(unsafe { ffi::AFontMatcher_create() })
+            .map(|p| unsafe { FontMatcher::from_ptr(p) })
+            .expect("AFontMatcher_create returned NULL")
```
MarijnS95, line 319: How about having the `.expect()` first, making the `map` unnecessary.

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    /// Set family variant to matcher.
```
MarijnS95, line 322: Bit of a crooked sentence. So `FamilyVariant` _is_ a matcher?

But `self` is already the matcher, so perhaps it meant to store/apply the argument **in/to** the matcher?
Same below.

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    ///
+    /// If this function is not called, the matcher performs with [`FamilyVariant::Default`].
```
MarijnS95, line 324:
```suggestion
    /// If this function is not called, the match is performed with [`FamilyVariant::Default`].
```

Same below.

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    pub fn set_family_variant(&mut self, family_variant: FamilyVariant) {
+        unsafe { ffi::AFontMatcher_setFamilyVariant(self.ptr.as_ptr(), family_variant as u32) }
+    }
+
+    /// Set font locales to matcher.
+    ///
+    /// If this function is not called, the matcher performs with empty locale list.
+    ///
+    /// # Arguments
+    /// * `language_tags`: a null character terminated comma separated IETF BCP47 compliant language tags.
```
MarijnS95, line 334: No need for a single argument? Also, drop "null character terminated" blah, that is implicit because of the `CStr` type :)

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    ptr: NonNull<ffi::ASystemFontIterator>,
+}
+
+#[cfg(feature = "api-level-29")]
+impl SystemFontIterator {
+    /// Create an `SystemFontIterator` from a pointer
+    ///
+    /// # Safety
+    /// By calling this function, you assert that the pointer is a valid pointer to a native
+    /// `ASystemFontIterator`.
```
MarijnS95, line 369: And the same again :)

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        NonNull::new(unsafe { ffi::ASystemFontIterator_next(self.ptr.as_ptr()) })
+            .map(|p| unsafe { Font::from_ptr(p) })
+    }
+}
+
+#[cfg(feature = "api-level-29")]
+impl Drop for SystemFontIterator {
+    fn drop(&mut self) {
+        unsafe { ffi::ASystemFontIterator_close(self.ptr.as_ptr()) }
```
MarijnS95, line 399: This is documented as "freeing any related resources". Does that include the fonts that are returned, or should the caller clean those up manually? (I assume the latter, since this seems to be the only way to get hold of an `AFont` object, and that is what the return docs for `ASystemFontIterator_next()` somewhat clarify)

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+            .map(|p| unsafe { SystemFontIterator::from_ptr(p) })
+    }
+}
+
+#[cfg(feature = "api-level-29")]
+impl Iterator for SystemFontIterator {
+    type Item = Font;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        NonNull::new(unsafe { ffi::ASystemFontIterator_next(self.ptr.as_ptr()) })
```
MarijnS95, line 391: The docs say the following about this return value:

> a font. If no more font is available, returns nullptr. You need to release the returned font by `ASystemFont_close` when it is no longer needed.

Does that exist, or are they referring to `AFont_close()` instead?

paxbun: I thought they meant `AFont_close`; there's no function named `ASystemFont_close`. The writer must have confused `AFont` and `ASystemFontIterator`.

ndk-sys/wrapper.h
```diff
@@ -20,8 +20,10 @@
 #include <android/dlext.h>
 #include <android/fdsan.h>
 #include <android/file_descriptor_jni.h>
-// #include <android/font.h>
-// #include <android/font_matcher.h>
+typedef struct AFont AFont;
```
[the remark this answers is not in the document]
paxbun: Without that, `bindgen` fails to generate the bindings, and I thought the `#include` directives were commented out due to this... `<android/font.h>` uses `AFont*` without the typedef (which is allowed in C++), but other headers have something like `typedef struct ASomeStruct ASomeStruct`... Would there be a better workaround?

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+            FontWeight::EXTRA_BOLD => writeln!(f, "Extra Bold (Ultra Bold)"),
+            FontWeight::BLACK => writeln!(f, "Black (Heavy)"),
+            _ => writeln!(f, "{}", self.0),
+        }
+    }
+}
+
+/// The error type returned when an invalie font weight value is passed.
+#[derive(Debug)]
+pub struct TryFromU16Error(());
```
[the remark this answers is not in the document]
paxbun: I made this from `core::num::error::TryFromIntError`; that private `()` field is to prevent constructions of the structs from outer crates.

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
+//! Bindings for [`AFont`], [`AFontMatcher`], and [`ASystemFontIterator`]
+//!
+//! [`AFont`]: https://developer.android.com/ndk/reference/group/font#afont_close
+//! [`AFontMatcher`]: https://developer.android.com/ndk/reference/group/font#afontmatcher_create
+//! [`ASystemFontIterator`]: https://developer.android.com/ndk/reference/group/font#asystemfontiterator_open
```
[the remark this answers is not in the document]
paxbun: Yes, I linked to `AFont_close` because it is the first function. I can't understand why `AFont` does not have typedefs for the structs. I'll link `AFont` to the page itself. Do you think linking `AFontMatcher` and `ASystemFontIterator` to `_create` or `_open` be useful? Since they're constructors (effectively), I think linking to those functions is useful.

ndk/src/font.rs
paxbun: I also assumed the latter because the current code (https://cs.android.com/android/platform/superproject/+/master:frameworks/base/native/android/system_fonts.cpp;l=373) returns an owning pointer.

ndk/src/font.rs
```diff
@@ -0,0 +1,401 @@
...
+/// See the followings for more details:
+/// * [`AFONT_WEIGHT_*`]
+/// * [`Font::weight`]
+///
+/// [`AFONT_WEIGHT_*`]: https://developer.android.com/ndk/reference/group/font#anonymous-enum-33
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug)]
+pub struct FontWeight(u16);
+
+impl FontWeight {
+    pub const fn new(value: u16) -> std::result::Result<Self, TryFromU16Error> {
```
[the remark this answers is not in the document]
paxbun: Fixed!

ndk/src/font.rs
paxbun: No problem, even I don't understand why I wrote so 🤣 . BTW, Do you think Prefixing `0x` is unnecessary in this case?

ndk/src/font.rs
paxbun: This is copied from the original documentation; since the IETF thing is also mentioned in `Font::locale`, I think leaving it alone is better. I'll remove the "null character terminated" part.

ndk/src/font.rs
paxbun: `from_be` only validates whether each byte is between 32 and 127; there is no way we can distinguish whether the tag is a standard `wght` tag or a custom-designed `thgw` tag.

ndk/src/font.rs
paxbun: I don't have deep knowledge of English grammar, but don't opened and openable mean the same thing in this case? How about "The font file specified by the returned path is ..."?

ndk/src/font.rs
paxbun: I usually use the form "Sets <field name> of <class name>." How about "Sets the family variant of the font to be matched."?

paxbun: BTW, I forgot to make a corresponding function of `AFontMatcher_match`! I'll make another commit for this.

ndk/src/font.rs
paxbun: I think both `MIN` and `MAX` are needed. The problem is, `AFONT_WEIGHT_MIN` is 0, but the OpenType spec says a font weight should be greater than 0. Considering the case like [std::num::NonZeroUsize](https://doc.rust-lang.org/stable/std/num/struct.NonZeroUsize.html#associatedconstant.MIN), `FontWeight::MIN` should be `FontWeight(1)`, but that's not consistent with other bindings.

paxbun: I edited that part as follows, how about this?
```rust
    pub const fn new(value: u16) -> Result<Self, TryFromU16Error> {
        if Self::MIN.0 <= value && value <= Self::MAX.0 {
        ...
    }

    // + 1 is to keep consistency between FontWeight and other types like std::num::NonZero*
    pub const MIN: FontWeight = FontWeight(ffi::AFONT_WEIGHT_MIN as u16 + 1);
```

ndk/src/font.rs
paxbun: Fixed!
```rust
/// `ptr` must be a valid owning pointer to an Android [`ffi::AFont`].
```

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: It seems not. Fixed!

ndk/src/font.rs
paxbun: Replaced all such occurrences with third person singular form. Fixed!

ndk/src/font.rs
paxbun: Replaced all such occurrences with third person singular form. Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: It is. Fixed!

ndk/src/font.rs
paxbun: I edited that like the following. How about this?
```rust
/// Creates a new [`FontMatcher`] object. [`FontMatcher`] selects the best font from the
/// parameters set by the user.
```

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: Fixed! Please see below :)

ndk/src/font.rs
```diff
@@ -0,0 +1,397 @@
...
+//! [`AFontMatcher`]: https://developer.android.com/ndk/reference/group/font#afontmatcher_create
+//! [`ASystemFontIterator`]: https://developer.android.com/ndk/reference/group/font#asystemfontiterator_open
+
+use std::convert::TryFrom;
+use std::ffi::{CStr, OsStr};
+use std::fmt::{self, Write};
+use std::os::unix::prelude::OsStrExt;
+use std::path::Path;
+use std::ptr::NonNull;
+use std::result::Result;
```
MarijnS95, line 13: I suggested to drop the qualification because:

1. You didn't use it in all places;
2. It is in scope by default through Rust's prelude: https://doc.rust-lang.org/std/prelude/index.html.

```suggestion
```

ndk/src/font.rs
MarijnS95: So `MIN` is exclusive, `MAX` is inclusive... We could just use a Rust `Range` to be more clear about the intent (may or may not be public):

```rust
    const VALID_RANGE: RangeInclusive<u16> =
        ffi::AFONT_WEIGHT_MIN as u16 + 1..=ffi::AFONT_WEIGHT_MAX as u16;
```

In favour of `MIN` and `MAX`?

ndk/src/font.rs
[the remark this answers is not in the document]
MarijnS95: Not seeing any harm in users constructing this themselves, but sure.

ndk-sys/wrapper.h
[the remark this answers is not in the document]
MarijnS95: A comment explaining that these headers are broken for C compilers would be nice.

(And yes, I think this list is a copy-paste of the file list of an older NDK, with everything commented out that didn't generate cleanly)

ndk/src/font.rs
paxbun: I fixed the documentation. I'll also rename that errors.

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: No problem. Fixed!

ndk/src/font.rs
paxbun: Fixed!

ndk/src/font.rs
paxbun: ptr is `*const u8`, we could use `let ptr = NonNull::new(ptr as *mut u8)?;`, but I want to avoid casting `*const` to `*mut`.

ndk/src/font.rs
paxbun: I forgot that `Result` is included in the prelude. Actually, I use the full path when it comes to `Result` or `Error` because many crates have types with the same name... Fixed!

ndk/src/font.rs
paxbun: Fixed! Also, I've just found that `0x20 ' '` is only for trailing padding, so I added checks for that as well.
```rust
        // Each byte in a tag must be in the range 0x20 to 0x7E. A space character cannot be
        // followed by a non-space character. A tag must have one to four non-space characters.
        // See https://learn.microsoft.com/en-us/typography/opentype/spec/otff#data-types for details.

        let bytes = value.to_be_bytes();

        macro_rules! check_if_valid {
            ($e:expr ; $($f:expr)+) => {
                if !(bytes[$e] as char).is_ascii_graphic() {
                    return if true $(&& bytes[$f] == b' ')+ {
                        Ok(Self(value))
                    } else {
                        Err(TryFromU32Error(()))
                    };
                }
            };
        }

        // A tag must have one to four non-space characters.
        if !(bytes[0] as char).is_ascii_graphic() {
            return Err(TryFromU32Error(()));
        }

        check_if_valid!(1; 1 2 3);
        check_if_valid!(2; 2 3);
        check_if_valid!(3; 3);

        Ok(Self(value))
```

ndk/src/font.rs
paxbun: I did my best, how about this?
image

ndk-sys/wrapper.h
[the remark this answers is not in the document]
paxbun: I added explanations like the following, Would it be sufficient with this?
```cpp
// <android/font.h> is broken without the following typedef in C
typedef struct AFont AFont;
#include <android/font.h>
// <android/font_matcher.h> is broken without the following typedef in C
typedef struct AFontMatcher AFontMatcher;
#include <android/font_matcher.h>
...
// <android/system_fonts.h> is broken without the following typedef in C
typedef struct ASystemFontIterator ASystemFontIterator;
#include <android/system_fonts.h>
```

ndk/src/font.rs
paxbun: There are many custom numeric types like `NonZero*` or `rust_decimal::Decimal` in Rust, and AFAIK, such types always define inclusive `MIN` and `MAX`, as standard types such as `i32` also do so. I think `MIN` and `MAX` is sufficient for this.

ndk/src/font.rs
paxbun: I'll replace them with `write_str`, but shouldn't error messages start with a lowercase letter? AFAIK, all error messages in `std` start with a lowercase letter.

ndk/src/font.rs
paxbun: Fixed! I'll keep the prefix `0x`.
```rust
write!(f, "AxisTag({} {:#x})", self, self.0)
```

ndk/src/font.rs
paxbun: I renamed the types. I also added detailed checks and made `AxisTagValueError` into an enum.
image

MarijnS95:
@paxbun just a quick heads-up that I haven't been able to catch up to this before leaving for a while. I hope that's not a problem and I'll do my very best to pick this up in about 2 weekes, apologies for all the delays :crossed_fingers:

ndk/src/font.rs
```diff
@@ -0,0 +1,457 @@
...
+            Err(FontWeightValueError(()))
+        }
+    }
+
+    pub const fn to_u16(self) -> u16 {
+        self.0
+    }
+
+    // + 1 is to keep consistency between FontWeight and other types like std::num::NonZero*
+    pub const MIN: FontWeight = FontWeight(ffi::AFONT_WEIGHT_MIN as u16 + 1);
```
MarijnS95, line 40: More specifically: consistency here means that it can be used as range-inclusive?

It is indeed very strange to have `0 < x <= MAX` while we're usually doing zero-based indexing with `0 <= x < MAX`, but maybe it makes sense for font weights.

ndk/src/font.rs
```diff
@@ -0,0 +1,457 @@
...
+                    // A tag must have one to four non-space characters.
+                    AxisTagValueError::EmptyTag
+                } else {
+                    // A space character cannot be followed by a non-space character.
+                    AxisTagValueError::InvalidSpacePadding
+                },
+            );
+        }
+
+        macro_rules! check_if_valid {
```
MarijnS95, line 126: Can we replace all this macro magic with:

1. A reverse iterator that counts the number of uninterrupted trailing spaces;
2. A `position()` iterator that finds the index of the first space;
3. If there is a space found in 2., and its index is lower than what the number of spaces found in 1. allows, error.

ndk/src/font.rs
```diff
@@ -0,0 +1,457 @@
...
+                )+
+            };
+        }
+        check_byte_range!(0 1 2 3);
+
+        if bytes[0] == b' ' {
+            return Err(
+                if bytes[1] == b' ' && bytes[2] == b' ' && bytes[3] == b' ' {
+                    // A tag must have one to four non-space characters.
+                    AxisTagValueError::EmptyTag
```
MarijnS95, line 118: Could this one-off case be much more simply: `if bytes == b"    "`?

ndk/src/font.rs
```diff
@@ -0,0 +1,457 @@
...
+        // See https://learn.microsoft.com/en-us/typography/opentype/spec/otff#data-types for details.
+
+        let bytes = value.to_be_bytes();
+
+        // Each byte in a tag must be in the range 0x20 to 0x7E.
+        macro_rules! check_byte_range {
+            ($($e:expr)+) => {
+                $(
+                    if !(bytes[$e] as char).is_ascii_graphic() && bytes[$e] != b' ' {
+                        return Err(AxisTagValueError::InvalidCharacter);
```
MarijnS95, line 107: This could also be done with an iterator, instead of a macro?

MarijnS95: Or... a `for` loop?

MarijnS95: That implies dropping `const` from the `fn`, I don't see why that is needed.

ndk/src/font.rs
```diff
@@ -0,0 +1,457 @@
...
+    }
+
+    pub const fn to_be_bytes(self) -> [u8; 4] {
+        self.0.to_be_bytes()
+    }
+}
+
+impl fmt::Display for AxisTag {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let bytes = self.0.to_be_bytes();
```
MarijnS95, line 161: Is `self.to_be_bytes()` not useful here?

ndk/src/font.rs
```diff
@@ -0,0 +1,457 @@
...
+
+        check_if_valid!(1; 2 3);
+        check_if_valid!(2; 3);
+
+        // Whether or not bytes[3] is b' ', value is a valid axis tag.
+        Ok(Self(value))
+    }
+
+    pub const fn from_be_bytes(value: [u8; 4]) -> Result<Self, AxisTagValueError> {
+        Self::from_be(u32::from_be_bytes(value))
```
MarijnS95, line 147: How about doing it the opposite way around, by having all the logic inside `from_be_bytes()` and make `from_be(u32)` call `from_be_bytes(value.to_be_bytes())`?

ndk/src/font.rs
MarijnS95: Perhaps, but that means they also botched "the returned font" which should have said "the returned font iterator".

Most likely we have to close both the font _and_ the iterator.

ndk/src/font.rs
MarijnS95: Indeed, I didn't mean to drop the `#`, thanks!

ndk/src/font.rs
MarijnS95: Perhaps monospacing `'wght 700`, `'slnt' -12` would help, maybe "returns `2`", and maybe a colon at the end of the sentence before the example starts. Maybe replace `For example,` with `# Example`. Looks good otherwise :+1: 

Oh and the `log::debug!` suggests there's a colon that you could add in the `// Output:` example :grimacing:

ndk/src/font.rs
```diff
@@ -0,0 +1,457 @@
...
+    /// ```no_run
+    /// use ndk::font::Font;
+    ///
+    /// let font: Font = unimplemented!();
+    /// for idx in 0..font.axis_count() {
+    ///     log::debug!("{}: {}", font.axis_tag_at(idx), font.axis_value_at(idx));
+    /// }
+    /// // Output:
+    /// // wght 700
+    /// // slnt -12
```
MarijnS95, line 245:
```suggestion
    /// // wght: 700
    /// // slnt: -12
```

ndk/src/font.rs
```diff
@@ -0,0 +1,457 @@
...
+    /// The font variation settings are provided as multiple tag-value pairs.
+    ///
+    /// For example, bold italic font may have following font variation settings: 'wght' 700,
+    /// 'slnt' -12. In this case, [`Font::axis_count()`] returns 2 and [`Font::axis_tag_at()`] and
+    /// [`Font::axis_value_at()`] return those variation names and the corresponding values.
+    ///
+    /// ```no_run
+    /// use ndk::font::Font;
+    ///
+    /// let font: Font = unimplemented!();
```
MarijnS95, line 239:
```suggestion
    /// # let font: Font = unimplemented!();
```

Maybe better to hide this? Feel free to also hide the `use` then?

ndk/src/font.rs
```diff
@@ -0,0 +1,457 @@
...
+    pub fn ptr(&self) -> NonNull<ffi::AFont> {
+        self.ptr
+    }
+
+    /// Returns a count of font variation settings associated with the current font.
+    ///
+    /// The font variation settings are provided as multiple tag-value pairs.
+    ///
+    /// For example, bold italic font may have following font variation settings: 'wght' 700,
+    /// 'slnt' -12. In this case, [`Font::axis_count()`] returns 2 and [`Font::axis_tag_at()`] and
```
MarijnS95, line 233:
```suggestion
    /// For example, bold italic font may have following font variation settings: `'wght' 700`,
    /// `'slnt' -12`. In this case, [`Font::axis_count()`] returns `2` and [`Font::axis_tag_at()`] and
```

ndk/src/font.rs
MarijnS95: `to be opened` AFAIK either implies that it has previously been opened and is still open when it is returned to you, or that "something somewhere else" validates that it will definitely be opened at some point?

It seems, but I am not sure, that neither is true here.

ndk/src/font.rs
MarijnS95: Indeed, except that there are no real field names here because it's all internal parameters. Feel free to either use your "sets" variant which is closer to Androids alien documentation, or the following suggestion which also describes what the `FontMatcher` will _end up doing_ with this information:

`Only match/return fonts with the given xxx` where `xxx` is either `locale(s)`, `language(s)` or `style`.

Or "limits", or any other sentence variant you're comfortable with.

ndk/src/font.rs
paxbun: To me, that sentence sounds like "It is guaranteed that the user can open the returned font file with `O_RDONLY`", I think "the user" part is dropped, and the sentence is written in the passive voice. If you feel confused, we can rewrite that part with the active voice like the following:
"It is guaranteed that the user can open the font file specified by the returned path with `O_RDONLY`."

ndk/src/font.rs
paxbun: I did write that part with macros to make constructors of `AxisTag` const. I thought `AxisTag::from_be` should be const because it is merely a wrapper of `u32` but with a value validation; since we can't include all kinds of axis tags in `ndk`, some would use `ndk::font` like the following.
```
const WEIGHT_TAG: AxisTag = AxisTag::from_be_bytes(*b"wght");
const MY_CUSTOM_TAG: AxisTag = AxisTag::from_be_bytes(*b"cstm");
```
Without the const constructor, users should use `lazy_static!` for integer constants.

BTW, I should have added constructors that panic for invalid values and return the unwrapped value in const context...

ndk/src/font.rs
paxbun: If `AxisTag::from_be` is not const, then I can use `if &bytes == b"   "`.

ndk/src/font.rs
paxbun: Yes, consistency here means that the `MIN` to `MAX` range is inclusive.

I investigated several implementations in other frameworks, and it seems that FontWeight classes usually don't include constants for `MIN` or `MAX`:
* Class has a constructor with an integer parameter
  * WPF `System.Windows.FontWeight`
  * Compose `androidx.compose.ui.text.font.FontWeight`
* Restricts the user to select one of the nine members of the enum
  * JavaFX `javafx.scene.text.FontWeight`
  * UIKit `UIFontWeight`
  * Flutter `FontWEight`

Usually, `MIN` to `MAX` range is always inclusive, not only for `std::num::NonZero*`, but also similar integer types in other languages.
* C++ `std::numeric_limits<T>::min()` & `std::numeric_limits<T>::max()`
* Java `Integer.MIN_VALUE` & `Integer.MAX_VALUE`
* Swift `Int.min` & `Int.max`
* Go `math.MinInt` & `math.MaxInt`

Making this range exclusive allows constructing an integer instant with an invalid value, and I think that's why `std::num::NonZeroU*::MIN` is 1.

MarijnS95: I mean to explain that (what you wrote in the first sentence) in the comment (that could perhaps be a doc-comment?).

ndk/src/font.rs
MarijnS95: Yeah, it seems we can `const`-ly `panic` then. Sure to keep this if you see usefulness in this constructor being const, but a comment above the macro magic explaining this would be nice (for a future contributor to replace once more of `std` becomes `const`-stable).

ndk/src/font.rs
MarijnS95: We can't, because this function has to be `const fn`. Disregard this suggestion.

ndk-sys/generate_bindings.sh
MarijnS95: Done. Unfortunately thanks to the way `wrapper.h` is ordered this adds some new `_bindgen_ty_xxx` which bump the numbers everywhere else in this file :(

ndk/src/font.rs
```diff
@@ -0,0 +1,457 @@
...
+}
+
+impl Font {
+    /// Creates a [`Font`] from a pointer.
+    ///
+    /// # Safety
+    /// `ptr` must be a valid owning pointer to an Android [`ffi::AFont`].
+    pub unsafe fn from_ptr(ptr: NonNull<ffi::AFont>) -> Self {
+        Self { ptr }
+    }
```
MarijnS95, line 221: I cannot seem to find this comment anymore (perhaps it was erroneously marked as resolved), but can you use the more descriptive form of how this affects ownership? I am specifically referring to https://github.com/rust-mobile/ndk/pull/207/files#diff-5b2b3fe248dc9e2d1e48c16ec4e228a6319be5d167dda15a40547fbac1a7bb71R28:

```
/// Assumes ownership of `ptr`
```

After all nothing is really "created" here, it's a very cheap (almost transparent) wrapper around the given pointer.

Same for the other `fn from_ptr()`'s added here; all types _assume ownership_ of the given pointer and free it on `Drop`.

MarijnS95:
By the way, a friendly request to not reply `Fixed!` to every review comment in what is hopefully the last round of changes so that we can merge this in. My notification inbox is filled with those messages while lacking context, and GitHub uses the "Resolve conversation" button for that.

paxbun:
@MarijnS95 I think I mostly applied your suggestion, and I also checked whether there are any `ffi::*` functions not wrapped. I also added constructors that constly panic to AxisTag.

MarijnS95:
Note that I reported the header issues at https://github.com/android/ndk/issues/1920, because `surface_control.h` (which I need and am writing a wrapper for) has the same (and many more C++-like) issues.

Hopefully we can drop the workarounds when upstream picks them up.

ndk/src/font.rs
```diff
@@ -0,0 +1,544 @@
...
+        unsafe {
+            Font::from_ptr(
+                NonNull::new(ffi::AFontMatcher_match(
+                    self.ptr.as_ptr(),
+                    family_name.as_ptr(),
+                    text.as_ptr(),
+                    text.len() as _,
+                    run_length_out.map_or(std::ptr::null_mut(), |u| u),
+                ))
+                .expect("AonftMatcher_match returned NULL"),
```
MarijnS95, line 466:
```suggestion
                .expect("AFontMatcher_match returned NULL"),
```

ndk/src/font.rs
```diff
@@ -0,0 +1,544 @@
...
+    ///
+    /// Here is a list of font formats returned by this method:
+    ///
+    /// * OpenType
+    /// * OpenType Font Collection
+    /// * TrueType
+    /// * TrueType Collection
+    ///
+    /// The file extension could be one of `*.otf`, `*.ttf`, `*.otc` or `*.ttc`.
+    /// The font file specified by the returned path is guaranteed to be opened with `O_RDONLY`.
```
MarijnS95, line 335:
```suggestion
    /// The font file specified by the returned path is guaranteed to be openable with `O_RDONLY`.
```

This is only a path that the user may only print or whatever, it is not guaranteed to _be opened_ at some point in the future.

MarijnS95 (commented):
Marked most of the conversations as resolved, nice job clearing it all up!

I only have nits now, and if you fix the conflicts to make the CI run it might have a complaint or two, but after that I'll merge this in!

MarijnS95 (approved):

ndk/src/font.rs
MarijnS95: It looks like you suggested alternative wording at https://github.com/rust-mobile/ndk/pull/397#discussion_r1275619971, feel free to push ahead with that, as long as it is different than what this comment says now :)

paxbun:
I applied all the suggestions you mentioned! (except for the axis tag one)

ndk/src/font.rs
MarijnS95: Sure let's trust the OS.

MarijnS95 (approved):

PR #398, opened 2023-06-15
ndk: Upgrade `num_enum` crate from `0.5.1` to `0.6`

### LINKED ISSUE #392: Update `num_enum` crate to `0.6.1`
This is also applicable to `android-activity`.

This is causing duplicates on the `syn` dependency.

MarijnS95:
Fixes #392, cc @kchibisov


kchibisov:
ack. Would suggest to look into other duplicated deps in winit brought by android crates.

I know that `bitflags` should be updated, but maybe more... We have deny checks and references in the `deny.toml` inside the winit, so it might give some hints.

MarijnS95:
@kchibisov yeah I know, that is the goal before the next breaking release and `bitflags` is the only remaining crate afaik, which is already bumped in another PR.

PR #399, opened 2023-06-15
ndk: Rework media error types

MarijnS95:
CC @paxbun

- Clean up the names: we're already in the `ndk` crate so no need to repeat that in the type name;
- Make the types available outside the `media` module, so that they can be used in `AMidi` bindings which are not located inside `libmediandk.so` (but `libamidi.so`);
- Add documentation comments;
- Guard `dead_code` helper functions with the right `feature` bound.

### DIFF
--- a/ndk/src/media_error.rs
+++ b/ndk/src/media_error.rs
@@ -1,9 +1,22 @@
-use super::Result;
+//! Bindings for NDK media status codes.
+//!
+//! Also used outside of `libmediandk.so` in `libmidi.so` for example.
+
+#![cfg(feature = "media")]
+// The cfg(feature) bounds for some pub(crate) fn uses are non-trivial and will become even more
+// complex going forward.  Allow them to be unused when compiling with certain feature combinations.
+#![allow(dead_code)]
+
+use std::{mem::MaybeUninit, ptr::NonNull};
+
 use thiserror::Error;
 
+pub type Result<T, E = MediaError> = std::result::Result<T, E>;
+
+/// Media Status codes for [`media_status_t`](https://developer.android.com/ndk/reference/group/media#group___media_1ga009a49041fe39f7bdc6d8b5cddbe760c)
 #[repr(i32)]
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub enum MediaErrorResult {
+pub enum MediaStatus {
     CodecErrorInsufficientResource = ffi::media_status_t::AMEDIACODEC_ERROR_INSUFFICIENT_RESOURCE.0,
     CodecErrorReclaimed = ffi::media_status_t::AMEDIACODEC_ERROR_RECLAIMED.0,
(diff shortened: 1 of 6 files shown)


MarijnS95:
Thinking of doing the same little tweaks to `BitmapResult` and friends.

paxbun:
LGTM! By the way, `construct` or `construct_never_null` emit unused warnings when both media or midi are not enabled. Is that okay? IIRC, I tried to put some `#[cfg(...)]` to remove those warnings, but we kept it unchanged because of this rewriting.

MarijnS95:
Yes indeed. It has bothered me too, but not enough to come up with a more restrictive cfg bound. Those are generally error-prone and would require extensive checking of all `media`, `api-level-24`, and later `midi` combinations of features in the CI.

With this PR `MediaError::from_status()` is also added to the list, fwiw :)

MarijnS95:
I'm considering at least constraining this with `media`, then you can add `midi` to it in https://github.com/rust-mobile/ndk/pull/353.

MarijnS95:
@paxbun done, moved and fixed the cfg feature bounds. You'll have to add `midi` to the top-level `#![cfg(feature = "media")]` -> `#![cfg(any(feature = "media", feature = "midi")]` but can hopefully leave the other `cfg`s intact as they automatically get enabled via `"midi" -> "api-level-29" -> "api-level-24"`.

Though as you don't use `construct_never_null()`, this might need yet another bound :(

MarijnS95:
Otherwise I might as well switch back to `#[allow(dead_code)]` (while still guarding the module as a whole) and document that the `cfg`s are nontrivial and would be hard to test and maintain all combinations.

MarijnS95:
> Though as you don't use `construct_never_null()`, this might need yet another bound :(

Yup it does :(

paxbun:
I also think guarding the whole module with `cfg` and adding `#[allow(dead_code)]` to each function is the best now. This would be the case where we should use explicit `allow`.

ndk/src/media_error.rs
```diff
@@ -1,9 +1,22 @@
-use super::Result;
+//! Bindings for NDK media status codes.
+//!
+//! Also used outside of `libmediandk.so` in `libmidi.so` for example.
+
+#![cfg(feature = "media")]
+// The cfg(feature) bounds for some pub(crate) fn uses are non-trivial and will become even more
+// complex going forward.  Allow them to be unused when compiling with certain feature combinations.
+#![allow(dead_code)]
```
MarijnS95, line 8: A good example of this is when `AMidi` comes online in #353. `construct_never_null()` is used by `media+api-level-24`, but unused by `midi` (implicitly enables `api-level-24`, hence `api-level-29`). Adding a `#[cfg(all(feature = "media", feature = "api-level-24"))]` to that might work for now, but will be complicated to maintain in the future as we do not test any of these feature combinations... and only to satisfy a lint!

rib (approved):
looks good to me!

PR #400, opened 2023-06-15
ndk/media_codec: Inline `get_unlikely_to_be_null()`

MarijnS95:
Back when I refactored these in #248 and #256 the closure argument should have been removed entirely as it is only called immediately without any arguments, so its return value could have been passed instead.  At that point the function becomes simple enough to warrant removal entirely, inlining the unwrap and even allowing a more natural string literal error message in `.expect()`.

CC @zarik5


zarik5:
I agree this is an improvement. The function name `get_unlikely_to_be_null` was a bit complex and obscure, that's usually a sign the function should be broken down or removed.

MarijnS95:
At least it's descriptive, which can't be said for the other functions (does it "assume" this is never null? Or validate that it isn't?).

PR #401, opened 2023-06-16
ndk/media_codec: Wrap common dequeued-buffer status codes in enum

MarijnS95:
Supersedes #316, CC @zarik5
Closes #316

The dequeue API for output buffers can commonly return `AMEDIACODEC_INFO_OUTPUT_FORMAT_CHANGED` or `AMEDIACODEC_INFO_OUTPUT_BUFFERS_CHANGED`, both of which are not caught in `NdkMediaError::from_status()` and show up as an "unknown" error.  Neither really is an error that the user would want to bubble up from the `Result<>` via `?`, but should rather handle this explicitly.

A similar case already exists for `AMEDIACODEC_INFO_TRY_AGAIN_LATER` which previously resulted in a `None`, but is now also encapsuled in a self-documenting manner, for the dequeue_input_buffer()` API as well.

### DIFF
--- a/ndk/src/media/media_codec.rs
+++ b/ndk/src/media/media_codec.rs
@@ -291,8 +291,7 @@ impl MediaCodec {
         }
     }
 
-    /// Returns [`None`] if timeout is reached.
-    pub fn dequeue_input_buffer(&self, timeout: Duration) -> Result<Option<InputBuffer>> {
+    pub fn dequeue_input_buffer(&self, timeout: Duration) -> Result<DequeuedInputBufferResult<'_>> {
         let result = unsafe {
             ffi::AMediaCodec_dequeueInputBuffer(
                 self.as_ptr(),
@@ -304,19 +303,21 @@ impl MediaCodec {
         };
 
         if result == ffi::AMEDIACODEC_INFO_TRY_AGAIN_LATER as isize {
-            Ok(None)
+            Ok(DequeuedInputBufferResult::TryAgainLater)
         } else if result >= 0 {
-            Ok(Some(InputBuffer {
+            Ok(DequeuedInputBufferResult::Buffer(InputBuffer {
                 codec: self,
                 index: result as usize,
             }))
         } else {
-            MediaError::from_status(ffi::media_status_t(result as _)).map(|()| None)
+            Err(MediaError::from_status(ffi::media_status_t(result as _)).unwrap_err())
         }
     }
 
-    /// Returns [`None`] if timeout is reached.
-    pub fn dequeue_output_buffer(&self, timeout: Duration) -> Result<Option<OutputBuffer>> {
+    pub fn dequeue_output_buffer(
+        &self,
+        timeout: Duration,
+    ) -> Result<DequeuedOutputBufferInfoResult<'_>> {
         let mut info: ffi::AMediaCodecBufferInfo = unsafe { std::mem::zeroed() };
 
         let result = unsafe {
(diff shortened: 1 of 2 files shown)


ndk/src/media/media_codec.rs
```diff
@@ -510,3 +529,11 @@ impl OutputBuffer<'_> {
         self.info.presentationTimeUs
     }
 }
+
+#[derive(Debug)]
+pub enum DequeuedOutputBufferResult<'a> {
```
rib, line 534: I wonder if this should be named more like the underlying NDK type `DequeuedOutputBufferInfo`.

In the C api the same info is also used in a few other places, namely [AMediaMuxer_writeSampleData](https://developer.android.com/ndk/reference/group/media#group___media_1gab516a49c6cb40440efe33fa2237b7e1b) and https://developer.android.com/ndk/reference/group/media#amediacodeconasyncoutputavailable so maybe it's worth double checking if it also makes sense to have a general type in the Rust API too.

`Info` feels like it also helps distinguish it conceptually from the `Result()` that it's returned within.

rib: I suppose the issue with that suggestion though is that this version isn't really just 'info' it may also encapsulate a buffer.

MarijnS95: `Info` also suits my fancy, wasn't really a fan of `Result` already: https://github.com/rust-mobile/ndk/pull/316#issuecomment-1594340070

MarijnS95: > In the C api the same info is also used in a few other places, namely [AMediaMuxer_writeSampleData](https://developer.android.com/ndk/reference/group/media#group___media_1gab516a49c6cb40440efe33fa2237b7e1b) and https://developer.android.com/ndk/reference/group/media#amediacodeconasyncoutputavailable so maybe it's worth double checking if it also makes sense to have a general type in the Rust API too.

Fair point, guess we should make a separate PR to generalize this type so that `AMediaMuxer` can be implemented at some point. CC @zarik5.

MarijnS95: Or how about `*BufferInfoResult`?

rib: hehe, yeah always secret option number three to combine them :)

no strong preference from me really. Feels pretty verbose but yeah maybe BufferInfoResult helps connect to the underlying NDK API while also representing that it's not pure info in this case.

MarijnS95: Let's go for it, I think this was enough bikeshedding. Leaving `DequeuedInputBufferResult` as it is because there's no `Info` struct involved.

Generalizing `DequeuedOutputBufferInfoResult` is TBD, the inner `OutputBuffer` struct is quite specific to `MediaCodec`.

rib (approved):
did all the archaeology for https://github.com/rust-mobile/ndk/pull/316 and these issues https://github.com/android/ndk/discussions/1768 + https://issuetracker.google.com/u/1/issues/255698045?pli=1

fun :/

I haven't had to use these media APIs but yeah this all seems reasonable to me

rib:
Maybe also add @zarik5 as a reviewer considering they are really depending on this API

MarijnS95:
We've already bikeshedded this a bit in #316, I cannot assign non-collaborators for reviewers and on a prior occasion their comment was basically that they did not have time to maintain the code they contributed... so I'm not super interested in waiting for that opinion either.

zarik5:
I already discussed these changes in the other PR, and beside the exact naming scheme,this API will suit my needs.

zarik5:
I switched to the rust-mobile/ndk git master in production and it works good so far.

MarijnS95:
@zarik5 thanks for the positive feedback, glad to hear that we might release this soon :)

PR #402, opened 2023-06-16
ndk/media_codec: Simplify `dequeue_output_buffer()` with `MaybeUninit`

MarijnS95:
Depends on #401, CC @zarik5


rib (approved):
looks good to me

PR #404, opened 2023-06-16
ndk/native_window: Add `lock()` to blit raw pixel data

MarijnS95:
For https://github.com/rust-windowing/softbuffer/issues/44, CC @notgull

### DIFF
--- a/ndk/src/hardware_buffer_format.rs
+++ b/ndk/src/hardware_buffer_format.rs
@@ -16,6 +16,7 @@ pub enum HardwareBufferFormat {
     R8G8B8_UNORM = ffi::AHardwareBuffer_Format::AHARDWAREBUFFER_FORMAT_R8G8B8_UNORM.0,
     /// Matches deprecated [`ffi::ANativeWindow_LegacyFormat::WINDOW_FORMAT_RGB_565`].
     R5G6B5_UNORM = ffi::AHardwareBuffer_Format::AHARDWAREBUFFER_FORMAT_R5G6B5_UNORM.0,
+    #[cfg(feature = "api-level-26")]
     R16G16B16A16_FLOAT = ffi::AHardwareBuffer_Format::AHARDWAREBUFFER_FORMAT_R16G16B16A16_FLOAT.0,
     #[cfg(feature = "api-level-26")]
     R10G10B10A2_UNORM = ffi::AHardwareBuffer_Format::AHARDWAREBUFFER_FORMAT_R10G10B10A2_UNORM.0,
@@ -39,4 +40,41 @@ pub enum HardwareBufferFormat {
     YCbCr_P010 = ffi::AHardwareBuffer_Format::AHARDWAREBUFFER_FORMAT_YCbCr_P010.0,
     #[cfg(feature = "api-level-26")]
     R8_UNORM = ffi::AHardwareBuffer_Format::AHARDWAREBUFFER_FORMAT_R8_UNORM.0,
+}
+
+impl HardwareBufferFormat {
+    /// Returns [`None`] when there is no immediate byte size available for this format, for
+    /// example on planar buffer formats.
+    pub fn bytes_per_pixel(self) -> Option<usize> {
+        Some(match self {
+            Self::R8G8B8A8_UNORM | Self::R8G8B8X8_UNORM => 4,
+            #[cfg(feature = "api-level-26")]
+            Self::R8G8B8_UNORM => 3,
+            Self::R5G6B5_UNORM => 2,
+            #[cfg(feature = "api-level-26")]
+            Self::R16G16B16A16_FLOAT => 8,
+            #[cfg(feature = "api-level-26")]
+            Self::R10G10B10A2_UNORM => 4,
+            #[cfg(feature = "api-level-26")]
+            Self::BLOB => 1,
+            #[cfg(feature = "api-level-26")]
+            Self::D16_UNORM => 2,
+            #[cfg(feature = "api-level-26")]
+            Self::D24_UNORM => 3,
+            #[cfg(feature = "api-level-26")]
+            Self::D24_UNORM_S8_UINT => 4,
+            #[cfg(feature = "api-level-26")]
+            Self::D32_FLOAT => 4,
+            #[cfg(feature = "api-level-26")]
+            Self::D32_FLOAT_S8_UINT => 5,
+            #[cfg(feature = "api-level-26")]
+            Self::S8_UINT => 1,
+            #[cfg(feature = "api-level-26")]
+            Self::Y8Cb8Cr8_420 => 3,
+            #[cfg(feature = "api-level-26")]
+            Self::YCbCr_P010 => return None,
+            #[cfg(feature = "api-level-26")]
+            Self::R8_UNORM => 1,
+        })
+    }
 }
--- a/ndk/src/native_window.rs
+++ b/ndk/src/native_window.rs
@@ -7,7 +7,9 @@ use crate::utils::status_to_io_result;
 pub use super::hardware_buffer_format::HardwareBufferFormat;
 use jni_sys::{jobject, JNIEnv};
 use raw_window_handle::{AndroidNdkWindowHandle, HasRawWindowHandle, RawWindowHandle};
-use std::{convert::TryFrom, ffi::c_void, io::Result, ptr::NonNull};
+use std::{convert::TryFrom, ffi::c_void, io::Result, mem::MaybeUninit, ptr::NonNull};
+
+pub type Rect = ffi::ARect;
 
 // [`NativeWindow`] represents the producer end of an image queue
 ///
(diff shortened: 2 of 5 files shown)


rib:
Hehe, did you maybe also see me using `ANativeWindow_lock` in my recent rust-android-examples changes per chance?

https://github.com/rust-mobile/rust-android-examples/pull/4/files#diff-80528ea769093da879768e086770c9b11f81b2cd8aa51b8218f5183772eb8b18R93

Yeah, in particular it was too much of a faff in that context to also deal with getting the bytes per-pixel which you really need to do anything with the buffers from the CPU since the API is pretty ghastly and doesn't tell you the size of the buffer in bytes - you have to work it out based on the pixel format.

Blob doesn't need to be `todo!()` it is a byte format with the extra constraint that the buffer will have a height of 1 so the width can be read as a size in bytes.

MarijnS95:
No I did not, this is actually a commit from almost 1 year ago (https://github.com/rust-mobile/ndk/commit/d104d01f116b537babdc94d8c9889581f8d63a02) that I just kept postponing to contribute back, because I didn't know what to do with the `pack_rgba()` (see that commit) helper and ended up scrapping that.

MarijnS95:
Regarding `BLOB`, just 1 byte per pixel, that's cool.

Also bpp is generally in bits, shall I change this to return bits instead of bytes?

rib:
bpp is just ambiguous - definitely wouldn't assume either way without looking for context.

bytes seems fine here.

There are some missing format that could be good to fill in for completeness here.

e.g. AHARDWAREBUFFER_FORMAT_R10G10B10A10_UNORM = VK_FORMAT_R10X6G10X6B10X6A10X6_UNORM_4PACK16 

which is 8 bytes per pixel:
> specifies a four-component, 64-bit unsigned normalized format that has a 10-bit R component in the top 10 bits of the word in bytes 0..1, a 10-bit G component in the top 10 bits of the word in bytes 2..3, a 10-bit B component in the top 10 bits of the word in bytes 4..5, and a 10-bit A component in the top 10 bits of the word in bytes 6..7, with the bottom 6 bits of each word unused.
(from https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkFormat.html)

rib:
> No I did not, this is actually a commit from almost 1 year ago ([d104d01](https://github.com/rust-mobile/ndk/commit/d104d01f116b537babdc94d8c9889581f8d63a02)) that I just kept postponing to contribute back, because I didn't know what to do with the `pack_rgba()` (see that commit) helper and ended up scrapping that.

nice coincidence then

I wanted to be able to clear the framebuffer at least when doing that but also didn't want to be inlining a bunch of byte-per-pixel format mappings just for that.

MarijnS95:
`AHARDWAREBUFFER_FORMAT_R10G10B10A10_UNORM` doesn't seem to be in NDK 25.2, but there are a few others that we should definitely add.

rib:
> `AHARDWAREBUFFER_FORMAT_R10G10B10A10_UNORM` doesn't seem to be in NDK 25.2, but there are a few others that we should definitely add.

oh, curious.

I was looking here https://developer.android.com/ndk/reference/group/a-hardware-buffer#ahardwarebuffer_format

MarijnS95:
Docs are typically ahead, guess they come with NDK r26. There's no Android version compatibility listed in the docs, though, so I assume 26 because that's when the non-legacy formats were added.

rib:
> because I didn't know what to do with the pack_rgba() (see that commit) helper and ended up scrapping that.

right that's probably the appropriate thing here. that might be a handy utility for testing but probably doesn't belong as part of the `ndk` crate.

ndk/src/native_window.rs
```diff
@@ -126,4 +126,61 @@ impl NativeWindow {
...
+    /// The number of _pixels_ that a line in the buffer takes in memory.
+    ///
+    /// This may be `>= width`.
+    pub fn stride(&self) -> i32 {
+        self.0.stride
+    }
+
+    /// The format of the buffer. One of [`HardwareBufferFormat`].
+    pub fn format(&self) -> HardwareBufferFormat {
+        unsafe { std::mem::transmute(self.0.format) }
```
rib, line 172: might get undefined behaviour here if the implementation isn't updated before new formats are potentially exposed on Android.

Could potentially use TryFromPrimitive from the num_enum crate or implement something equivalent manually that will ensure the u32 from the NDK is a known Rust enum variant.

rib: oh, yeah, you're already using `num_enum` here and have `TryFromPrimitive`

ndk/src/native_window.rs
```diff
@@ -126,4 +126,61 @@ impl NativeWindow {
...
+    }
+}
+
+/// Lock holding the next drawing surface for writing.  It is unlocked and posted on [`drop()`].
+#[derive(Debug)]
+pub struct NativeWindowBufferLockGuard<'a>(ffi::ANativeWindow_Buffer, &'a NativeWindow);
+
+impl<'a> NativeWindowBufferLockGuard<'a> {
+    /// The number of pixels that are shown horizontally.
+    pub fn width(&self) -> i32 {
```
rib, line 154: seems like width, height and stride could be exposed as `usize` in Rust

ndk/src/native_window.rs
MarijnS95: Yeah indeed, that's sloppy. Shall I adopt the `.unwrap()` or return `Result<, TryFromPrimitive>` for both?

ndk/src/native_window.rs
MarijnS95: Indeed, that makes it easier to work with in loops and `ops::Idex`.

ndk/src/native_window.rs
rib: yeah tricky. we already know there are r26 formats that aren't added here so the possibility of unknown formats seems worth considering.

in practice I'd guess only a small number of common formats ever show up as formats you'd see for rendering to a window from the cpu though - most of these formats probably only really show up in connection with GLES/vulkan.

ndk/src/hardware_buffer_format.rs
```diff
@@ -35,4 +36,43 @@ pub enum HardwareBufferFormat {
...
+            #[cfg(feature = "api-level-26")]
+            Self::D32_FLOAT => 4,
+            #[cfg(feature = "api-level-26")]
+            Self::D32_FLOAT_S8_UINT => 5,
+            #[cfg(feature = "api-level-26")]
+            Self::S8_UINT => 1,
+            #[cfg(feature = "api-level-26")]
+            Self::Y8Cb8Cr8_420 => 3,
+            #[cfg(feature = "api-level-26")]
+            Self::YCbCr_P010 => unimplemented!("Semiplanar"),
```
rib, line 73: ah, right, I was wondering if there were going to be planar video formats to handle

panicking for planar video formats is a bit harsh - maybe this should return an Option though because there's no practical bytes per pixel for video formats with different sized planes.

In practice as long as we cover the common rgbx formats that should be all that's required for helping enable apps to render from the CPU, whereas video formats will show up if your pulling frames from a camera and more context-aware code won't be trying to ask for a bytes_per_pixel for those formats anyway.

ndk/src/native_window.rs
```diff
@@ -126,4 +126,62 @@ impl NativeWindow {
...
         ffi::ANativeWindow_toSurface(env, self.ptr().as_ptr())
     }
+
+    /// Lock the window's next drawing surface for writing.
+    ///
+    /// Optionally pass the region you intend to draw into `dirty_bounds`.  When this function
+    /// returns it is updated (commonly enlarged) with the actual area the caller needs to redraw.
+    pub fn lock(
+        &self,
+        dirty_bounds: Option<&mut ffi::ARect>,
```
notgull, line 136: I would avoid having a mutable reference like this if all possible. I would pass in a rectangle and then return a rectangle as a tuple with the lock.

ndk/src/native_window.rs
```diff
@@ -126,4 +126,62 @@ impl NativeWindow {
...
+    }
+
+    /// The format of the buffer. One of [`HardwareBufferFormat`].
+    pub fn format(&self) -> HardwareBufferFormat {
+        let format = u32::try_from(self.0.format).unwrap();
+        HardwareBufferFormat::try_from(format).unwrap()
+    }
+
+    /// The actual bits.
+    pub fn bits(&mut self) -> *mut std::ffi::c_void {
```
notgull, line 177: Maybe it would be possible to have a function that returns a mutable slice instead of a pointer?

notgull (approved):
I'm not familiar enough with Android to comment, but overall this looks good to me.

ndk/src/native_window.rs
MarijnS95: I had considered that, and especially in the context of `softbuffer` it is UB to cast a pointer to a safe slice, since these APIs are mostly for writing to memory.

It'll have to be `&[MaybeUninit<u8>]`.

ndk/src/native_window.rs
MarijnS95: I instead considered putting it in the returned guard.

ndk/src/hardware_buffer_format.rs
MarijnS95: Returning an option now and documented this.

ndk/src/native_window.rs
MarijnS95: Done that now, but it might make it less clear that when the user passes this to the function, they have to read back _the updated value_ from `NativeWindowBufferLockGuard`. But that would be the same when they already define their local variable `mut` for other reasons.

ndk/src/native_window.rs
MarijnS95: @notgull done this now. Perhaps it is also interesting to add API's that give access to individual scanlines, by offsetting the pointer by `stride() * num_rows *bpp` and giving the user only a slice of length `width() * bpp`? And then wrap that in an iterator again so that it becomes very natural to update pixels?

That won't help for `softbuffer` though which currently assumes contiguous memory.

notgull (approved):
Looks good to me!

rib (approved):
I'm not entirely convinced that having the separate API for getting the modified dirty bounds from the lock ends up being better than the `&mut` argument.

My nagging concern would probably be that it's now easier to miss the fact that if you do pass a dirty bounds you _need_ to additionally remember to query back the modified bounds - whereas with the `&mut` argument, even if a developer hadn't fully appreciated that the region might get expanded they will any way probably reference the correct region in their code.

Either way is workable though and this is a fairly low-level detail so this seems reasonable enough to me.

Apart from that, everything else looks good to me.

MarijnS95:
Thanks. @rib I've reverted that suggestion from @notgull now as I indeed agree that when a user is forced to pass `Some(&mut dirty_bounds)` over `Some(dirty_bounds)` this hopefully nags/pushes them towards reading (or understanding from the get-go) that their bounds will be updated.

It is only a problem when someone requests explicit bounds anyway, most might just pass `None` and touch the whole window.

I've also added the per-line iterator API hinted at above and tested it out in:

https://github.com/rust-mobile/rust-android-examples/compare/na-pixels-lock

Rust's `chunks_mut` and subslicing does make it rather trivial to implement, without messing around with pointer offsets :)

PR #407, opened 2023-07-05
hardware_buffer_format: Add catch-all variant

spencercw:
Hardware buffers can have private formats which don't appear in the NDK header. Add a catch-all variant to avoid panicking in `HardwareBuffer::describe()`.

### DIFF
--- a/ndk/CHANGELOG.md
+++ b/ndk/CHANGELOG.md
@@ -13,6 +13,7 @@
 - **Breaking:** media_codec: Return `MaybeUninit` bytes in `buffer_mut()`. (#403)
 - native_window: Add `lock()` to blit raw pixel data. (#404)
 - hardware_buffer_format: Add `YCbCr_P010` and `R8_UNORM` variants. (#405)
+- **Breaking:** hardware_buffer_format: Add catch-all variant. (#407)
 
 # 0.7.0 (2022-07-24)
 
--- a/ndk/src/native_window.rs
+++ b/ndk/src/native_window.rs
@@ -81,7 +81,7 @@ impl NativeWindow {
     pub fn format(&self) -> HardwareBufferFormat {
         let value = unsafe { ffi::ANativeWindow_getFormat(self.ptr.as_ptr()) };
         let value = u32::try_from(value).unwrap();
-        HardwareBufferFormat::try_from(value).unwrap()
+        HardwareBufferFormat::from(ffi::AHardwareBuffer_Format(value))
     }
 
     /// Change the format and size of the window buffers.
@@ -99,7 +99,7 @@ impl NativeWindow {
         height: i32,
         format: Option<HardwareBufferFormat>,
     ) -> Result<()> {
-        let format: u32 = format.map_or(0, |f| f.into());
+        let format = format.map_or(0, |f| ffi::AHardwareBuffer_Format::from(f).0);
         let status = unsafe {
             ffi::ANativeWindow_setBuffersGeometry(self.ptr.as_ptr(), width, height, format as i32)
         };
@@ -179,7 +179,7 @@ impl<'a> NativeWindowBufferLockGuard<'a> {
     /// The format of the buffer. One of [`HardwareBufferFormat`].
     pub fn format(&self) -> HardwareBufferFormat {
         let format = u32::try_from(self.buffer.format).unwrap();
-        HardwareBufferFormat::try_from(format).unwrap()
+        HardwareBufferFormat::from(ffi::AHardwareBuffer_Format(format))
     }
 
     /// The actual bits.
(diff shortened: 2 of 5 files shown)


spencercw:
The MSRV issue is a bit unfortunate. Seems explicit discriminants on enums with fields is only supported as of Rust 1.66 (Dec 2022). Is it ok to bump the MSRV?

MarijnS95:
I'd rather panic and extend the crate in this case. Is this a real problem right now or are you anticipating a future problem?

spencercw:
This is a real problem right now. I am creating an `ImageReader` with `ImageFormat::PRIVATE` and `           HardwareBufferUsage::GPU_SAMPLED_IMAGE`, which I then feed in with decoded video frames from `MediaCodec`. I then acquire images from the `ImageReader` and get the `HardwareBuffer` from those images. The `format` returned by calling `describe` on the `HardwareBuffer` varies depending on the device. For example, on a Samsung device the format is `261`. On a Xiaomi device it is `2141391878`. I'm not really interested in the format here, but I need the other fields in `HardwareBufferDesc`, which I can't currently get because it panics trying to handle the format.

MarijnS95 (approved):

MarijnS95:
Sure, then we need an `Other` (rename to `Unknown`?) case as we also cannot return a `Result` to support your "I want `HardwareBufferDesc` but don't care about the specific format" case.

I'll think about the MSRV bump for a bit: this'll be used by `winit` and we have to keep them into account as the upper bound of our MSRV. It's still as `1.64` and I don't know if there are plans to bump it to `>= 1.66`.

MarijnS95:
> Seems explicit discriminants on enums with fields

Note also that this PR makes it impossible to use `as u32` on the type. Perhaps we should just drop `#[repr(u32)]` then and not care about the size of the discriminant, as `num_enum` should generate this for us with `IntoPrimitive`. We might have to document that more clearly, though.

MarijnS95:
> `#[repr(u32)]` seems to be required by `num-enum`. Removing it elicits a 'Missing `#[repr({Integer})]` attribute' error.

Yeah, because it needs these discriminants to generate a `match` table for conversion... Perhaps it could even discard the discriminants after it's done, so that the type relies on default Rust numbering and `From/IntoPrimitive` convert from and to NDK numbering. That'd get rid of the MSRV issue.

(Otherwise the implementation for `IntoPrimitive` is nothing more as `self as u32`... yet it works even when one or more enum variants have fields)

spencercw:
Sounds like a reasonable idea for `num-enum`, but it's not something I particularly have time to get into at the moment. One day I might learn how to use procedural macros, but right now they are just black magic to me.

MarijnS95:
It's more so that I don't have a clear overview of what's needed now, but we might as well ditch the macros and implement `From` conversions manually (both ways) without setting `repr(u32)` nor the discriminants. This is done for other `enum`s in the ndk crate as well.

spencercw:
That could work, although unless I'm missing something obvious, I can't see any custom `From`/`Into` implementations in the repository.

spencercw:
Added manual `From` implementations.

MarijnS95:
> I can't see any custom `From`/`Into` implementations in the repository.

Turns out it was a custom function instead (otherwise it'd have to be a `TryFrom`/`TryInto`), but check out `MediaStatus`/`MediaError`.

Note that `AHardwareBuffer_Format`, just like the `media_status_t` demonstrated there, _derives_ `PartialEq, Eq` which allows you to match on it _without `x if x == ffi::Type::VAL.0`_ as long as it is wrapped in `AHardwareBuffer_Format`. Or perhaps better: implement the `From` directly for `AHardwareBuffer_Format` so that it is clear what the native type is "supposed to be".

spencercw:
Ok that makes sense. I've applied those changes.

ndk/src/native_window.rs
```diff
@@ -99,7 +99,8 @@ impl NativeWindow {
         height: i32,
         format: Option<HardwareBufferFormat>,
     ) -> Result<()> {
-        let format: u32 = format.map_or(0, |f| f.into());
+        let format: Option<ffi::AHardwareBuffer_Format> = format.map(|f| f.into());
+        let format: u32 = format.map_or(0, |f| f.0);
```
MarijnS95, line 103: Can we just write this as:

```suggestion
        let format = format.map_or(0, |f| ffi::AHardwareBuffer_Format::from(f).0);
```

Without additional explicit types etc?

MarijnS95: :+1:. Perhaps a similar thing (minus the `map_or()`) in `into_native()`?

ndk/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - **Breaking:** media_codec: Return `MaybeUninit` bytes in `buffer_mut()`. (#403)
 - native_window: Add `lock()` to blit raw pixel data. (#404)
 - hardware_buffer_format: Add `YCbCr_P010` and `R8_UNORM` variants. (#405)
+- hardware_buffer_format: Add catch-all variant. (#407)
```
MarijnS95, line 16: We should probably have a `**Breaking:**` suffix here?

MarijnS95 (approved):

rib (approved):

Issue #268, opened 2022-05-01
aaudio callbacks are unsound


nicbn:
https://github.com/rust-windowing/android-ndk-rs/blob/f24606cc840af2f97f0daebe1e2e7cd3aca52287/ndk/src/aaudio.rs#L545-L565

If `callback` panics and unwinds, `ffi_callback` unwinds, which is [undefined behaviour](https://doc.rust-lang.org/nomicon/unwinding.html).

> Rust's unwinding strategy is not specified to be fundamentally compatible with any other language's unwinding. As such, unwinding into Rust from another language, or unwinding into another language from Rust is Undefined Behavior. You must absolutely catch any panics at the FFI boundary! What you do at that point is up to you, but something must be done. If you fail to do this, at best, your application will crash and burn. At worst, your application won't crash and burn, and will proceed with completely clobbered state.

MarijnS95:
Feel free to submit a PR to address this.

CC @JasperDeSutter.

PR #410, opened 2023-07-31
media_codec: Add support for asynchronous notification callbacks

spencercw:
Note that unlike the audio data callback, I do not provide a reference to the codec as a callback parameter. Happy to be overruled on this, but I find the notion of constructing a new `MediaCodec` which wraps an `ffi::AMediaCodec` already owned by another `MediaCodec` somewhere else to be a bit questionable. If you really need access to the `MediaCodec` within the callback than you can use a `Mutex` or something, but the reality is that you can't actually do very much within the callback. For example, if you try to release the output buffer in the 'output available' callback then the whole thing just deadlocks.

Typical usage might look something like this (sans proper error handling):
```rust
let (input_tx, input_rx) = channel(32);
let (output_tx, output_rx) = channel(32);
codec
    .set_async_notify_callback(Some(AsyncNotifyCallback {
        on_input_available: Some(Box::new(move |index| {
            input_tx.blocking_send(index).unwrap();
        })),
        on_output_available: Some(Box::new(move |index, _info| {
            output_tx.blocking_send(index).unwrap();
        })),
        on_format_changed: None,
        on_error: Some(Box::new(move |error, action_code, detail| panic!()))
    })).unwrap();
```

Elsewhere...
```rust
select! {
    buffer_index = input_rx.recv() => {
        let buffer_index = buffer_index.unwrap();
        let input_buffer = codec.input_buffer(buffer_index).unwrap();
        // Fill buffer...
        codec.queue_input_buffer_by_index(buffer_index, 0, buffer_len, 0, 0).unwrap();
    }
    buffer_index = output_rx.recv() => {
        codec.release_output_buffer_by_index(buffer_index.unwrap(), true).unwrap();
    }
}
```

### DIFF
--- a/ndk/CHANGELOG.md
+++ b/ndk/CHANGELOG.md
@@ -15,6 +15,7 @@
 - native_window: Add `lock()` to blit raw pixel data. (#404)
 - hardware_buffer_format: Add `YCbCr_P010` and `R8_UNORM` variants. (#405)
 - **Breaking:** hardware_buffer_format: Add catch-all variant. (#407)
+- **Breaking:** media_codec: Add support for asynchronous notification callbacks. (#410)
 - Add panic guards to callbacks. (#412)
 - looper: Add `remove_fd()` to unregister events/callbacks for a file descriptor. (#416)
 
--- a/ndk/src/media/media_codec.rs
+++ b/ndk/src/media/media_codec.rs
@@ -5,10 +5,12 @@
 
 use crate::media_error::{MediaError, Result};
 use crate::native_window::NativeWindow;
+use crate::utils::abort_on_panic;
 use std::{
-    ffi::{CStr, CString},
+    ffi::{c_char, c_void, CStr, CString},
     fmt::Display,
     mem::MaybeUninit,
+    pin::Pin,
     ptr::{self, NonNull},
     slice,
     time::Duration,
@@ -42,6 +44,14 @@ impl Default for MediaFormat {
 }
 
 impl MediaFormat {
+    /// Assumes ownership of `ptr`
+    ///
+    /// # Safety
+    /// `ptr` must be a valid pointer to an Android [`ffi::AMediaFormat`].
+    pub unsafe fn from_ptr(ptr: NonNull<ffi::AMediaFormat>) -> Self {
+        Self { inner: ptr }
+    }
+
     fn as_ptr(&self) -> *mut ffi::AMediaFormat {
         self.inner.as_ptr()
     }
@@ -216,7 +226,75 @@ impl Drop for MediaFormat {
 #[derive(Debug)]
 pub struct MediaCodec {
     inner: NonNull<ffi::AMediaCodec>,
+    async_notify_callback: Option<Pin<Box<AsyncNotifyCallback>>>,
 }
+
+pub struct AsyncNotifyCallback {
+    /// Called when an input buffer becomes available.
+    ///
+    /// The specified index is the index of the available input buffer.
+    pub on_input_available: Option<InputAvailableCallback>,
+
+    /// Called when an output buffer becomes available.
+    ///
+    /// The specified index is the index of the available output buffer. The specified
+    /// [`BufferInfo`] contains information regarding the available output buffer.
+    pub on_output_available: Option<OutputAvailableCallback>,
+
+    /// Called when the output format has changed.
+    ///
+    /// The specified format contains the new output format.
+    pub on_format_changed: Option<FormatChangedCallback>,
+
+    /// Called when the [`MediaCodec`] encountered an error.
+    ///
+    /// The specified [`ActionCode`] indicates the possible actions that client can take, and it can
+    /// be checked by calling [`ActionCode::is_recoverable`] or [`ActionCode::is_transient`]. If
+    /// both [`ActionCode::is_recoverable`] and [`ActionCode::is_transient`] return [`false`], then
+    /// the codec error is fatal and the codec must be deleted. The specified detail string may
+    /// contain more detailed messages about this error.
+    pub on_error: Option<ErrorCallback>,
+}
+
+impl std::fmt::Debug for AsyncNotifyCallback {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("AsyncNotifyCallback")
+            .field(
+                "on_input_available",
+                match &self.on_input_available {
+                    Some(_) => &"Some(_)",
+                    None => &"None",
+                },
+            )
+            .field(
+                "on_output_available",
+                match &self.on_output_available {
+                    Some(_) => &"Some(_)",
+                    None => &"None",
+                },
+            )
+            .field(
+                "on_format_changed",
+                match &self.on_format_changed {
+                    Some(_) => &"Some(_)",
+                    None => &"None",
+                },
+            )
+            .field(
+                "on_error",
+                match &self.on_error {
+                    Some(_) => &"Some(_)",
+                    None => &"None",
+                },
+            )
+            .finish()
+    }
+}
+
+pub type InputAvailableCallback = Box<dyn FnMut(usize)>;
+pub type OutputAvailableCallback = Box<dyn FnMut(usize, &BufferInfo)>;
+pub type FormatChangedCallback = Box<dyn FnMut(&MediaFormat)>;
+pub type ErrorCallback = Box<dyn FnMut(MediaError, ActionCode, &CStr)>;
 
 impl MediaCodec {
     fn as_ptr(&self) -> *mut ffi::AMediaCodec {
@@ -227,6 +305,7 @@ impl MediaCodec {
         let c_string = CString::new(name).unwrap();
         Some(Self {
             inner: NonNull::new(unsafe { ffi::AMediaCodec_createCodecByName(c_string.as_ptr()) })?,
+            async_notify_callback: None,
         })
     }
 
@@ -236,6 +315,7 @@ impl MediaCodec {
             inner: NonNull::new(unsafe {
                 ffi::AMediaCodec_createDecoderByType(c_string.as_ptr())
             })?,
+            async_notify_callback: None,
         })
     }
 
@@ -245,7 +325,148 @@ impl MediaCodec {
             inner: NonNull::new(unsafe {
                 ffi::AMediaCodec_createEncoderByType(c_string.as_ptr())
             })?,
+            async_notify_callback: None,
         })
+    }
+
+    /// Set an asynchronous callback for actionable [`MediaCodec`] events.
+    ///
+    /// When asynchronous callback is enabled, it is an error for the client to call
+    /// [`MediaCodec::dequeue_input_buffer()`] or [`MediaCodec::dequeue_output_buffer()`].
+    ///
+    /// [`MediaCodec::flush()`] behaves differently in asynchronous mode. After calling
+    /// [`MediaCodec::flush()`], the client must call [`MediaCodec::start()`] to "resume" receiving
+    /// input buffers. Even if the client does not receive
+    /// [`AsyncNotifyCallback::on_input_available`] callbacks from video encoders configured with an
+    /// input surface, the client still needs to call [`MediaCodec::start()`] to resume the input
+    /// surface to send buffers to the encoders.
+    ///
+    /// When called with [`None`] callback, this method unregisters any previously set callback.
+    ///
+    /// Refer to the definition of [`AsyncNotifyCallback`] on how each callback function is called
+    /// and what are specified.
+    ///
+    /// Once the callback is unregistered or the codec is reset / released, the previously
+    /// registered callback will not be called.
+    ///
+    /// All callbacks are fired on one NDK internal thread.
+    /// [`MediaCodec::set_async_notify_callback()`] should not be called on the callback thread. No
+    /// heavy duty task should be performed on callback thread.
+    #[cfg(feature = "api-level-28")]
+    pub fn set_async_notify_callback(
+        &mut self,
+        callback: Option<AsyncNotifyCallback>,
+    ) -> Result<()> {
+        unsafe extern "C" fn ffi_on_input_available(
+            _codec: *mut ffi::AMediaCodec,
+            user_data: *mut c_void,
+            index: i32,
+        ) {
+            abort_on_panic(|| {
+                let callback = &mut *(user_data as *mut AsyncNotifyCallback);
+                if let Some(f) = callback.on_input_available.as_mut() {
+                    f(index as usize);
+                }
+            })
+        }
+
+        unsafe extern "C" fn ffi_on_output_available(
+            _codec: *mut ffi::AMediaCodec,
+            user_data: *mut c_void,
+            index: i32,
+            buffer_info: *mut ffi::AMediaCodecBufferInfo,
+        ) {
+            abort_on_panic(|| {
+                let callback = &mut *(user_data as *mut AsyncNotifyCallback);
+                if let Some(f) = callback.on_output_available.as_mut() {
+                    let buffer_info = BufferInfo {
+                        inner: *buffer_info,
+                    };
+                    f(index as usize, &buffer_info);
+                }
+            })
+        }
+
+        unsafe extern "C" fn ffi_on_format_changed(
+            _codec: *mut ffi::AMediaCodec,
+            user_data: *mut c_void,
+            format: *mut ffi::AMediaFormat,
+        ) {
+            abort_on_panic(|| {
+                // Ownership of the format is not documented, but the implementation allocates a new instance and does
+                // not free it, so assume it is ok for us to do so
+                // https://cs.android.com/android/platform/superproject/main/+/refs/heads/main:frameworks/av/media/ndk/NdkMediaCodec.cpp;l=248-254;drc=5e15c3e22f3fa32d64e57302201123ce41589adf
+                let format = MediaFormat::from_ptr(NonNull::new_unchecked(format));
+
+                let callback = &mut *(user_data as *mut AsyncNotifyCallback);
+                if let Some(f) = callback.on_format_changed.as_mut() {
+                    f(&format);
+                }
+            })
+        }
+
+        unsafe extern "C" fn ffi_on_error(
+            _codec: *mut ffi::AMediaCodec,
+            user_data: *mut c_void,
+            error: ffi::media_status_t,
+            action_code: i32,
+            detail: *const c_char,
+        ) {
+            abort_on_panic(|| {
+                let callback = &mut *(user_data as *mut AsyncNotifyCallback);
+                if let Some(f) = callback.on_error.as_mut() {
+                    f(
+                        MediaError::from_status(error).unwrap_err(),
+                        ActionCode(action_code),
+                        CStr::from_ptr(detail),
+                    );
+                }
+            })
+        }
+
+        let (callback, ffi_callback, user_data) = if let Some(callback) = callback {
+            // On Android 12 and earlier, due to faulty null checks, if a callback is not set, but at least one other
+            // callback *is* set, then it will segfault in when trying to invoke the unset callback. See for example:
+            // https://cs.android.com/android/platform/superproject/+/android-12.0.0_r34:frameworks/av/media/ndk/NdkMediaCodec.cpp;l=161-162;drc=ef058464777739e2d9ffad5f00d0e57b186d9a13
+            // To work around this we just enable all callbacks and do nothing if the corresponding callback is not set
+            // in AsyncNotifyCallback
+            let ffi_callback = ffi::AMediaCodecOnAsyncNotifyCallback {
+                onAsyncInputAvailable: Some(ffi_on_input_available),
+                onAsyncOutputAvailable: Some(ffi_on_output_available),
+                onAsyncFormatChanged: Some(ffi_on_format_changed),
+                onAsyncError: Some(ffi_on_error),
+            };
+
+            let mut boxed = Box::pin(callback);
+            let ptr: *mut AsyncNotifyCallback = &mut *boxed;
+
+            (Some(boxed), ffi_callback, ptr as *mut c_void)
+        } else {
+            let ffi_callback = ffi::AMediaCodecOnAsyncNotifyCallback {
+                onAsyncInputAvailable: None,
+                onAsyncOutputAvailable: None,
+                onAsyncFormatChanged: None,
+                onAsyncError: None,
+            };
+
+            (None, ffi_callback, ptr::null_mut())
+        };
+
+        let status = unsafe {
+            ffi::AMediaCodec_setAsyncNotifyCallback(self.as_ptr(), ffi_callback, user_data)
+        };
+        let result = MediaError::from_status(status);
+
+        // This behavior is not documented, but the implementation always clears the callback on failure, so we must
+        // clear any callback that may have been previously registered
+        // https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/media/ndk/NdkMediaCodec.cpp;l=581-584;drc=8c4e619c7461ac1a8c20c55364643662e9185e4d
+        if result.is_ok() {
+            self.async_notify_callback = callback;
+        } else {
+            self.async_notify_callback = None;
+        }
+
+        result
     }
 
     pub fn configure(
@@ -341,7 +562,9 @@ impl MediaCodec {
             Ok(DequeuedOutputBufferInfoResult::Buffer(OutputBuffer {
                 codec: self,
                 index: result as usize,
-                info: unsafe { info.assume_init() },
+                info: BufferInfo {
+                    inner: unsafe { info.assume_init() },
+                },
             }))
         } else {
             Err(MediaError::from_status(ffi::media_status_t(result as _)).unwrap_err())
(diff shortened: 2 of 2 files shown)


MarijnS95:
> Happy to be overruled on this, but I find the notion of constructing a new `MediaCodec` which wraps an `ffi::AMediaCodec` already owned by another `MediaCodec` somewhere else to be a bit questionable.

It is not only questionable, this should break on `Drop` _unless `AMediaCodec` is internally refcounted **and we can increase the refcount** like we do for other Android objects in a custom `Clone` implementation_.

Perhaps you can work something out to give a `&MediaCodec` back? Though this does rely on some weak/strong handle to a `Mutex`/`Arc` indeed, so that the callback can't technically get to a `MediaCodec` if it happens to be fired after the object has already been destroyed and deallocated. On the other hand Android might define this API to not call any callbacks after it was destroyed, meaning some "working around the borrow checker" might've been adequate.

In the past I've been wanting to have some `unsafe fn borrrow_from_ptr<'a>(x: NonNull<...>) -> &'a Self` but we can't really borrow it "from somewhere" nor cast a pointer to a _borrow of_ said pointer.

ndk/src/media/media_codec.rs
```diff
@@ -246,9 +313,133 @@ impl MediaCodec {
...
+    pub fn set_async_notify_callback(
+        &mut self,
+        callback: Option<AsyncNotifyCallback>,
+    ) -> Result<()> {
+        unsafe extern "C" fn ffi_on_input_available(
+            _codec: *mut ffi::AMediaCodec,
+            user_data: *mut c_void,
+            index: i32,
+        ) {
+            match panic::catch_unwind(|| {
```
MarijnS95, line 330: Didn't we create a wrapper for this at some point, or was that on `android-activity`?

ndk/src/media/media_codec.rs
```diff
@@ -354,6 +545,28 @@ impl MediaCodec {
         MediaError::from_status(status)
     }
 
+    pub fn input_buffer(&self, index: usize) -> Option<&mut [u8]> {
```
MarijnS95, line 548: Can `InputBuffer::buffer_mut()` now simply call `self.codec.input_buffer(self.index)`? Would like to deduplicate things as much as possible.

Same for `OutputBuffer::buffer()` calling `self.codec.output_buffer(self.index)`.

ndk/src/media/media_codec.rs
```diff
@@ -538,3 +774,39 @@ pub enum DequeuedOutputBufferInfoResult<'a> {
     OutputFormatChanged,
     OutputBuffersChanged,
 }
+
+#[derive(Copy, Clone, Debug)]
+pub struct BufferInfo {
+    inner: ffi::AMediaCodecBufferInfo,
+}
```
MarijnS95, line 781: Should `OutputBuffer` contain (and `Deref` to) this so that it also has consistent access to all fields?

ndk/src/media/media_codec.rs
MarijnS95: It at least also printed the panic reason, which one can typically downcast to a `String` or `&'static str`.

MarijnS95:
@zarik5 you might want to see this!

ndk/src/media/media_codec.rs
```diff
@@ -246,9 +313,133 @@ impl MediaCodec {
...
+            // in AsyncNotifyCallback
+            let ffi_callback = ffi::AMediaCodecOnAsyncNotifyCallback {
+                onAsyncInputAvailable: Some(ffi_on_input_available),
+                onAsyncOutputAvailable: Some(ffi_on_output_available),
+                onAsyncFormatChanged: Some(ffi_on_format_changed),
+                onAsyncError: Some(ffi_on_error),
+            };
+
+            let mut boxed = Box::new(callback);
+            let ptr: *mut AsyncNotifyCallback = &mut *boxed;
```
MarijnS95, line 415: You'd typically want to wrap the `Box` (or other pointer-types) in `Pin` to prevent the allocation from accidentally moving around?

MarijnS95:
Let's not have the PR link in the commit title (as if this was a final merge-commit) please :)

spencercw:
> On the other hand Android might define this API to not call any callbacks after it was destroyed, meaning some "working around the borrow checker" might've been adequate.

Looking at the Android source code, when an `AMediaCodec` is deleted, it [stops the looper](https://cs.android.com/android/platform/superproject/main/+/refs/heads/main:frameworks/av/media/ndk/NdkMediaCodec.cpp;l=499;drc=e5dbe3b9c2669b5394133391c074b94535694c95) used for asynchronous callbacks, and this appears to [block while waiting for the looper thread to exit](https://cs.android.com/android/platform/superproject/main/+/refs/heads/main:frameworks/av/media/module/foundation/ALooper.cpp;l=166;drc=e5dbe3b9c2669b5394133391c074b94535694c95). Naturally this is not documented, but I don't think I would be too bothered about just assuming the `AMediaCodec` is still valid when we get called. It gives us an `AMediaCodec` pointer after all, so it would be a bit weird for that pointer to be potentially invalid.

Maybe we can pass a reference to the callback, but.. why? Even something as simple as `AMediaCodec_getName` deadlocks when invoked from a callback. As far as I can tell there is nothing useful you can do with the codec handle. It's just a footgun, luring you into doing something which is going to break your program.

> Let's not have the PR link in the commit title (as if this was a final merge-commit) please :)

I was just copying the convention from previous commits. Is the PR number added to the commit automatically or something?

MarijnS95:
@spencercw oh I may have skimmed your question too quickly then and assumed the deadlocks were caused by wrapping things in a `Mutex` on the Rust side for lack of "temporarily borrowing" (e.g. preventing `Drop` to be called) a `MediaCodec` based on the `ffi::AMediaCodec` given to you. If you say that Android givind you that pointer is only a misleading black hole, let's not use that pointer for now.

(The only use I can think of then is typically using the same callback function for multiple objects, and being able to distinguish which of the codec objects just called the callback... But you can also use the userdata pointer for that, as already done by storing extra data in `Box<dyn FnMut>` closures)

> > Let's not have the PR link in the commit title (as if this was a final merge-commit) please :)
> 
> I was just copying the convention from previous commits. Is the PR number added to the commit automatically or something?

It'll be added automatically on _squash_ merge by GitHub indeed :slightly_smiling_face: (and I think it'd try to add it twice otherwise).

ndk/src/media/media_codec.rs
spencercw: Happy to borrow from an existing example if you (or someone) can point to one.

Printing the panic reason is tricky because if this is running inside an app then you're not going to see the console output anywhere. We could use `log::error` on the assumption that the host application is using `android_logger`, but I'm not sure it's worth bringing in `log` just for this. If you panic from a callback you've kind of only got yourself to blame..

(The audio callback also needs something like this, as pointed out in #268.)

MarijnS95: Turns out it was `android-activity`: https://github.com/rust-mobile/android-activity/pull/68/files#diff-45f471af850e20c82d7fa80b3248a1ca0be8df53d3f2950f95b0a3ade881e674R42

Though also consider related issue https://github.com/rust-mobile/android-activity/issues/80, cc @rib.

> Printing the panic reason is tricky because if this is running inside an app then you're not going to see the console output anywhere.

Most Android backends forward `stdout`/`stderr` to Android logs, though the example above calls `ndk_sys::__android_log_write()` directly (on the Error level). Using `log` also has my preference (so that it could be redirected to log files if using e.g. [`fern`](https://crates.io/crates/fern). also CC @rib because `android-activity` doesn't currently do that) but hinges on the app author to initialize the logger.

Sadly `log` is designed such that we cannot provide a default `log` implementation and allow the user to replace it later if they desire... Though there are probably ways or default constructors to work around that: we had this with `ndk-glue`.

> (The audio callback also needs something like this, as pointed out in https://github.com/rust-mobile/ndk/issues/268.)

It'd be lovely if you could open _a separate PR_ that introduces the wrapper and fixes it on `aaudio`, and then put the same wrapper to use in this PR on top.

MarijnS95: Oh yeah, nice workaround and I think we should have something like that on `android-activity` as well, but I'd hate to have to copy it around...

Not sure if `android-logger` is a great place, or the `ndk` crate itself (as a internal or public util) since we also have the sys binding.

spencercw: I don't really like it in the general case because you can't distinguish between 'no logger installed' and 'logging is filtered'. Like, if we try to log something at the info level and the application has filtered to only logging errors, I'm not sure we should be going behind the application's back and writing to the Android log anyway. Logging a panic before aborting is a bit of a special case, but I'm not wild about the idea of exposing such functionality as part of the public API.

ndk/src/media/media_codec.rs
spencercw: This feels somewhat abusive of the `Deref` trait. `OutputBuffer` is not really a smart pointer[^1] for `BufferInfo` (if anything, it's a smart pointer for `&[u8]`). I'd prefer to add a `pub fn info(&self) -> &BufferInfo` method to `OutputBuffer`, but this would imply dropping the existing `flags` and `presentation_time_us` methods, which would obviously be a breaking change.

[^1]:  [`Deref` should only be implemented for smart pointers to avoid confusion.](https://doc.rust-lang.org/std/ops/trait.Deref.html)

ndk/src/media/media_codec.rs
MarijnS95: Fair enough, can we check if a logger is configured _at all_?

There appears to be no level higher than `Error` (e.g. `Fatal`) and I'd be very surprised if users intentionally disable it. This would be their last "breath of life" before the app goes down which might become very hard to trace, so I don't really mind "going behind their back" and emitting some form of error context _somewhere_ at the very least?

ndk/src/media/media_codec.rs
MarijnS95: Let's have `fn info()`, totally fine to break the original API here!

MarijnS95: FYI we'll do a breaking release soon™ to match up with `winit` (a major consumer of the `ndk` crate), and any (breaking) cleanups we could get done to `MediaCodec` the better, so you're more than welcome to.

ndk/src/media/media_codec.rs
spencercw: Not that I can see. `log::logger` returns a no-op logger if nothing is registered, which is rather unhelpful.

Just had another thought. You stated a preference for `log` because it can be redirected to file, but if we log followed immediately by an abort, isn't there a decent chance the message never makes it to disk? I'm kind of leaning towards what `android-activity` already does and just unconditionally writing to the Android system log.

MarijnS95: > Just had another thought. You stated a preference for `log` because it can be redirected to file, but if we log followed immediately by an abort, isn't there a decent chance the message never makes it to disk?

Depending on whether `log` buffers or immediately performs the `write()`, doesn't the OS flush those out on `abort()`?

> I'm kind of leaning towards what `android-activity` already does and just unconditionally writing to the Android system log.

I don't recall us being very thoughtful when writing that up, and am using this opportunity to see if we can easily change/rectify it.

If it becomes too complex we can stick to writing to Android log for now, until someone complains ("the app crashed/aborted, but the logs contain no info, and it took us way too long to spot this `abort()` which 'hid' a mistake in our code").

spencercw: Looks like fern [flushes every log record](https://docs.rs/fern/latest/fern/struct.Output.html#method.from-10), so, I guess that's fine. I'll come up with something tomorrow for you to have a look at.

spencercw: Oh you can [manually flush](https://docs.rs/log/0.4.19/log/trait.Log.html#tymethod.flush) the logger. Handy.

MarijnS95:
@spencercw thanks for #412, it's merged now. Can you rebase this and perhaps look at / mark-as-resolved the above discussion items?

spencercw:
Rebased on master. I think that's all comments addressed now.

ndk/src/media/media_codec.rs
```diff
@@ -246,9 +313,121 @@ impl MediaCodec {
...
+
+        let status = unsafe {
+            ffi::AMediaCodec_setAsyncNotifyCallback(self.as_ptr(), ffi_callback, user_data)
+        };
+        let result = MediaError::from_status(status);
+
+        if result.is_ok() {
+            self.async_notify_callback = callback;
+        } else {
+            self.async_notify_callback = None;
```
MarijnS95, line 425: Should we really clear this here? What if I called this before with success, and now call it again and get an error? If that leaves the original `setAsyncNotifyCallback()` unaffected this line shouldn't be here I think? Otherwise a comment is appreciated.

MarijnS95 (commented):
Looking good already, thanks for doing so many drive-by cleanups and improvements to the `media_codec.rs` API too!

Probably my main concern is not having doc-comments on any of these, but (for functions that were already here before your PR) that should be tackled in a separate PR.

Let me know if you're up for that or if I should tackle this, as I am quite pedantic towards tailoring the upstream docs to Rust :wink:

ndk/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - hardware_buffer_format: Add `YCbCr_P010` and `R8_UNORM` variants. (#405)
 - **Breaking:** hardware_buffer_format: Add catch-all variant. (#407)
 - Add panic guards to callbacks. (#412)
+- **Breaking:** media_codec: Add support for asynchronous notification callbacks. (#410)
```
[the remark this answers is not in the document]
spencercw: Done

ndk/src/media/media_codec.rs
```diff
@@ -214,11 +217,73 @@ impl Drop for MediaFormat {
...
 }
 
+impl fmt::Debug for MediaCodec {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("MediaCodec")
+            .field("inner", &self.inner)
+            .field(
+                "async_notify_callback",
+                match &self.async_notify_callback {
+                    Some(_) => &"Some(_)",
```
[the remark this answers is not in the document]
spencercw: Done

ndk/src/media/media_codec.rs
```diff
@@ -246,9 +313,121 @@ impl MediaCodec {
...
+            user_data: *mut c_void,
+            error: ffi::media_status_t,
+            action_code: i32,
+            detail: *const c_char,
+        ) {
+            abort_on_panic(|| {
+                let callback = &mut *(user_data as *mut AsyncNotifyCallback);
+                if let Some(f) = callback.on_error.as_mut() {
+                    f(
+                        MediaError::from_status(error),
```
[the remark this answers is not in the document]
spencercw: Yeah I didn't really like this. It was just to avoid a panic in case we're called with a zero error for some insane reason. I have changed it to use `unwrap_err` now.

ndk/src/media/media_codec.rs
```diff
@@ -246,9 +313,121 @@ impl MediaCodec {
...
+                        ActionCode(action_code),
+                        CStr::from_ptr(detail),
+                    );
+                }
+            })
+        }
+
+        let (callback, ffi_callback, user_data) = if let Some(callback) = callback {
+            // On Android 12 and earlier, due to faulty null checks, if a callback is not set, but at least one other
+            // callback *is* set, then it will segfault in NdkMediaCodec.cpp when trying to invoke the unset callback.
```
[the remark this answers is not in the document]
spencercw: Not sure if there is a bug report, but I have linked the source code.

ndk/src/media/media_codec.rs
```diff
@@ -386,11 +589,22 @@ impl MediaCodec {
         size: usize,
         time: u64,
         flags: u32,
+    ) -> Result<()> {
+        self.queue_input_buffer_by_index(buffer.index, offset, size, time, flags)
```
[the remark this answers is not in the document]
spencercw: This is sick.

ndk/src/media/media_codec.rs
```diff
@@ -538,3 +747,39 @@ pub enum DequeuedOutputBufferInfoResult<'a> {
...
+    pub fn flags(&self) -> u32 {
+        self.inner.flags
+    }
+}
+
+#[derive(Copy, Clone, Debug)]
+pub struct ActionCode(pub i32);
+
+impl ActionCode {
+    pub fn is_recoverable(&self) -> bool {
```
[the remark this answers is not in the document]
spencercw: Done

ndk/src/media/media_codec.rs
```diff
@@ -246,9 +313,121 @@ impl MediaCodec {
             inner: NonNull::new(unsafe {
                 ffi::AMediaCodec_createEncoderByType(c_string.as_ptr())
             })?,
+            async_notify_callback: None,
         })
     }
 
+    #[cfg(feature = "api-level-28")]
+    pub fn set_async_notify_callback(
```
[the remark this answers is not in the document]
spencercw: Do you want to deal with this along with documenting the other methods, since you offered to do so? :)

ndk/src/media/media_codec.rs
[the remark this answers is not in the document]
MarijnS95: Nit: the affected `||` is on the next line, and you can link multiline:

https://cs.android.com/android/platform/superproject/+/android-12.0.0_r34:frameworks/av/media/ndk/NdkMediaCodec.cpp;l=161-162;drc=ef058464777739e2d9ffad5f00d0e57b186d9a13

ndk/src/media/media_codec.rs
MarijnS95: Here too, let's link to the affected comment and lines instead of the `EXPORT` word at the beginning of the function:

https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/media/ndk/NdkMediaCodec.cpp;l=581-584;drc=8c4e619c7461ac1a8c20c55364643662e9185e4d

ndk/src/media/media_codec.rs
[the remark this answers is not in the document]
MarijnS95: Maybe, maybe not :)

ndk/src/media/media_codec.rs
```diff
@@ -245,9 +305,127 @@ impl MediaCodec {
...
+        unsafe extern "C" fn ffi_on_format_changed(
+            _codec: *mut ffi::AMediaCodec,
+            user_data: *mut c_void,
+            format: *mut ffi::AMediaFormat,
+        ) {
+            abort_on_panic(|| {
+                // Ownership of the format is not documented, but the implementation allocates a new instance and does
+                // not free it, so assume it is ok for us to do so
+                // https://cs.android.com/android/platform/superproject/main/+/refs/heads/main:frameworks/av/media/ndk/NdkMediaCodec.cpp;l=254;drc=5e15c3e22f3fa32d64e57302201123ce41589adf
+                let format = MediaFormat::from_ptr(NonNull::new_unchecked(format));
```
MarijnS95, line 356: I think the comment above mentions that it indeed makes a copy specifically for the client to optionally modify this type, let's include that in the link:


https://cs.android.com/android/platform/superproject/main/+/refs/heads/main:frameworks/av/media/ndk/NdkMediaCodec.cpp;l=248-254;drc=5e15c3e22f3fa32d64e57302201123ce41589adf

MarijnS95 (approved):
Looking good! The only remaining comments are to update the line-ranges in the links, then this is good to merge (and I know I've been asking a lot, thanks for bearing with me...).

ndk/src/media/media_codec.rs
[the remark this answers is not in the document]
spencercw: Ah nice, didn't know you could do multi-line links. Done.

ndk/src/media/media_codec.rs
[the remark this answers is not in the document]
spencercw: Added docs. Is the formatting ok? I have omitted the comments about `AMediaCodec_getInputBuffers()` and `AMediaCodec_getOutputBuffers()` since these functions seem not to exist, and the comments about `userdata` since this is hidden in the Rust API.

ndk/src/media/media_codec.rs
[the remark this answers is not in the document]
spencercw: Sorry this was just a stupid joke. The notion that someone might dequeue an input buffer and then try to queue it into another `MediaCodec` is quite disturbing. 😂 I did actually make this change.

ndk/src/media/media_codec.rs
spencercw: Seems more like a safety thing rather than specifically being for the benefit of the client. In principle we could pass ownership of the `MediaFormat` to the callback instead of passing by reference, but given the ownership is not documented I don't think this is a very good idea. Acquiring ownership in the way we do is kind of a necessary evil to avoid a memory leak.

At any rate, I have updated the link.

spencercw:
No worries. It's good to get it all polished up before merging :)

ndk/src/media/media_codec.rs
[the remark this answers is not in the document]
MarijnS95: Not sure if you want to make the same change for `release_output_buffer()`?

ndk/src/media/media_codec.rs
```diff
@@ -245,9 +325,150 @@ impl MediaCodec {
...
+    /// When called with [`None`] callback, this method unregisters any previously set callback.
+    ///
+    /// Refer to the definition of [`AsyncNotifyCallback`] on how each callback function is called
+    /// and what are specified.
+    ///
+    /// Once the callback is unregistered or the codec is reset / released, the previously
+    /// registered callback will not be called.
+    ///
+    /// All callbacks are fired on one NDK internal thread.
+    /// [`MediaCodec::set_async_notify_callback`] should not be called on the callback thread. No
```
MarijnS95, line 353:
```suggestion
    /// [`MediaCodec::set_async_notify_callback()`] should not be called on the callback thread. No
```

Or just "this"?

ndk/src/media/media_codec.rs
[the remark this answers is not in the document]
spencercw: Done.

ndk/src/media/media_codec.rs
spencercw: Added brackets. Doesn't hurt to be a bit more explicit.

ndk/src/media/media_codec.rs
```diff
@@ -400,18 +657,31 @@ impl MediaCodec {
...
+            unsafe { ffi::AMediaCodec_releaseOutputBuffer(self.as_ptr(), buffer_index, render) };
         MediaError::from_status(status)
     }
 
     pub fn release_output_buffer_at_time(
         &self,
         buffer: OutputBuffer<'_>,
         timestamp_ns: i64,
+    ) -> Result<()> {
+        self.release_output_buffer_at_time_by_index(buffer.index, timestamp_ns)
```
MarijnS95, line 675: Really nitty, but let's be consistent.


```suggestion
         debug_assert!(ptr::eq(self, buffer.codec));
         self.release_output_buffer_at_time_by_index(buffer.index, timestamp_ns)
```

ndk/src/media/media_codec.rs
```diff
@@ -245,9 +325,150 @@ impl MediaCodec {
...
+        };
+
+        let status = unsafe {
+            ffi::AMediaCodec_setAsyncNotifyCallback(self.as_ptr(), ffi_callback, user_data)
+        };
+        let result = MediaError::from_status(status);
+
+        // This behavior is not documented, but the implementation always clears the callback on failure, so we must
+        // clear any callback that may have been previously registered
+        // https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/media/ndk/NdkMediaCodec.cpp;l=559;drc=8c4e619c7461ac1a8c20c55364643662e9185e4d
```
MarijnS95, line 462: https://github.com/rust-mobile/ndk/pull/410/files#r1294049521

```suggestion
        // https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/media/ndk/NdkMediaCodec.cpp;l=581-584;drc=8c4e619c7461ac1a8c20c55364643662e9185e4d
```

ndk/src/media/media_codec.rs
```diff
@@ -245,9 +325,150 @@ impl MediaCodec {
...
+    /// input surface, the client still needs to call [`MediaCodec::start()`] to resume the input
+    /// surface to send buffers to the encoders.
+    ///
+    /// When called with [`None`] callback, this method unregisters any previously set callback.
+    ///
+    /// Refer to the definition of [`AsyncNotifyCallback`] on how each callback function is called
+    /// and what are specified.
+    ///
+    /// Once the callback is unregistered or the codec is reset / released, the previously
+    /// registered callback will not be called.
```
MarijnS95, line 350: Really nitty, but want to make the same remark that previous callbacks are _also unregistered_ when this returns an error?

ndk/src/media/media_codec.rs
MarijnS95: I won't be opposed to moving it, this is one of those implicit API things that Google cannot change anymore since everyone is expecting this to be a memory leak when they don't "own" the pointer (which is silly, but alas).

Also just noticing that all the `fn set_*()` functions take `&self` _im_mutable :eyes:

MarijnS95 (approved):

ndk/src/media/media_codec.rs
spencercw: Done.

ndk/src/media/media_codec.rs
spencercw: Not sure how I feel about documenting implementation behaviour that is not mentioned in the upstream documentation. I guess it's not the worst thing since we rely on that behaviour within this function anyway, just feels a bit icky..

ndk/src/media/media_codec.rs
spencercw: I feel like most of the functions in this file probably ought to be `&mut self`. I'm not sure it makes sense to be allowing such free reign on a `MediaCodec` via a shared reference.

ndk/src/media/media_codec.rs
MarijnS95: Then let's not, and let's have our Rust docs as incomplete as the NDK ones :ok_hand:

MarijnS95 (approved):

ndk/src/media/media_codec.rs
MarijnS95: We can still fix that before the next release, if you want @spencercw. CC @zarik5 in case there was a thoughtful reason for this?

ndk/src/media/media_codec.rs
zarik5: That's a bit sad to be honest but 🤷‍♂️

MarijnS95: @zarik5 what's sad? The fact that AOSP unregisters previously-successfully-registered callbacks on error, or the fact that we're not going to document it?

ndk/src/media/media_codec.rs
MarijnS95: @zarik5 feel free to contribute that!

Also, are you saying that `MediaCodec` could perhaps be `Send`/`Sync`, does the NDK implement synchronization internally?

ndk/src/media/media_codec.rs
zarik5: The documentation bit.

PR #412, opened 2023-08-01
Add panic guards to callbacks

### LINKED ISSUE #268: aaudio callbacks are unsound

spencercw:
As discussed in #410. Mostly copied from the equivalent code in [android-activity](https://github.com/rust-mobile/android-activity/blob/96497f9da990c0006ff88076d06cc50a7870d1dc/android-activity/src/util.rs). Resolves #268.

### DIFF
--- a/ndk/src/utils.rs
+++ b/ndk/src/utils.rs
@@ -1,4 +1,6 @@
 //! Internal utilities
+use log::{error, log_enabled, Level};
+use std::ffi::{c_int, CStr, CString};
 use std::io::{Error, Result};
 
 /// Turns standard `<errno.h>` status codes - typically rewrapped by Android's [`Errors.h`] - into
@@ -11,4 +13,72 @@ pub(crate) fn status_to_io_result<T>(status: i32, value: T) -> Result<T> {
         r if r < 0 => Err(Error::from_raw_os_error(-r)),
         r => unreachable!("Status is positive integer {}", r),
     }
+}
+
+pub(crate) fn android_log(level: Level, tag: &CStr, msg: &CStr) {
+    let prio = match level {
+        Level::Error => ffi::android_LogPriority::ANDROID_LOG_ERROR,
+        Level::Warn => ffi::android_LogPriority::ANDROID_LOG_WARN,
+        Level::Info => ffi::android_LogPriority::ANDROID_LOG_INFO,
+        Level::Debug => ffi::android_LogPriority::ANDROID_LOG_DEBUG,
+        Level::Trace => ffi::android_LogPriority::ANDROID_LOG_VERBOSE,
+    };
+    unsafe {
+        ffi::__android_log_write(prio.0 as c_int, tag.as_ptr(), msg.as_ptr());
+    }
+}
+
+pub(crate) fn log_panic(panic: Box<dyn std::any::Any + Send>) {
+    let rust_panic = unsafe { CStr::from_bytes_with_nul_unchecked(b"RustPanic\0") };
+
+    // Use the Rust logger if installed and enabled, otherwise fall back to the Android system
+    // logger so there is at least some record of the panic
+    let use_log = log_enabled!(Level::Error);
+
+    match panic.downcast::<String>() {
+        Ok(panic_string) => {
+            if use_log {
+                error!("RustPanic: {}", panic_string);
+            } else if let Ok(msg) = CString::new(*panic_string) {
+                android_log(Level::Error, rust_panic, &msg);
+            }
+        }
+        Err(panic) => match panic.downcast::<&str>() {
+            Ok(panic_str) => {
+                if use_log {
+                    error!("RustPanic: {}", panic_str);
+                } else if let Ok(msg) = CString::new(*panic_str) {
+                    android_log(Level::Error, rust_panic, &msg);
+                }
+            }
+            Err(_) => {
+                if use_log {
+                    error!("UnknownPanic");
+                } else {
+                    let unknown_panic =
+                        unsafe { CStr::from_bytes_with_nul_unchecked(b"UnknownPanic\0") };
+                    android_log(Level::Error, unknown_panic, Default::default());
+                }
+            }
+        },
+    }
+
+    if use_log {
+        log::logger().flush();
+    }
+}
+
+/// Run a closure and abort the program if it panics.
+///
+/// This is generally used to ensure Rust callbacks won't unwind past the FFI boundary, which leads
+/// to undefined behaviour.
+pub(crate) fn abort_on_panic<R>(f: impl FnOnce() -> R) -> R {
+    std::panic::catch_unwind(std::panic::AssertUnwindSafe(f)).unwrap_or_else(|panic| {
+        // Try logging the panic before aborting
+        //
+        // Just in case our attempt to log a panic could itself cause a panic we use a
+        // second catch_unwind here.
+        let _ = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| log_panic(panic)));
+        std::process::abort();
+    })
 }
--- a/ndk/src/utils.rs
+++ b/ndk/src/utils.rs
@@ -29,42 +29,28 @@ pub(crate) fn android_log(level: Level, tag: &CStr, msg: &CStr) {
 }
 
 pub(crate) fn log_panic(panic: Box<dyn std::any::Any + Send>) {
-    let rust_panic = unsafe { CStr::from_bytes_with_nul_unchecked(b"RustPanic\0") };
+    fn log_panic(panic_str: &str) {
+        const RUST_PANIC_TAG: &CStr =
+            unsafe { CStr::from_bytes_with_nul_unchecked(b"RustPanic\0") };
 
-    // Use the Rust logger if installed and enabled, otherwise fall back to the Android system
-    // logger so there is at least some record of the panic
-    let use_log = log_enabled!(Level::Error);
+        let panic_str = CString::new(panic_str).unwrap_or_default();
 
-    match panic.downcast::<String>() {
-        Ok(panic_string) => {
-            if use_log {
-                error!("RustPanic: {}", panic_string);
-            } else if let Ok(msg) = CString::new(*panic_string) {
-                android_log(Level::Error, rust_panic, &msg);
-            }
+        // Use the Rust logger if installed and enabled, otherwise fall back to the Android system
+        // logger so there is at least some record of the panic
+        if log_enabled!(Level::Error) {
+            error!("RustPanic: {}", panic_str.to_string_lossy());
+            log::logger().flush();
+        } else {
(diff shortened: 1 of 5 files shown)


ndk/src/utils.rs
```diff
@@ -12,3 +14,73 @@ pub(crate) fn status_to_io_result<T>(status: i32, value: T) -> Result<T> {
...
+                }
+            }
+            Err(_) => {
+                if use_log {
+                    error!("UnknownPanic");
+                } else {
+                    let unknown_panic =
+                        unsafe { CStr::from_bytes_with_nul_unchecked(b"UnknownPanic\0") };
+                    android_log(Level::Error, unknown_panic, unsafe {
+                        CStr::from_bytes_with_nul_unchecked(b"\0")
```
MarijnS95, line 61: I'd be happier with `CStr::default()`.

ndk/src/utils.rs
```diff
@@ -12,3 +14,73 @@ pub(crate) fn status_to_io_result<T>(status: i32, value: T) -> Result<T> {
...
+        ffi::__android_log_write(prio.0 as c_int, tag.as_ptr(), msg.as_ptr());
+    }
+}
+
+pub(crate) fn log_panic(panic: Box<dyn std::any::Any + Send>) {
+    let rust_panic = unsafe { CStr::from_bytes_with_nul_unchecked(b"RustPanic\0") };
+
+    // Use the Rust logger if installed and enabled, otherwise fall back to the Android system
+    // logger so there is at least some record of the panic
+    let use_log = log_enabled!(Level::Error);
```
MarijnS95, line 36: @rib we might want to port this back to `android-activity`.

MarijnS95 (approved):
Thanks!

ndk/src/utils.rs
spencercw: `Default` is actually implemented on `&CStr`, and it's not at all clear to me how to invoke this explicitly, so I have just done it as `Default::default()`.

MarijnS95: Good point, I didn't think through that the returned type is `&CStr`, not `CStr`. In case you're curious, parenthesis that are normally used to adjust precedence in expressions are replaced by angle brackets when doing this to types: `<&CStr>::default()`.

ndk/src/utils.rs
```diff
@@ -12,3 +14,71 @@ pub(crate) fn status_to_io_result<T>(status: i32, value: T) -> Result<T> {
...
+
+    match panic.downcast::<String>() {
+        Ok(panic_string) => {
+            if use_log {
+                error!("RustPanic: {}", panic_string);
+            } else if let Ok(msg) = CString::new(*panic_string) {
+                android_log(Level::Error, rust_panic, &msg);
+            }
```
MarijnS95, line 44: Wasn't sure about asking this last time, but is there no way to simply abstract away the `if - else if let` to a function in such a way that it can be reused 3 times?

spencercw: I did try, but the third case is dissimilar enough to the first two to make it pretty messy. It's a bit shorter by line count, but total number of characters is about the same. My judgement was that the additional complexity is not worth it to save a few lines.

```rust
pub(crate) fn log_panic(panic: Box<dyn std::any::Any + Send>) {
    fn log_panic_string<S: Into<Vec<u8>> + Display>(tag: &CStr, panic_string: Option<S>) {
        // Use the Rust logger if installed and enabled, otherwise fall back to the Android system
        // logger so there is at least some record of the panic
        if log_enabled!(Level::Error) {
            if let Some(s) = panic_string {
                error!("{}: {}", tag.to_str().unwrap(), s);
            } else {
                error!("{}", tag.to_str().unwrap());
            }
            log::logger().flush();
        } else {
            let panic_c_string = panic_string.map_or(None, |s| CString::new(s).ok());
            let panic_c_str = panic_c_string
                .as_ref()
                .map(|s| s.as_c_str())
                .unwrap_or_default();
            android_log(Level::Error, tag, panic_c_str);
        }
    }

    let rust_panic = unsafe { CStr::from_bytes_with_nul_unchecked(b"RustPanic\0") };
    match panic.downcast::<String>() {
        Ok(panic_string) => log_panic_string(rust_panic, Some(*panic_string)),
        Err(panic) => match panic.downcast::<&str>() {
            Ok(panic_str) => log_panic_string(rust_panic, Some(*panic_str)),
            Err(_) => log_panic_string::<String>(
                unsafe { CStr::from_bytes_with_nul_unchecked(b"UnknownPanic\0") },
                None,
            ),
        },
    }
}
```

ndk/src/utils.rs
spencercw: That's good to know, thanks. Want me to change it or shall I just leave it?

ndk/src/utils.rs
MarijnS95: I would have personally gone for something much more simple. Starting with an `UnknownPanic` still being a `RustPanic`, I prefer keeping the tag the same and getting away of all the complexity around it. Then it becomes as simple as accepting an `Option<&str>`, to which a `Box<String>` also deref's :)

Pushed the suggsetion straight away after "prototyping" this idea, saving you the hassle of having to rewrite it. Let me know what you think!

spencercw: Looks good. The reason I used `<S: Into<Vec<u8>> + Display>` is because it lets `CString::new()` take the inner `Vec` from `String` and push a `0` on the end, potentially saving an allocation. This may be an unnecessary micro-optimisation now that I think of it..

MarijnS95: It seems likely to me that `length == capacity` in most cases, and a reallocation is necessary either way.

If not we could also implement this with a simple `panic_string: String`, then the owned copy from `&str` to something that contains a `Vec<u8>` happens without generics.

spencercw: I think you're right that `String` is likely to be tightly-packed most of the time. Taking `String` would mean that a `&str` panic requires two allocations (one from `&str` to `String`, then another from `String` to `CString`). I think just leave it as it is with `&str`.

MarijnS95:
@spencercw not sure if worth/relevant here, but I learned about [`resume_unwind()`](https://doc.rust-lang.org/std/panic/fn.resume_unwind.html) today which is able to resume an unwind (via a type-erased boxed error) after we've caught it before entering an FFI boundary. Assuming we can sometimes channel (a raw pointer to) that boxed error back to the call site - _and the callback is fired somewhere within Rust code_ - we could resume the unwind instead of plainly aborting.

Not that it adds much though; it's basically a panic as if you were to `unwrap()` the boxed error, but without invoking the panic hook, so no extra stacktrace is printed (but unwinding and drop handlers _should_ continue?).

For example, I already want to port this code to `android-activity` and I think we had a use-case there where we use an FFI layer which is just Rust on both sides (or we deduced that the use of C symbols to go Rust -> Rust wasn't actually an FFI boundary - or there were just a few frames of non-Rust), and it might come in handy there.

---

Final note: I think we forgot to wrap the callback in `Looper` in this new helper :)
(And this is where we _could_ - but probably not worth it - set aside the boxed error and `resume_unwind()` it in the `poll` functions)

spencercw:
`resume_unwind` can be useful, but you need to be reasonably confident that you are actually going to be called somewhere that you can resume the panic. The `MediaCodec` async callback for example is tricky because if the callback function panics, then it's likely that the user's main thread won't be woken up and won't call any other `MediaCodec` functions, so the panic would be swallowed and the application will likely hang.

I do actually use `resume_unwind` quite a lot in my code to transport panics from worker threads. I have the main thread receiving events from the worker thread on a channel. If the `recv` fails then I know the worker thread must have exited, so at that point I `join` the thread which returns the panic in the `Err`, which I can then `resume_unwind`. This works nicely because the main thread is automatically woken up when the worker thread panics, just by virtue of the `Sender` being dropped when it goes out of scope.

I'm not super familiar with `Looper`, but it sounds like this is a scenario where `resume_unwind` could work.

MarijnS95:
Exactly, if there's no clear point where the callback is going to be called this won't work. 

What's your thought about not getting a second backtrace from the point you called `resume_unwind()` back to the start of your application?

---

For `Looper` I am not sure either. There are these specific `poll*()` functions that report back whether they executed a callback, and because the looper is associated with a thread I think those callbacks run synchronously when the user calls those `poll*()` functions. In turn the NDK wrapper can simply check if an `Err` was stored after `poll*()` returns.

However, there are APIs like [`AndroidBitmap_compress()`](https://developer.android.com/ndk/reference/group/bitmap#androidbitmap_compress) which call a callback (I think, not specified in the docs!) only for the duration of the function. This could simply be a Rust closure that stores the panic reason externally so that we can propagate it when the FFI function returns.

spencercw:
> What's your thought about not getting a second backtrace from the point you called `resume_unwind()` back to the start of your application?

Not quite sure what you mean by this.

> > This method can be called on any thread. This method may block briefly if it needs to wake the poll.
> 
> Not sure if that means "any thread, as long as it's the thread that calls `_poll*()`.

I presume this means you can submit file descriptors to a looper from any thread, but the polling will happen on the thread that the looper is actually running on.

> However, there are APIs like [`AndroidBitmap_compress()`](https://developer.android.com/ndk/reference/group/bitmap#androidbitmap_compress) which call a callback (I think, not specified in the docs!) only for the duration of the function. This could simply be a Rust closure that stores the panic reason externally so that we can propagate it when the FFI function returns.

This does seem like an ideal use case for `resume_unwind`.

MarijnS95:
Apologies for the late reply, it's been a while since I could find time.

> Not quite sure what you mean by this.

`resume_unwind()` intentionally avoids the panic handler, which is the mechanism that prints panic traces to the log. When being used, the panic trace from a `panic!()` up to `catch_unwind()` is being printed, but the trace from `resume_unwind()` up to the entry point is not. For that we pretty much need to `panic!()`/`unwrap()` the returned boxed error, which may or may not be pretty/desired?

> I presume this means you can submit file descriptors to a looper from any thread, but the polling will happen on the thread that the looper is actually running on.

Likely, but there's no wording excluding that. We could test to find out but no guarantees... I'll keep this in mind regardless as this API needs to be revisited anyway.

> This does seem like an ideal use case for `resume_unwind`.

It is, hoping to submit the PR for this soon including mappings for `DataSpace`.

PR #414, opened 2023-08-10
media_error: Add `MediaStatus::from_status_if_negative()` helper

MarijnS95:
CC @paxbun

Covers a prevalent `unwrap_err()` pattern for negative status codes, where we want to use any non-negative status code as a useful value and otherwise assert that `from_status()` turns the raw value into an appropriate error code (never into `Ok(())`, which should be unreachable in the current implementation).

Note that we cannot rely on `from_status()` returning `Ok(())` or `Err(_)` and translate the `Ok(())` back to the original value, as it handles (as of writing) two positive error codes which only relate to a specific `AMediaCodec` API, and any unrecognized codes including >0 are treated as `UnknownStatus`.

The function consumes `Into<isize>` instead of `i32/c_int` or `ffi::media_status_t` to support passing through numbers from various APIs that return 64-bit numbers (e.g. `MediaCodec`).  The value only needs to be truncated (with a panic on `try_into()`) when it is lower than zero and needs to be converted to `ffi::media_status_t`.

### DIFF
--- a/ndk/src/media/media_codec.rs
+++ b/ndk/src/media/media_codec.rs
@@ -525,13 +525,12 @@ impl MediaCodec {
 
         if result == ffi::AMEDIACODEC_INFO_TRY_AGAIN_LATER as isize {
             Ok(DequeuedInputBufferResult::TryAgainLater)
-        } else if result >= 0 {
+        } else {
+            let index = MediaError::from_status_if_negative(result)? as usize;
             Ok(DequeuedInputBufferResult::Buffer(InputBuffer {
                 codec: self,
-                index: result as usize,
+                index,
             }))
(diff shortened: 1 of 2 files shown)


rib (approved):
Urgh, kinda horrible that `media_status_t` has some non-negative status codes if some apis are designed to convey errors via negative values or else useful values :/

This utility API looks good to me though.

MarijnS95:
Yeah doesn't have too much to do with this extra util function, but yes it is totally unusable because these error codes are supposedly only returned by API/functions that also have other non-zero _positive_ return values :weary: 

Come to think of it, I don't think the current code handles those positive codes at all.

PR #416, opened 2023-08-10
ndk/looper: Add `remove_fd()` method

MarijnS95:
We've previously only been able to unregister file descriptors by using a callback and returning `false` from it to be unregistered.  Add the missing method that allows users to unregister their file descriptors from the looper for any reason.

There has always been a "concern" about leaking the `Box` inside `add_fd_with_callback()`, but this is already very easy by never returning `false` from the callback or registering a different callback / event on the same `fd`, so this new function doesn't really expose that.

### DIFF
--- a/ndk/CHANGELOG.md
+++ b/ndk/CHANGELOG.md
@@ -16,6 +16,7 @@
 - hardware_buffer_format: Add `YCbCr_P010` and `R8_UNORM` variants. (#405)
 - **Breaking:** hardware_buffer_format: Add catch-all variant. (#407)
 - Add panic guards to callbacks. (#412)
+- looper: Add `remove_fd()` to unregister events/callbacks for a file descriptor. (#416)
 
 # 0.7.0 (2022-07-24)
 
--- a/ndk/src/looper.rs
+++ b/ndk/src/looper.rs
@@ -267,11 +267,14 @@ impl ForeignLooper {
 
     /// Adds a file descriptor to be polled, with a callback.
     ///
-    /// The callback takes as an argument the file descriptor, and should return true to continue
-    /// receiving callbacks, or false to have the callback unregistered.
+    /// The callback takes as an argument the file descriptor, and should return [`true`] to
+    /// continue receiving callbacks, or [`false`] to have the callback unregistered.
     ///
     /// See also [the NDK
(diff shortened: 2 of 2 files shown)


rib (approved):
Looks reasonable to me.

Maybe for the issue with leaking the Box there could be some way of having `add_fd_with_callback` return some form of handle that could then be used to remove that callback, which could also make sure to drop the Box.

Not sure how that would work with respect to how removeFd doesn't guarantee that there can't be more calls to the callback after removing the fd.

I wonder if it's possible to get better guarantees with `addFd(fd, callback=null)` to explicitly clear the callback before `removeFd`

Skimming the implementation I'm not super sure: https://android.googlesource.com/platform/system/core/+/master/libutils/Looper.cpp#432

MarijnS95:
> Maybe for the issue with leaking the Box there could be some way of having `add_fd_with_callback` return some form of handle that could then be used to remove that callback, which could also make sure to drop the Box.

Perhaps we could, since a `ForeignLooper` can call `remove_fd` and is `Send`/`Sync` and probably also easily clonable.

Alternatively we could keep a `HashMap` of sorts mapping `fd` to `Box`. Then we can _also_ catch _overwrites_ when the same `fd` in `add_fd(_with_callback)()`. Not sure what to do when users close `fd`s and get the same `fd` back on subsequent `open()` etc, it's up to them if they forgot to `remove_fd` it.

---

Anyway I tried to not mention (in the docs) the fact that it might be hard to trigger the callback just to return `false` currently (maybe via `EPOLLHUP`) to clean up the `Box`. But let's mention it here anyway.

MarijnS95:
Regardless, leaking memory isn't unsafe/UB (these APIs playing with random `fd`s that we have no control over _are_, maybe they should've been marked `unsafe` to that effect), so I'll just merge this now I think and revisit this once we have a concrete user with this problem and maybe some more time / interest in thinking up the right solution.

MarijnS95:
I intentionally didn't bring that up. Many parts of the `ndk` are trivial to get raw pointers from (or otherwise) and process "things" outside of our Rust wrapper, and if we have to concern ourselves with that it becomes much harder to build a fully sound wrapper. This is anyway similar to the user e.g. closing the fd while it is still recorded inside the `Looper`, as mentioned before.

PR #417, opened 2023-08-10
ndk: Use `BorrowedFd` and `OwnedFd` to clarify ownership transitions

MarijnS95:
Depends on #416

Some functions consume a file descriptor (will close them on their own regard) or return ownership over a file descriptor (expect the caller to close it), but this is not always clarified in the documentation nor upheld by the caller.  Use the new stabilized `BorrowedFd` and `OwnedFd` types since Rust 1.63 to clarify this in the API, noting that `OwnedFd` will instinctively `close()` the file descriptor on drop and doesn't implement `Copy` nor `Clone` (but does provide a `try_clone()` helper using `fcntl()` to create a new owned file descriptor if needed, and if possible by the kernel).

For example, while not obvious from `AHardwareBuffer_lock()` docs (though there are hints in the [graphics sync docs]) the source for gralloc buffer locking many function calls down clarifies that the [`acquireFence` is indeed owned and will be closed].  The same [applies to `AImageReader` and its async aquire functions].

[graphics sync docs]: https://source.android.com/docs/core/graphics/sync
[`acquireFence` is indeed owned and will be closed]: https://cs.android.com/android/platform/superproject/main/+/refs/heads/main:frameworks/native/libs/ui/Gralloc4.cpp;l=320-323;drc=34edaadf5297f2c066d2cb09a5cc9366dc35b24b
[applies to `AImageReader` and its async aquire functions]: https://cs.android.com/android/platform/superproject/main/+/refs/heads/main:frameworks/av/media/ndk/NdkImageReader.cpp;l=498-501;drc=34edaadf5297f2c066d2cb09a5cc9366dc35b24b

### DIFF
--- a/ndk/src/looper.rs
+++ b/ndk/src/looper.rs
@@ -12,9 +12,9 @@
 use bitflags::bitflags;
 use std::mem::ManuallyDrop;
 use std::os::raw::c_void;
-use std::os::unix::io::RawFd;
+// TODO: Import from std::os::fd::{} since Rust 1.66
+use std::os::unix::io::{AsRawFd, BorrowedFd, RawFd};
 use std::ptr;
-use std::ptr::NonNull;
 use std::time::Duration;
 use thiserror::Error;
 
@@ -42,7 +42,7 @@ bitflags! {
 
 /// The poll result from a [`ThreadLooper`].
 #[derive(Debug)]
-pub enum Poll {
+pub enum Poll<'fd> {
     /// This looper was woken using [`ForeignLooper::wake()`]
     Wake,
     /// For [`ThreadLooper::poll_once*()`][ThreadLooper::poll_once()], an event was received and processed using a callback.
@@ -52,7 +52,10 @@ pub enum Poll {
     /// An event was received
     Event {
         ident: i32,
-        fd: RawFd,
+        /// # Safety
+        /// The caller should guarantee that this file descriptor remains open after it was added
+        /// via [`ForeignLooper::add_fd()`] or [`ForeignLooper::add_fd_with_callback()`].
+        fd: BorrowedFd<'fd>,
         events: FdEvent,
         data: *mut c_void,
     },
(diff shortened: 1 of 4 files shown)


rib (approved):

ndk/src/looper.rs
```diff
@@ -52,7 +52,10 @@ pub enum Poll {
     /// An event was received
     Event {
         ident: i32,
-        fd: RawFd,
+        /// # Safety
+        /// The caller should guarantee that this file descriptor remains open after it was added
+        /// via [`ForeignLooper::add_fd()`] or [`ForeignLooper::add_fd_with_callback()`].
+        fd: BorrowedFd<'fd>,
```
MarijnS95, line 58: I will think about this a bit more before merging (cc @rib).

This is an `fd` the user gave to us (as a `BorrowedFd`, though) so they're already required to make sure it isn't closed until it is removed from the `Looper`. Second, as written above the `unsafe` block that creates this:

```
                // SAFETY: Even though this FD at least shouldn't outlive self, a user could have
                // closed it after calling add_fd or add_fd_with_callback.
```

There's barely any relevance to `self` here; the user can totally keep using _their own_ `fd` outside the scope of `Looper`, we really have no clue what its lifetime should be.

Most importantly, let's check if `winit` can easily handle the lifetime that we've now added to `Poll<'fd>`.

MarijnS95:
Turns out I made the biggest mistake I could here: by calling `as_raw_fd()` on an `OwnedFd` we're merely borrowing the `OwnedFd`, meaning it gets dropped by us at the end of the Rust function rather than "transferring" ownership to the FFI function, meaning `close()` ends up being called twice.

Converted this PR to draft for that as I want to test out the new codepaths first before merging this... and while setting up a test-case I ran into more lifetime/reference issues in the NDK :weary:. Release will probably be bleeding into next week.

MarijnS95:
I managed to test the `lock()` ownership transition with an `ImageReader` where I use its producer end (the `NativeWindow` that we refcounted wrongly, #418) in EGL: this lets the `ImageReader` return a `SyncFence` (blocking on GL GPU completion) file descriptor when asynchronously acquiring the latest `Image`, which can subsequently be used to `.lock()` the `HardwareBuffer` inside the `Image`. I tested the ownership transitions here with some `libc::close()` (since the `drop()` for `OwnedFd` discards the error) and confirmed that the NDK closes it for us.

PR #423, opened 2023-08-17
ndk: Add bindings for `sync.h`

MarijnS95:
Bind a few free functions and conveniently re-wrap the resulting struct types to access sync information from file descriptors, and merge sync file descriptors into one.

### DIFF
--- a/ndk/src/sync.rs
+++ b/ndk/src/sync.rs
@@ -1,0 +1,143 @@
+//! Bindings for [sync functions]
+//!
+//! [sync functions]: https://developer.android.com/ndk/reference/group/sync
+#![cfg(feature = "sync")]
+
+use std::{
+    ffi::CStr,
+    fmt::Debug,
+    // TODO: Import from std::os::fd::{} since Rust 1.66
+    os::unix::io::{AsRawFd, BorrowedFd, FromRawFd, OwnedFd},
+    ptr::NonNull,
+};
+
+#[doc(alias = "sync_file_info")]
+#[repr(transparent)]
+pub struct SyncFileInfo {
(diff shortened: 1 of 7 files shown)


ndk/src/sync.rs
```diff
@@ -0,0 +1,140 @@
+//! Bindings for [sync functions]
+//!
+//! [sync functions]: https://developer.android.com/ndk/reference/group/sync
+#![cfg(feature = "sync")]
+
+use core::ffi::FromBytesUntilNulError;
```
MarijnS95, line 6: This `CStr` API unfortunately requires MSRV 1.69... :(

ndk/src/sync.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// Retrieve detailed information about a sync file and its fences.
+    #[doc(alias = "sync_file_info")]
+    pub fn new(fd: BorrowedFd<'_>) -> Option<Self> {
+        let inner = NonNull::new(unsafe { ffi::sync_file_info(fd.as_raw_fd()) })?;
+        Some(Self { inner })
+    }
+
+    pub fn name(&self) -> Result<&CStr, FromBytesUntilNulError> {
+        let inner = unsafe { self.inner.as_ref() };
+        CStr::from_bytes_until_nul(&inner.name)
```
MarijnS95, line 43: Stupidly, unlike `CStr::from_ptr`, this takes `&[u8]` instead of `&[c_char]`. The underlying FFI arrays are defined with `c_char`, which will be `i8` and incompatible with this function when we build for a non-Android target to e.g. run "tests" on the host machine in the CI :(

(And the 1.69 MSRV...)

PR #424, opened 2023-08-22
ndk/bitmap: Provide detailed implementation for `AndroidBitmapInfoFlags`

MarijnS95:
As per the TODO, wrap this flags type with some helpers that fish out the two bit-representations for the `IS_HARDWARE` flag and `ALPHA` value.

Also add missing documentation and aliases across the `bitmap` module.

CC @dcsommer since you most recently touched this module and are likely relying on it: can you give this a review?

### DIFF
--- a/ndk/src/bitmap.rs
+++ b/ndk/src/bitmap.rs
@@ -14,11 +14,14 @@ use std::mem::MaybeUninit;
 use crate::hardware_buffer::HardwareBufferRef;
 
 #[repr(i32)]
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum BitmapError {
     Unknown,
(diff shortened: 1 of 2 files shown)


ndk/src/bitmap.rs
```diff
@@ -111,27 +137,87 @@ impl AndroidBitmap {
...
+    /// Pixel components are independent of alpha.
+    #[doc(alias = "ANDROID_BITMAP_FLAGS_ALPHA_UNPREMUL")]
+    Unpremultiplied,
+}
+
+/// Bitfield containing information about the bitmap.
+#[cfg(feature = "api-level-30")]
+#[derive(Clone, Copy)]
+#[repr(transparent)]
+pub struct AndroidBitmapInfoFlags(u32);
```
MarijnS95, line 159: Unsure if I should make this into a proper `bitflags`, there are currently only two valid values (the alpha mask, and the `IS_HARDWARE` flag).

PR #425, opened 2023-08-22
ndk/native_window: Add missing function wrappers

MarijnS95:
- ndk/native_window: Add `set_buffers_transform()` API
- ndk/native_window: Add `try_allocate_buffers()` API
- ndk/native_window: Add `set_frame_rate*()` APIs since level 30/31

Note that this _lacks_ `ANativeWindow_readFromParcel`/`writeToParcel` which'll only appear in API level 34 with R26 which is still in beta.  Besides, we don't yet bind `AParcel` either.

### DIFF
--- a/ndk/Cargo.toml
+++ b/ndk/Cargo.toml
@@ -13,7 +13,7 @@ repository = "https://github.com/rust-mobile/ndk"
 rust-version = "1.66"
 
 [features]
-all = ["audio", "bitmap","media", "api-level-30"]
+all = ["audio", "bitmap","media", "api-level-31"]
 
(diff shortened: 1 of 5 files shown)


ndk/Cargo.toml
```diff
@@ -13,7 +13,7 @@ repository = "https://github.com/rust-mobile/ndk"
 rust-version = "1.64"
 
 [features]
-all = ["audio", "bitmap","media", "api-level-30"]
+all = ["audio", "bitmap","media", "api-level-31"]
```
MarijnS95, line 16: Looks like I forgot this in #428.

PR #427, opened 2023-08-22
ndk: Add bindings for `ASharedMemory`

MarijnS95:
`ASharedMemory` allows allocating shared memory as a file descriptor that can be `mmap`ed or shared to other processes via parcels or Unix domain sockets.

Its Java counterpart is `android.os.SharedMemory` from which (when available as a JNI `jobject`) an owned file descriptor clone can be created to access the memory object in native code.

Common Rust traits such as `FromRawFd` are provided so that a `SharedMemory` object can also be created from a received file descriptor.

### DIFF
--- a/ndk/Cargo.toml
+++ b/ndk/Cargo.toml
@@ -51,6 +51,10 @@ package = "ndk-sys"
 path = "../ndk-sys"
 version = "0.5.0-beta.0"
 
+[dev-dependencies]
+# Only for use in documentation and doc-tests
+libc = "0.2"
+
 [package.metadata.docs.rs]
 features = ["jni", "all"]
 rustdoc-args = ["--cfg", "docsrs"]
--- a/ndk/src/shared_memory.rs
+++ b/ndk/src/shared_memory.rs
@@ -1,0 +1,153 @@
+//! Bindings for [`ASharedMemory`]
+//!
+//! [`ASharedMemory`]: https://developer.android.com/ndk/reference/group/memory
+#![cfg(feature = "api-level-26")]
+
+use std::{
+    ffi::CStr,
+    io::{Error, Result},
+    // TODO: Import from std::os::fd::{} since Rust 1.66
+    os::unix::io::{AsFd, AsRawFd, BorrowedFd, FromRawFd, IntoRawFd, OwnedFd, RawFd},
+    ptr,
+};
+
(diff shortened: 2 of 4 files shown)


ndk/Cargo.toml
```diff
@@ -47,6 +47,10 @@ package = "ndk-sys"
 path = "../ndk-sys"
 version = "0.5.0-beta.0"
 
+[dev-dependencies]
+# Only for use in documentation and doc-tests
+libc = "0.2"
```
MarijnS95, line 52: Before merging this - if folks find this a good idea at all - I'd like to go over existing documentation and replace their mentions of e.g. `mmap` and other flags with `libc` intradoc links.

ndk/src/shared_memory.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// Creates shared memory region and returns a file descriptor. The resulting file descriptor
+    /// can be `mmap`'ed to process memory space with `PROT_READ | PROT_WRITE | PROT_EXEC`. Access
+    /// to shared memory region can be restricted with [`set_prot()`][Self::set_prot()].
+    ///
+    /// Use [`android.os.ParcelFileDescriptor`] to pass the file descriptor to another process.
+    /// File descriptors may also be sent to other processes over a Unix domain socket with
+    /// `sendmsg` and `SCM_RIGHTS`. See `sendmsg(3)` and `cmsg(3)` man pages for more information.
+    ///
+    /// If you intend to share this file descriptor with a child process after calling `exec(3)`,
+    /// note that you will need to use `fcntl(2)` with `FD_SETFD` to clear the `FD_CLOEXEC` flag
```
MarijnS95, line 32: And through the magic of intradoc links I learned that `FD_SETFD` must be a typo in the AOSP documentation; it is `F_SETFD` :)

ndk/Cargo.toml
MarijnS95: Appears this is not true at all. `dev-dependencies` applies to (doc!) tests, but not to documentation itself.

I can add it behind the `docsrs` cfg then... :(

MarijnS95: Note too that many of these constants - except the `PROT_` ones, are (accidentally?) included in the `ndk-sys` ffi bindings...

But better to reference public/obvious ones from `libc` unless there's a "kernel header" target?

PR #434, opened 2023-10-07
ndk/native_window: Implement `HasWindowHandle` for 0.4 and 0.6 (and maintain 0.5)

MarijnS95:
Following [the same strategy in winit], move `raw-window-handle 0.5` behind a `rwh_05` feature and add `raw-window-handle 0.4` and the newly released `raw-window-handle 0.6` behind a corresponding `rwh04`/`rwh_06` feature.  The new **non-raw** trait and **lifetimed** type since `0.6` have been implemented instead, as `raw-window-handle` provides blanket implementations to coerce/convert back to `Raw*` types.

[the same strategy in winit]: https://github.com/rust-windowing/winit/pull/3126

### DIFF
--- a/ndk/src/native_window.rs
+++ b/ndk/src/native_window.rs
@@ -6,8 +6,7 @@ use crate::utils::status_to_io_result;
 
 pub use super::hardware_buffer_format::HardwareBufferFormat;
 use jni_sys::{jobject, JNIEnv};
-use raw_window_handle::{AndroidNdkWindowHandle, HasRawWindowHandle, RawWindowHandle};
-use std::{ffi::c_void, io::Result, mem::MaybeUninit, ptr::NonNull};
+use std::{ffi::c_void, io, mem::MaybeUninit, ptr::NonNull};
 
 pub type Rect = ffi::ARect;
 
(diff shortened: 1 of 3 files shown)


ndk/src/native_window.rs
```diff
@@ -39,11 +38,22 @@ impl Clone for NativeWindow {
...
+        rwh_05::RawWindowHandle::AndroidNdk(handle)
+    }
+}
+
+#[cfg(feature = "rwh_06")]
+impl rwh_06::HasWindowHandle for NativeWindow {
+    fn window_handle(&self) -> Result<rwh_06::WindowHandle<'_>, rwh_06::HandleError> {
+        let handle = rwh_06::AndroidNdkWindowHandle::new(self.ptr.cast());
+        let handle = rwh_06::RawWindowHandle::AndroidNdk(handle);
+        // SAFETY: All fields of the "raw"  `AndroidNdkWindowHandle` struct are filled out
```
notgull, line 55: Nit: you should add a safety comment about how the handle is guaranteed to be valid.

notgull (approved):

ndk/src/native_window.rs
MarijnS95: I expected to have to deal with this, but the documentation only seems to discuss nullability of the handles since such "validity claims" should be "irrelevant" given the lifetime?

https://docs.rs/raw-window-handle/latest/raw_window_handle/struct.WindowHandle.html#safety

Unless I'm misunderstanding your docs, which somehow seem to jump from nullability to vailidity of handles within the same sentence?

notgull: Yeah that's my bad. Just add something to the tune of "this pointer will always be valid."

MarijnS95: Sure, done.

MarijnS95:
@kchibisov perhaps you also want to take a look at this PR in conjunction with https://github.com/rust-windowing/winit/pull/3126?

kchibisov (changes requested):
In winit we kept the `raw-window-handle` `v0.4`, which is missing here. It would be nice to have it as well?

notgull (approved):

MarijnS95:
@kchibisov we never had it so I don't really feel like it adds any value unless there are compelling reasons to? Perhaps to keep the implementations in `winit` consistent?

MarijnS95:
In addition, what do folks think about having any of these features in the `default` set (e.g. the _latest_ `rwh_06`)?

kchibisov:
@MarijnS95 yes, just for consistency reasons. But I don't care that much if it's fine with you.

>In addition, what do folks think about having any of these features in the default set (e.g. the latest rwh_06)?

Whatever you prefer as a library writer. You could suggest updating to 0.6.0 if you do breaking changes anyway.

MarijnS95:
@kchibisov added. I'll think about the default features for a bit... `0.5` used to be available OOTB so I feel like that "please upgrade" nudge might work if we provide 0.6 by default as well. `default-features = false` is pretty easy.

kchibisov (approved):

PR #447, opened 2023-11-16
README: Move modularization comment down and use new blockquote syntax

MarijnS95:
The previous `> **Note**` syntax has been deprecated in favour of `> [!NOTE]` sytnax on November 14th (2 days ago), and newer block types were added: https://github.com/orgs/community/discussions/16925

I'm not quite sure if `IMPORTANT` carries the right load since `Crucial information necessary for users to succeed` is no longer really applicable; most users know where to find the crates and their names haven't changed on crates.io.

Finally, also mention `xbuild` being favoured over `cargo-apk` (even though there's _still_, after all these years some feature disparity).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,3 @@
-> **Note**
-> This repository was recently [modularized](https://github.com/rust-mobile/ndk/issues/372) and the following crates were split into separate repositories:
->
-> Crate | New Location | Notes
-> ------|--------------|------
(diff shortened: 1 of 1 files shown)


MarijnS95:
Rendered: https://github.com/rust-mobile/ndk/blob/9d8fc730c298526ec86190a983225f3888859a19/README.md

rib (approved):
lgtm

PR #452, opened 2023-11-20
ndk/media_format: Make all mutating functions take `self` by `&mut`

### LINKED ISSUE #451: Investigate `MediaFormat` mutability / lifetimes
Barely any of the `MediaFormat` functions from the "safe" `MediaCodec` PR #216 take `self` as `mut`, even though there are quite a few functions that return borrows from pointers which are very likely invalidated when the `MediaFormat` is modified in some way. Even "strict getters" have the ability to invalidate previously returned results and should be considered `mut` for that case.

[See for example `toString()`](https://developer.android.com/ndk/reference/group/media#amediaformat_tostring) which is mentioned in https://github.com/rust-mobile/ndk/pull/450:

> The returned string is owned by the format, and remains valid until the next call to toString, or until the format is deleted.

Fortunately this function is not directly exposed to users, only via `Display` which copies the resulting string via `write_str()` in the same function where `AMediaFormat_toString()` is called.

[`getString()`](https://developer.android.com/ndk/reference/group/media#amediaformat_getstring) has the same reentrancy limitation:

> The returned string is owned by the format, and remains valid until the next call to getString, or until the format is deleted.

[`getBuffer()`](https://developer.android.com/ndk/reference/group/media#amediaformat_getbuffer) strangely is not documented to be invalidated on repeated invocations, only when it (obviously) is removed from the `MediaFormat`:

> The returned data is owned by the format and remains valid as long as the named entry is part of the format.

Note that I'm implying for both strings and buffers that they are invalidated when the same key is set to a different string/buffer/value (and maybe even when _another_ key is changed).

---

Given these notions I _**think**_ we should make all `set` functions, _and `fn string()` `mut`.

CC @zarik5 @spencercw in case I'm missing something in my observation.

MarijnS95:
Fixes #451, CC @zarik5

All setters can inherently change the internals of the format, and should furthermore invalidate - or not be allowed while - borrowed strings or buffers (are) acquired from it.

Unforatunately [`getString()` is documented as]:

    The returned string is owned by the format, and remains valid until the next call to getString, or until the format is deleted.

Meaning that it is not reentrant, and invalidates any data previously acquired from `str()`; hence this getter is marked as `&mut self` too. The same is however not the case for `getBuffer()`.

[`getString()` is documented as]: https://github.com/rust-mobile/ndk/issues/451

### DIFF
--- a/ndk/CHANGELOG.md
+++ b/ndk/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Move `MediaFormat` from `media::media_codec` to its own `media::media_format` module. (#442)
 - media_format: Expose `MediaFormat::copy()` and `MediaFormat::clear()` from API level 29. (#449)
+- **Breaking:** media_format: Mark all `fn set_*()` and `fn str()` as taking `self` by `&mut`. (#452)
 - input_queue: Add `from_java()` constructor, available since API level 33. (#456)
 - event: Add `from_java()` constructors to `KeyEvent` and `MotionEvent`, available since API level 31. (#456)
 - event: Implement `SourceClass` `bitflag` and provide `Source::class()` getter. (#458)
--- a/ndk/src/media/media_format.rs
+++ b/ndk/src/media/media_format.rs
@@ -122,8 +122,10 @@ impl MediaFormat {
         .then(|| unsafe { slice::from_raw_parts(out_buffer.cast(), out_size) })
     }
 
+    /// The returned `&str` borrow is only valid until the next call to [`MediaFormat::str()`] for
+    /// the same key.
     #[doc(alias = "AMediaFormat_getString")]
-    pub fn str(&self, key: &str) -> Option<&str> {
+    pub fn str(&mut self, key: &str) -> Option<&str> {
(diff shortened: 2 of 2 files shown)


spencercw:
The doc comments on `buffer` and `str` seem a little redundant because the rules are enforced by the type system anyway. The `str` comment is also not strictly accurate because the string is only invalidated when you re-fetch the same key.[^1] Not suggesting we should add document this implementation detail, but I don't think the adapted NDK docs are adding much value.

Looks good otherwise.

[^1]: https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/media/ndk/NdkMediaFormat.cpp;l=232-258;drc=65d530890d50e7a39f29fc70c0b6e55b402fe10a

MarijnS95:
Fair enough for `buffer`, I mostly copied the NDK docs for the few (two) functions that actually have docs upstream.

I'd like to keep "something" for `fn str()` however, to justify to documentation readers _why_ a getter is borrowing `self` mutably.

spencercw:
Yes that makes sense. A `mut` getter does look a bit odd on the face of it.

PR #455, opened 2023-12-11
ndk: Requre all off-thread `dyn Fn*` callback types to implement `Send`

### LINKED ISSUE #454: `MediaCodec`/`ImageReader` callbacks not `Send`
Both `MediaCodec` and `ImageReader` use callbacks, such as `set_async_notify_callback`, but none of the functions are marked as `Send`.

The documentation for `MediaCodec::set_async_notify_callback` specifically states that the callbacks will be invoked on an NDK internal thread, so `Send` should be mandatory. 

For `ImageReader::set_image_listener` and `ImageReader::set_buffer_removed_listener` the documentation doesn't say anything, but from the looks of it they must run from an internal thread as well.

And after this we should see if making `MediaCodec` and `ImageReader` at least `Send` would make sense, as they are already internally shared between threads.

**EDIT**

While looking more thoroughly at the API, it may be necessary to also make `Image` `Send`, as the current recommended workflow for `ImageReader` is to acquire the image in the `image_listener` callback, and then dispatch it to another thread for processing. 

Keeping `Image` `!Send` while also making the callback `Send` basically means that all image processing must be done in the callback, which is not ideal.

MarijnS95:
Fixes #454

In all these cases the NDK either documents or implements the callback to be invoked from a single, separate thread.  For this to be allowed by Rust thread safety, the closures and their (moved) contents are effectively  "moved" to a different thread (`Send`) but not accessed concurrently (`Sync`), but the type definitions were never requiring this marker trait which could lead to undefined/invalid behaviour.

In addition some `Box`ed callbacks may have been dropped before a new callback is passed to the NDK, leading to a possible race condition. By storing the new callback _after_ registering it with the NDK, the previous callback is now dropped later to ensure such race conditions no longer happen (assuming AOSP callback invocations are properly synchronized with the setters).

A request has been filed to solidify this "behavioural contract" in the documentation where not done so yet: https://issuetracker.google.com/issues/300602767#comment12

### DIFF
--- a/ndk/CHANGELOG.md
+++ b/ndk/CHANGELOG.md
@@ -3,6 +3,16 @@
 - Move `MediaFormat` from `media::media_codec` to its own `media::media_format` module. (#442)
 - media_format: Expose `MediaFormat::copy()` and `MediaFormat::clear()` from API level 29. (#449)
 - **Breaking:** media_format: Mark all `fn set_*()` and `fn str()` as taking `self` by `&mut`. (#452)
+- **Breaking:** Require all `dyn Fn*` types to implement `Send` when the FFI implementation invokes them on a separate thread: (#455)
+  - `audio::AudioStreamDataCallback`;
+  - `audio::AudioStreamErrorCallback`;
+  - `media::image_reader::BufferRemovedListener`;
+  - `media::image_reader::ImageListener`;
+  - `media::media_codec::ErrorCallback`;
+  - `media::media_codec::FormatChangedCallback`;
+  - `media::media_codec::InputAvailableCallback`;
+  - `media::media_codec::OutputAvailableCallback`.
+- Drop previous `Box`ed callbacks _after_ registering new ones, instead of before. (#455)
 - input_queue: Add `from_java()` constructor, available since API level 33. (#456)
 - event: Add `from_java()` constructors to `KeyEvent` and `MotionEvent`, available since API level 31. (#456)
 - event: Implement `SourceClass` `bitflag` and provide `Source::class()` getter. (#458)
--- a/ndk/src/audio.rs
+++ b/ndk/src/audio.rs
@@ -500,9 +500,9 @@ impl fmt::Debug for AudioStreamBuilder {
 
 #[doc(alias = "AAudioStream_dataCallback")]
 pub type AudioStreamDataCallback =
-    Box<dyn FnMut(&AudioStream, *mut c_void, i32) -> AudioCallbackResult>;
+    Box<dyn FnMut(&AudioStream, *mut c_void, i32) -> AudioCallbackResult + Send>;
 #[doc(alias = "AAudioStream_errorCallback")]
-pub type AudioStreamErrorCallback = Box<dyn FnMut(&AudioStream, AudioError)>;
+pub type AudioStreamErrorCallback = Box<dyn FnMut(&AudioStream, AudioError) + Send>;
 
 impl AudioStreamBuilder {
     fn from_ptr(inner: NonNull<ffi::AAudioStreamBuilder>) -> Self {
@@ -643,7 +643,6 @@ impl AudioStreamBuilder {
     pub fn data_callback(mut self, callback: AudioStreamDataCallback) -> Self {
         let mut boxed = Box::new(callback);
         let ptr: *mut AudioStreamDataCallback = &mut *boxed;
-        self.data_callback = Some(boxed);
 
         unsafe extern "C" fn ffi_callback(
             stream: *mut ffi::AAudioStreamStruct,
@@ -671,6 +670,8 @@ impl AudioStreamBuilder {
                 ptr as *mut c_void,
             )
         };
+
+        self.data_callback = Some(boxed);
 
         self
     }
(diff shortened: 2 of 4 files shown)


MarijnS95:
@spencercw since you've been active around callbacks and seem to be using this crate, would you mind helping me out reviewing this and other PRs? I don't like "forcing" my own work into the codebase without at least one extra pair of eyes.

I've also been wondering why the NDK is mixing both `Box<Box<dyn Fn...>>` types together with `F: Fn` generics (and `extern "C" fn native_func<F: Fn...>() {}`), and whether some APIs should deregister their callbacks in `drop()` to prevent possible data races.

spencercw:
Looks like a reasonable change.

I'm not actively using this right now (changed jobs), but don't mind being an extra pair of eyes from time to time.

> I've also been wondering why the NDK is mixing both `Box<Box<dyn Fn...>>` types together with `F: Fn` generics

This does seem to be a little superfluous. I guess from an API point of view it means the API user doesn't need to box the callback themselves, but would be nice to be consistent across the codebase.

> and whether some APIs should deregister their callbacks in `drop()` to prevent possible data races.

Tricky. I think this is probably not necessary, but it depends on the underlying NDK API. For example, `AMediaCodec_setAsyncNotifyCallback` says 'Once the callback is unregistered or the codec is reset / released, the previously registered callback will not be called.' So hopefully we can assume that once `AMediaCodec_delete` has returned, it is safe to drop the corresponding Rust callback state.

MarijnS95:
> I'm not actively using this right now (changed jobs), but don't mind being an extra pair of eyes from time to time.

Thanks! That's unfortunate but I have quite a bunch of PRs open now. @rib appears to be inactive because of the newyears' holidays so I'll wait for that.

> > I've also been wondering why the NDK is mixing both `Box<Box<dyn Fn...>>` types together with `F: Fn` generics
> 
> This does seem to be a little superfluous. I guess from an API point of view it means the API user doesn't need to box the callback themselves, but would be nice to be consistent across the codebase.

I thought so too at first, until it hit me: fat pointers. The pointer coming from a `Box<dyn Fnxxx>` is a fat pointer with the vtable and storage for whatever is moved/referenced inside the closure:

```rust
let foo: Box<dyn Fn() -> u32> = Box::new(|| 1337);
let raw = dbg!(Box::into_raw(foo));
dbg!(std::mem::size_of_val(&raw));
```

Gives:

```
[src/main.rs:3] Box::into_raw(foo) = 0x0000000000000001
[src/main.rs:4] std::mem::size_of_val(&raw) = 16
```

(Note that nothing is captured, so there's no data associated with the `dyn Fn`)

So we'll need to box twice if we wish to pass an `8`-byte pointer around, or find another solution such as using generics to convey the closure type (hence the box pointer becomes `8`-bytes again, but the callback is also monomorphized).

---

> Tricky. I think this is probably not necessary, but it depends on the underlying NDK API. For example, `AMediaCodec_setAsyncNotifyCallback` says 'Once the callback is unregistered or the codec is reset / released, the previously registered callback will not be called.' So hopefully we can assume that once `AMediaCodec_delete` has returned, it is safe to drop the corresponding Rust callback state.

Yeah I've been giving this some thought:

1. For one-off callbacks (`Choreographer` seems to have some) we can `Box::into_raw()` in the callback. In the unlikely even that the callback is never called (there's no way to unregister it, and `Choreographer` is a "singleton" with a global "instance" and hence no `impl Drop`), we'll leak a box;
2. For functions called more often, we'll hold on to the `Box` and destroy it on `drop()` (and eventually deregister it if necessary, as discussed before - but let's check docs first);
3. If there's ever a structure that can be "cloned" (in this crate) via incrementing the NDK refcount, we suddenly no longer have a sensible place to track the "lifetime" of a boxed callback (but we'll investigate/solve that when the case appears, I don't think we have any ATM).

Either way, deregistering callbacks or having an explicit comment that the `delete()` causes the callback to no longer fire is fine by me; as long as we can safely get rid of the `Box` and invalidate the pointer.

---

Then secondly I've been thinking about a general lack of when and where callbacks are called, and what functions/objects/pointers are thread-safe, but that's for another time :)

MarijnS95:
Made an attempt to request some clarifications in https://issuetracker.google.com/issues/318944941.

Notice that it may be terrible for us to take an `Fn` instead of an `FnMut`. The latter allows interior mutability of captured variables and should _not be used on concurrently-called closures_ AFAIK. However, I think it may be fine on these functions as they're moved to another thread and only invoked serially there.

Thoughts?

spencercw:
I think `FnMut` makes sense for `AMediaCodec`, with `FnOnce` for `AChoreographer` and `ASurfaceTransaction`. This does have the built-in assumption that these callbacks indeed are only called once, but the NDK documentation being what it is we might just have to live with this and hope at some point they improve it.

MarijnS95:
@spencercw I ended up opening a new issue #464 to once again go over all callbacks and document their (lack of) standardization, so that we have a guideline how to build our callbacks going forward.

On the topic of `Box<Box<dyn Fn>>` vs `Box<T>` with `T: Fn`, the latter can only be implemented for one-off callbacks that don't store their callback in the owning object.

PR #456, opened 2023-12-23
Add `from_java()` constructors to `InputQueue`, `KeyEvent` and `MotionEvent`

MarijnS95:
These were added in API 31 (`KeyEvent`/`MotionEvent`) and API 33 (`InputEvent`), but not yet mapped in the NDK crate.

### DIFF
--- a/ndk/src/event.rs
+++ b/ndk/src/event.rs
@@ -6,12 +6,15 @@
 //! [`android.view.KeyEvent`].
 //!
 //! [`AInputEvent`, `AKeyEvent` and `AMotionEvent`]: https://developer.android.com/ndk/reference/group/input
-//! [`android.view.InputEvent`]: https://developer.android.com/reference/android/view/InputEvent.html
-//! [`android.view.MotionEvent`]: https://developer.android.com/reference/android/view/MotionEvent.html
+//! [`android.view.InputEvent`]: https://developer.android.com/reference/android/view/InputEvent
+//! [`android.view.MotionEvent`]: https://developer.android.com/reference/android/view/MotionEvent
 //! [`android.view.KeyEvent`]: https://developer.android.com/reference/android/view/KeyEvent
 
-use num_enum::{IntoPrimitive, TryFromPrimitive};
 use std::ptr::NonNull;
+
+#[cfg(feature = "api-level-31")]
+use jni_sys::{jobject, JNIEnv};
+use num_enum::{IntoPrimitive, TryFromPrimitive};
 
 /// A native [`AInputEvent *`]
 ///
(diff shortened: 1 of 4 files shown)


rib (approved):
Nice, these look good to me.

As commented, I'd be tempted to avoid exposing `jni_sys` types in the public API but I think either way is probably OK.

ndk/src/input_queue.rs
```diff
@@ -6,6 +6,9 @@ use std::io::Result;
 use std::os::raw::c_int;
 use std::ptr::{self, NonNull};
 
+#[cfg(feature = "api-level-33")]
+use jni_sys::{jobject, JNIEnv};
```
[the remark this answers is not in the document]
MarijnS95: Sure, let's think about it a bit and do that in a followup https://github.com/rust-mobile/ndk/labels/impact%3A%20breaking change :+1:

MarijnS95:
Force-pushed to fix some wrong intradoc links :)

ndk/src/event.rs
```diff
@@ -373,7 +398,26 @@ impl MotionEvent {
...
 
-    /// Returns a pointer to the native [`ffi::AInputEvent`]
+    /// Creates a native [`InputEvent`] object that is a copy of the specified
+    /// Java [`android.view.MotionEvent`]. The result may be used with generic and
+    /// [`MotionEvent`]-specific functions.
+    ///
+    /// # Safety
+    ///
+    /// This function should be called with a healthy JVM pointer and with a non-null
+    /// [`android.view.MotionEvent`], which must be kept alive on the Java/Kotlin side.
```
MarijnS95, line 408: @rib I copied the `which must be kept alive on the Java/Kotlin side.` portion from other `from_java()` functions, but I don't think it holds true for these APIs.

1. As the copied doc-comment above suggests: `... that is a copy of the specified Java ...`;
2. And `AInputEvent_release()` which must be called states: `The underlying Java object remains valid and does not change its state.`.

It seems the two objects are independent and have their own lifetime after this call.

If you agree I'll remove this text.

rib: Hmm, I'm not sure tbh - from when I last tried to get my head around the implementation details for input event handling I recall some pretty funky lifetime tracking for events that went via an `InputQueue` since there's some special callbacks in java for events after they are handled by native code from what I vaguely recall and I don't really know how the bits all fit together if using these without really digging into the implementation.

My guess would be that you do actually need to ensure the event is kept alive in the jvm if they only hold a weak reference in the native event and guess that the release isn't just about freeing the struct but also tracking when the event is finished with on the jvm side so that it can progress to the next stage of being handled in the jvm.

I'd er on the side of keeping that requirement in the docs unless we really look into the implementation details more closely.

rib (approved):

ndk/src/event.rs
MarijnS95: It constructs it here:

https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/native/android/input.cpp;l=90;drc=f5c06a171c6eb51c6c914540b36a8211036fe735

Which copies over a bunch of fields:

https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/jni/android_view_KeyEvent.cpp;l=116-139;drc=ea4a158fa1e4620658dab38305ddde79700843f7

All information is copied as raw `jint`s by value, and I don't think the various `id` fields have a back-reference to the Java representation.

---

Likewise the motion events are fully copied:

https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/native/android/input.cpp;l=312-314;drc=f5c06a171c6eb51c6c914540b36a8211036fe735

https://cs.android.com/android/platform/superproject/main/+/main:frameworks/native/libs/input/Input.cpp;l=547-580;drc=ea4a158fa1e4620658dab38305ddde79700843f7

With nothing that appears to be a reference.

---

Let's remove the doc text.

MarijnS95: > and guess that the release isn't just about freeing the struct but also tracking when the event is finished with on the jvm side so that it can progress to the next stage of being handled in the jvm.

To confirm, the terribly named `_release` (why not `_destroy`... release sounds like refcounting) function only calls `delete` and doesn't perform any "event finished" handling later. That, after all, happens elsewhere via `AInputQueue_finishEvent()` which has a reference to the input queue and exists specifically for explicitness of this purpose.

https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/native/android/input.cpp;l=57-59;drc=f5c06a171c6eb51c6c914540b36a8211036fe735

---

Looks like I already removed the wording in the last force-push, let's merge this.

PR #458, opened 2023-12-25
ndk/event: Implement `SourceClass` `bitflag` and provide `Source::class()` getter

MarijnS95:
This `enum` was never `pub` even though it serves a slight purpose to more easily classify various event sources, which is accounted for in the bitmask-like value of `AINPUT_SOURCE_XXX`.

### DIFF
--- a/ndk/src/event.rs
+++ b/ndk/src/event.rs
@@ -72,18 +72,28 @@ pub enum Source {
     Any = ffi::AINPUT_SOURCE_ANY,
 }
 
-/// An enum representing the class of an [`InputEvent`] source.
-///
-/// See [the NDK docs](https://developer.android.com/ndk/reference/group/input#anonymous-enum-35)
-#[derive(Debug, Clone, Copy, PartialEq, Eq, TryFromPrimitive, IntoPrimitive)]
-#[repr(u32)]
-enum Class {
-    None = ffi::AINPUT_SOURCE_CLASS_NONE,
-    Button = ffi::AINPUT_SOURCE_CLASS_BUTTON,
-    Pointer = ffi::AINPUT_SOURCE_CLASS_POINTER,
-    Navigation = ffi::AINPUT_SOURCE_CLASS_NAVIGATION,
-    Position = ffi::AINPUT_SOURCE_CLASS_POSITION,
-    Joystick = ffi::AINPUT_SOURCE_CLASS_JOYSTICK,
+impl Source {
+    pub fn class(self) -> SourceClass {
+        let class = u32::from(self) & ffi::AINPUT_SOURCE_CLASS_MASK;
+        SourceClass::from_bits_retain(class)
+    }
+}
+
+bitflags::bitflags! {
+    /// Flags representing the class of an [`InputEvent`] [`Source`].
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    pub struct SourceClass : u32 {
+        #[doc(alias = "AINPUT_SOURCE_CLASS_BUTTON")]
+        const BUTTON = ffi::AINPUT_SOURCE_CLASS_BUTTON;
+        #[doc(alias = "AINPUT_SOURCE_CLASS_POINTER")]
+        const POINTER = ffi::AINPUT_SOURCE_CLASS_POINTER;
+        #[doc(alias = "AINPUT_SOURCE_CLASS_NAVIGATION")]
+        const NAVIGATION = ffi::AINPUT_SOURCE_CLASS_NAVIGATION;
+        #[doc(alias = "AINPUT_SOURCE_CLASS_POSITION")]
+        const POSITION = ffi::AINPUT_SOURCE_CLASS_POSITION;
+        #[doc(alias = "AINPUT_SOURCE_CLASS_JOYSTICK")]
+        const JOYSTICK = ffi::AINPUT_SOURCE_CLASS_JOYSTICK;
+    }
 }
 
 impl InputEvent {
--- a/ndk/src/event.rs
+++ b/ndk/src/event.rs
@@ -75,24 +75,28 @@ pub enum Source {
 impl Source {
     pub fn class(self) -> SourceClass {
         let class = u32::from(self) & ffi::AINPUT_SOURCE_CLASS_MASK;
-        SourceClass::from_bits_retain(class)
+        // The mask fits in a u8.
+        SourceClass::from_bits_retain(class as u8)
     }
 }
 
 bitflags::bitflags! {
     /// Flags representing the class of an [`InputEvent`] [`Source`].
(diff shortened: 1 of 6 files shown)


ndk/src/event.rs
```diff
@@ -47,12 +47,20 @@ pub enum Source {
     Any = ffi::AINPUT_SOURCE_ANY,
 }
 
+impl Source {
+    pub fn class(self) -> SourceClass {
+        let class = u32::from(self) & ffi::AINPUT_SOURCE_CLASS_MASK;
+        // TODO: Don't use None
+        class.try_into().unwrap_or(SourceClass::None)
```
MarijnS95, line 54: I'll merge it like this and clean this up in a followup PR that revises all `enum` types.

ndk/src/event.rs
```diff
@@ -47,12 +47,20 @@ pub enum Source {
...
+    }
+}
+
 /// An enum representing the class of an [`InputEvent`] source.
 ///
 /// See [the NDK docs](https://developer.android.com/ndk/reference/group/input#anonymous-enum-35)
 #[derive(Debug, Clone, Copy, PartialEq, Eq, TryFromPrimitive, IntoPrimitive)]
 #[repr(u32)]
-enum Class {
+pub enum SourceClass {
```
rib, line 63: It's slightly awkward / inconsistent that `_SOURCE_CLASS_NAVIGATION` is named `_SOURCE_CLASS_TRACKBALL` in the SDK, and that's the name used in `android-activity`. Probably makes sense to go with the NDK naming here though.

In general though, the SourceClass is a bitmask and I'm not sure that the class is strictly mutually exclusive. In `android-activity` this is bound as a `SourceFlags` bitmask but only exposed via `Source` via methods like `Source::is_button_class()` - I'd maybe be tempted to do something similar here too.

MarijnS95: Yeah, "navigation" doesn't mean much. Do you want me to solve this with an extra doc alias?

MarijnS95: You are totally light, this must be `bitflags`!

C definition:

```c
/**
 * Input source masks.
 *
 * Refer to the documentation on android.view.InputDevice for more details about input sources
 * and their correct interpretation.
 */
enum {
    /** mask */
    AINPUT_SOURCE_CLASS_MASK = 0x000000ff,

    /** none */
    AINPUT_SOURCE_CLASS_NONE = 0x00000000,
    /** button */
    AINPUT_SOURCE_CLASS_BUTTON = 0x00000001,
    /** pointer */
    AINPUT_SOURCE_CLASS_POINTER = 0x00000002,
    /** navigation */
    AINPUT_SOURCE_CLASS_NAVIGATION = 0x00000004,
    /** position */
    AINPUT_SOURCE_CLASS_POSITION = 0x00000008,
    /** joystick */
    AINPUT_SOURCE_CLASS_JOYSTICK = 0x00000010,
};
```

ndk/src/event.rs
```diff
@@ -47,18 +47,30 @@ pub enum Source {
     Any = ffi::AINPUT_SOURCE_ANY,
 }
 
-/// An enum representing the class of an [`InputEvent`] source.
-///
-/// See [the NDK docs](https://developer.android.com/ndk/reference/group/input#anonymous-enum-35)
```
MarijnS95, line 52: I've stripped off this doc because it's been renumbered upstream and points nowhere.

ndk/src/event.rs
```diff
@@ -47,18 +47,30 @@ pub enum Source {
...
-    Position = ffi::AINPUT_SOURCE_CLASS_POSITION,
-    Joystick = ffi::AINPUT_SOURCE_CLASS_JOYSTICK,
+impl Source {
+    pub fn class(self) -> SourceClass {
+        let class = u32::from(self) & ffi::AINPUT_SOURCE_CLASS_MASK;
+        SourceClass::from_bits_retain(class)
+    }
+}
+
+bitflags::bitflags! {
```
rib, line 57: If the bitflags API is exposed directly I wonder if we should add a a compatibility member like `const _ = !0;` so `all()` wouldn't assume these flags won't be extended and would set all bits:

> [Externally defined flags](https://docs.rs/bitflags/latest/bitflags/#externally-defined-flags)
> If you’re generating flags types for an external source, such as a C API, you can define an extra unnamed flag as a  mask of all bits the external source may ever set. Usually this would be all bits (!0):

I guess we could remove the `AINPUT_SOURCE_CLASS_NONE` which is zero - so not actually a flag. Instead `::empty()` and `::is_empty()` can be used from the bitflags api.

MarijnS95: That's why we haven't used `bitflags` in the `ash` vulkan crate, and even removed the `all()` and `not()` inversion methods a long time ago. That was also because `not()` was implemented as `self ^ all()`, so only known-at-compile-time bits get flipped. Fortunately `bitflags` implements it differently.

Let's stick with the described workaround.

Also yeah, I've been thinking about the `NONE` flag, which now has an alias and is represented upstream.

rib (approved):
cool, all looking good to me now!

rib (changes requested):
oh, actually, maybe we shouldn't use `!0` but should use `AINPUT_SOURCE_CLASS_MASK` ?

ndk/src/event.rs
```diff
@@ -47,18 +47,32 @@ pub enum Source {
...
+        const POINTER = ffi::AINPUT_SOURCE_CLASS_POINTER as u8;
+        #[doc(alias = "AINPUT_SOURCE_CLASS_NAVIGATION")]
+        const NAVIGATION = ffi::AINPUT_SOURCE_CLASS_NAVIGATION as u8;
+        #[doc(alias = "AINPUT_SOURCE_CLASS_POSITION")]
+        const POSITION = ffi::AINPUT_SOURCE_CLASS_POSITION as u8;
+        #[doc(alias = "AINPUT_SOURCE_CLASS_JOYSTICK")]
+        const JOYSTICK = ffi::AINPUT_SOURCE_CLASS_JOYSTICK as u8;
+
+        // https://docs.rs/bitflags/latest/bitflags/#externally-defined-flags
+        const _ = !ffi::AINPUT_SOURCE_CLASS_MASK as u8;
```
rib, line 74: Er, I don't think we want to negate this right? I think the `_MASK` is `0xff` and we want a value with all bits set here.

The alternative of going back to `!0` now that this is `u8` also sounds like it would be ok here - but either way is fine.

rib (changes requested):
Looking good except I think there's a spurious negate for the `const _` value for `SourceClass` flags.

ndk/src/event.rs
MarijnS95: Editor detritus, you're totally right. Think I pasted/replaced this constant while forgetting to remove the `!`.

PR #461, opened 2023-12-25
ndk/hardware_buffer: Convert `HardwareBufferUsage` to `bitflags`

MarijnS95:
`HardwareBufferUsage` really is a structure that comprises various usage flags, and was modeled as a hard-to-get `u64` value while being wrapped inside _two_ newtypes (one in the `ndk`, one in `ndk-sys`) and without the typical bit-ops (i.e. `BitOr`) one comes to expect.

This is all provided out of the box - and more helper functions - via `bitflags` with more convenient access to the raw value to beat.

Also inherit the upstream docs for convenience and completeness.


rib (approved):
Looks good to me

PR #465, opened 2024-01-15
ndk/native_window: Link against `libnativewindow` for API >= 26 functions

MarijnS95:
Per [the `libandroid` function map][1] newer `NativeWindow` functions are no longer available; these instead need to be linked [from the `libnativewindow` library][2] directly.

[1]: https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/native/android/libandroid.map.txt;l=1?q=libandroid.map.txt&sq=&ss=android%2Fplatform%2Fsuperproject%2Fmain
[2]: https://cs.android.com/android/platform/superproject/main/+/main:frameworks/native/libs/nativewindow/libnativewindow.map.txt;l=1?q=libnativewindow.map.txt&sq=&ss=android%2Fplatform%2Fsuperproject%2Fmain

Ref: https://github.com/android/ndk/issues/1982

### DIFF
--- a/ndk/src/native_window.rs
+++ b/ndk/src/native_window.rs
@@ -7,7 +7,7 @@ use std::{ffi::c_void, io, mem::MaybeUninit, ptr::NonNull};
 use jni_sys::{jobject, JNIEnv};
 
 use super::{hardware_buffer_format::HardwareBufferFormat, utils::status_to_io_result};
-#[cfg(feature = "api-level-28")]
+#[cfg(all(feature = "nativewindow", feature = "api-level-28"))]
 use crate::data_space::DataSpace;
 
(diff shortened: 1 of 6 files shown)


ndk/src/native_window.rs
```diff
@@ -377,7 +378,7 @@ impl<'a> Drop for NativeWindowBufferLockGuard<'a> {
     }
 }
 
-#[cfg(feature = "api-level-26")]
+#[cfg(all(feature = "nativewindow", feature = "api-level-26"))]
```
MarijnS95, line 381: I'm just realizing that I'll revert this and the change for the two `enum`s below for `SurfaceTransaction` again (in an upcoming PR), as these are exposed directly from `libandroid`.

PR #469, opened 2024-01-23
looper: Require `Send` when adding fd event callbacks on `ForeignLooper`

MarijnS95:
`ForeignLooper` allows (re)registering callbacks on threads that are not the `ThreadLooper` (current) thread.  Since these calls will be executed on that `ThreadLooper` thread when the user calls any of the `poll()` functions, these closures have to pass a thread boundary which requires `Send`.

### DIFF
--- a/ndk/CHANGELOG.md
+++ b/ndk/CHANGELOG.md
@@ -24,6 +24,7 @@
 - native_window: Require linking against `libnativewindow` for most API >= 26 functions. (#465)
 - **Breaking:** audio: Merge `AudioResult` variant enum into `AudioError`. (#467)
 - data_space: Add missing `DataSpaceRange::Unspecified` variant. (#468)
+- **Breaking:** looper: Require `Send` marker when adding fd event callbacks on `ForeignLooper`. (#469)
 
 # 0.8.0 (2023-10-15)
 
--- a/ndk/src/looper.rs
+++ b/ndk/src/looper.rs
@@ -211,6 +211,42 @@ impl ThreadLooper {
         )
     }
 
+    /// Adds a file descriptor to be polled, with a callback that is invoked when any of the
+    /// [`FdEvent`]s described in `events` is triggered.
+    ///
+    /// The callback receives the file descriptor it is associated with and a bitmask of the poll
+    /// events that were triggered (typically [`FdEvent::INPUT`]).  It should return [`true`] to
+    /// continue receiving callbacks, or [`false`] to have the callback unregistered.
+    ///
+    /// See also [the NDK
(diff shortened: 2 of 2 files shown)


ndk/src/looper.rs
```diff
@@ -209,6 +209,72 @@ impl ThreadLooper {
...
+    /// See also [the NDK
+    /// docs](https://developer.android.com/ndk/reference/group/looper.html#alooper_addfd).
+    ///
+    /// Note that this will leak a [`Box`] unless the callback returns [`false`] to unregister
+    /// itself.
+    ///
+    /// # Threading
+    /// This function will be called on the current thread when this [`ThreadLooper`] is
+    /// polled. A callback can also be registered from other threads via the equivalent
+    /// [`ThreadLooper::add_fd_with_callback()`] function, which requires a [`Send`] bound.
```
rib, line 228:
```suggestion
    /// [`ForeignLooper::add_fd_with_callback()`] function, which requires a [`Send`] bound.
```

rib (approved):
Nice, this looks good to me!

ndk/src/looper.rs
```diff
@@ -209,6 +209,72 @@ impl ThreadLooper {
...
+    /// callback.
+    #[doc(alias = "ALooper_addFd")]
+    pub fn add_fd_with_callback<F: FnMut(BorrowedFd<'_>, FdEvent) -> bool>(
+        &self,
+        fd: BorrowedFd<'_>,
+        events: FdEvent,
+        callback: F,
+    ) -> Result<(), LooperError> {
+        // let callback = unsafe { std::mem::transmute(callback) };
+        // self.foreign.add_fd_with_callback(fd, events, callback)
```
MarijnS95, line 243: A transmute doesn't seem to work here to deduplicate the two functions, unfortunately :/

MarijnS95:
Found a way, but I think `add_fd_with_callback_assume_send()` should be `unsafe`.

ndk/CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 - native_window: Require linking against `libnativewindow` for most API >= 26 functions. (#465)
 - **Breaking:** audio: Merge `AudioResult` variant enum into `AudioError`. (#467)
 - data_space: Add missing `DataSpaceRange::Unspecified` variant. (#468)
+- **Breaking:** looper: Require `Send` marker when adding fd event callbacks on `ForeignLooper`. (#469)
```
MarijnS95, line 27: As soon as I click merge, I realized I should have probably mentioned that `ThreadLooper::add_fd_with_callback()` was added without the `Send` bound, if already on the same thread.

PR #470, opened 2024-02-28
Expose internal pointer to AImage

ShootingKing-AM:
- needed for doing some direct operation on AImage ptr (maybe unsafe)
- but anyways i think the library need not enforce AImage ptr to be hidden


MarijnS95:
Before doing this, I'd like to consistenize the current split between `fn as_ptr(&self) -> *mut T` versus `fn ptr(&self) -> NonNull<T>`, as well as the documentation comments and safety constraints.

PR #475, opened 2024-05-14
ndk-sys: Drop cfg for inexistant `target_arch = "armv7"`

MarijnS95:
[Rust nightly since May 6 2024] points out that `armv7` is not a known, valid value for the `target_arch` cfg variable.  This is confirmed by the docs not listing it either: https://doc.rust-lang.org/reference/conditional-compilation.html#target_arch

Hence drop this entirely, and rely purely on `target_arch = "arm"`.

[Rust nightly since May 6 2024]: https://blog.rust-lang.org/2024/05/06/check-cfg.html


MarijnS95:
According to #10 `armv7` isn't a `target_arch` either. The issue was closed without being fixed in the `ndk-sys` crate.

PR #479, opened 2024-08-03
ndk: Add definitions and bindings for API levels

MarijnS95:
Depends on #495

`android/api-levels.h` from the NDK defines constants for all (NDK-supported) API levels and both a getter for the "current apps' target API level" as well as the API level of the device the app is running on.  Create bindings to allow users to query this information at runtime from their Rust code as well.

The equivalent of the NDK's `__ANDROID_API__` define are our `api-level-xx` crate features, used to restrict APIs (and inexistant linker symbols) at build-time.
(As there is no full-fledged Android build system involved, users of the  `ndk` crate are expected to keep this in sync with the minimum/target  API level passed to their build tool of choice.)

### DIFF
--- a/ndk-sys/CHANGELOG.md
+++ b/ndk-sys/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 - Regenerate bindings with `bindgen 0.71.1`. (#487)
+- Include API bindings from `sys/system_properties.h`. (#495)
 
 # 0.6.0 (2024-04-26)
 
--- a/ndk-sys/CHANGELOG.md
+++ b/ndk-sys/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- **Breaking:** Removed `__ANDROID_API__` constant as it is always defined to the value of `__ANDROID_API_FUTURE__`. (#479)
 - Regenerate bindings with `bindgen 0.71.1`. (#487)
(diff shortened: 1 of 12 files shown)


MarijnS95:
CC @daxpedda

MarijnS95:
```
error: struct `crash_detail_t` is never constructed
     --> ndk-sys/src/ffi_aarch64.rs:16024:12
      |
16024 | pub struct crash_detail_t {
      |            ^^^^^^^^^^^^^^
      |
      = note: `-D dead-code` implied by `-D warnings`
      = help: to override `-D warnings` add `#[allow(dead_code)]`

error: struct `ACameraOutputTargets` is never constructed
     --> ndk-sys/src/ffi_aarch64.rs:21233:12
      |
21233 | pub struct ACameraOutputTargets {
      |            ^^^^^^^^^^^^^^^^^^^^
```

I think this is a Rust 1.80 clippy false-positive. The types may not be constructed by us, but they are `pub` visible and could be used in downstream code.

However, no API currently takes these symbols, so we might as well skip & hide them until they become useful?

(`ACameraOutputTarget` without `-s` does exist and is used)

MarijnS95:
Re-running CI jobs because some changes to `dead_code` causing these false-positives appear to have been reverted in 1.80.1: https://blog.rust-lang.org/2024/08/08/Rust-1.80.1.html#false-positives-in-the-dead_code-lint

PR #480, opened 2024-08-03
Add bindings for Performance Hint Manager

MarijnS95:
https://developer.android.com/ndk/reference/group/a-performance-hint

Performance Hint Manager allows apps to inform the system about workloads running on threads to more accurately allocate power for them. After the fact, applications should report back how long their operation actually ended up taking.

The new work duration API (driven by `AWorkDuration`) also includes GPU timings.

### DIFF
--- a/ndk/src/performance_hint.rs
+++ b/ndk/src/performance_hint.rs
@@ -1,0 +1,331 @@
+//! Bindings for [`APerformanceHintManager`] and [`APerformanceHintSession`]
+//!
+//! `APerformanceHint` allows apps to create performance hint sessions for groups of
+//! threads, and provide hints to the system about the workload of those threads, to help the
+//! system more accurately allocate power for them. It is the NDK counterpart to the [Java
+//! PerformanceHintManager SDK API].
+//!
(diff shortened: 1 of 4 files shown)


ndk/src/performance_hint.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+impl PerformanceHintManager {
+    /// Retrieve a reference to the performance hint manager.
+    ///
+    /// Returns [`None`] on failure.
+    #[doc(alias = "APerformanceHint_getManager")]
+    pub fn new() -> Option<Self> {
+        NonNull::new(unsafe { ffi::APerformanceHint_getManager() }).map(|ptr| Self { ptr })
+    }
```
MarijnS95, line 38: As it turns out there was a hidden `static` keyword in this C function, this should always return the same singleton for the process:

https://issuetracker.google.com/issues/357214902#comment7

Meaning we can also derive `Send`/`Sync` and `Copy`/`Clone` since the user could anyway cheaply retrieve the same singleton from any thread.

PR #481, opened 2024-08-03
Add bindings for Thermal Manager

MarijnS95:
The header for this got [fixed just yesterday](https://android-review.googlesource.com/c/platform/frameworks/native/+/3205910) to contain valid C so that we can parse it (without making trivial custom changes...) so now seems to be about the right time to actually write out the bindings and at least have them open in a PR!

---

https://developer.android.com/ndk/reference/group/thermal

`AThermal` allows querying the current thermal (throttling) status, as well as forecasts of future thermal statuses to allow applications to respond and mitigate possible throttling in the (near) future.

### DIFF
--- a/ndk/src/thermal.rs
+++ b/ndk/src/thermal.rs
@@ -1,0 +1,331 @@
+//! Bindings for [`AThermalManager`]
+//!
+//! Structures and functions to access thermal status and register/unregister thermal status
+//! listener in native code.
+//!
+//! [`AThermalManager`]: https://developer.android.com/ndk/reference/group/thermal#athermalmanager
+#![cfg(feature = "api-level-30")]
+
+#[cfg(doc)]
+use std::io::ErrorKind;
+use std::{io::Result, os::raw::c_void, ptr::NonNull};
+
+use num_enum::{FromPrimitive, IntoPrimitive};
+
+use crate::utils::abort_on_panic;
+
+/// Workaround for <https://issuetracker.google.com/issues/358664965>.  `status_t` should only
+/// contain negative error codes, but the underlying `AThermal` implementation freely passes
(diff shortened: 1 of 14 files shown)


ndk/src/thermal.rs
```diff
@@ -0,0 +1,331 @@
...
+
+/// Workaround for <https://issuetracker.google.com/issues/358664965>.  `status_t` should only
+/// contain negative error codes, but the underlying `AThermal` implementation freely passes
+/// positive error codes around. At least the expected return codes are "implicitly" documented to
+/// be positive.
+fn status_to_io_result(status: i32) -> Result<()> {
+    // Intentionally not imported in scope (and an identically-named function) to prevent
+    // accidentally calling this function without negation.
+    crate::utils::status_to_io_result(-status)
+}
```
MarijnS95, line 25: Unlike all other NDK APIs, this one returns _positive_ errno values :upside_down_face: 

https://issuetracker.google.com/issues/358664965

ndk/src/thermal.rs
```diff
@@ -0,0 +1,331 @@
...
+        value.0.into()
+    }
+}
+
+/// Prototype of the function that is called when thermal status changes. It's passed the updated
+/// thermal status as parameter.
+///
+/// # Warning
+/// [`ThermalManager`] is synchronized internally, and its lock is held while this callback is
+/// called.  Interacting with [`ThermalManager`] inside this closure *will* result in a deadlock.
```
MarijnS95, line 77: While all the extra locking makes it possible to add `Send`+`Sync` on `ThermalManager`, it means the user can now trivially pass this type into the callback and (if it's the same instance!) get a deadlock when interacting with it.

PR #483, opened 2024-08-10
ndk: Implement common traits where sensible, and drop some `Ord`

MarijnS95:
Provided traits in the `ndk` crate are all over the place.  Some pointer wrappers and regular enumerations derive `(Partial)Ord` even though there is no sense in using an ordering for these types.  Others don't derive `(Partial)Eq` and `Hash` which makes it hard to compare if objects are the same (by-pointer) or to store them inside i.e. `HashMap`.  Deriving these types follows Rust's [C-COMMON-TRAITS] convention.

Additionally, sort `derives` by their relation, followed by sorting them alphabetically.

[C-COMMON-TRAITS]: https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits

## TODO
- [ ] Enable deriving some traits in `ndk-sys` to be able to implement them on structures that own `ffi` structures;
- [ ] Derive comparison traits on objects that keep callbacks for lifetime purposes?
- [ ] Omit derives on `Iter` wrappers?
- [ ] Have `Ord` on pointer wrappers after all, to store them in i.e. `BTreeMap` even though their ordering has no useful definition?

CC @daxpedda for suggestions based on https://github.com/rust-windowing/winit/pull/3796 :)


ndk/src/media/media_codec.rs
```diff
@@ -26,7 +26,7 @@ pub enum MediaCodecDirection {
 /// A native [`AMediaCodec *`]
 ///
 /// [`AMediaCodec *`]: https://developer.android.com/ndk/reference/group/media#amediacodec
-#[derive(Debug)]
```
daxpedda, line 29: No clue about this API, but if there aren't multiple `MediaCodec` with  the same `AMediaCodec`, then you could implement `Eq`/`Hash` by hand over just that and not use the callback.

MarijnS95: `AMediaCodec` doesn't implement refcounting semantics so there should always be exactly one `MediaCodec` for a given `AMediaCodec` pointer, so this is definitely possible here.

That is also why we can lifetime the callbacks within the wrapper object in the first place (as we deduced is possible in for example Thermal Manager as well, but not in Performance Manager which is a global `static`).

ndk/src/asset.rs
```diff
@@ -14,7 +14,7 @@ use std::{
 /// A native [`AAssetManager *`]
 ///
 /// [`AAssetManager *`]: https://developer.android.com/ndk/reference/group/asset#aassetmanager
-#[derive(Debug)]
+#[derive(Debug, PartialEq, Eq, Hash)]
 #[doc(alias = "AAssetManager")]
 pub struct AssetManager {
```
MarijnS95, line 19: Can be `Clone`/`Copy` since this is a cheap reference/handle from `Activity` (but should carry a lifetime on `Activity` realistically).

PR #486, opened 2024-09-03
ndk/event: Replace open-coded `bool` getters with `bitflags` wrappers

MarijnS95:
And "correct" their ABI to match how it is used in the actual NDK getter functions.  Also add missing aliases and upstream documentation where sensible.

---

CC @rib, I've been thinking about this to make our wrapper types more consistent, but it may be more inconvenient to use now except in the case when checking multiple flags at once...

### DIFF
--- a/ndk/src/event.rs
+++ b/ndk/src/event.rs
@@ -86,7 +86,7 @@ impl Source {
 
 bitflags::bitflags! {
     /// Flags representing the class of an [`InputEvent`] [`Source`].
-    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
     pub struct SourceClass : u8 {
         #[doc(alias = "AINPUT_SOURCE_CLASS_BUTTON")]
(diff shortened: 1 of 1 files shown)


ndk/src/event.rs
```diff
@@ -335,72 +321,59 @@ pub enum ToolType {
...
-        self.0 & ffi::AMOTION_EVENT_BUTTON_PRIMARY != 0
-    }
-    #[inline]
-    pub fn secondary(self) -> bool {
-        self.0 & ffi::AMOTION_EVENT_BUTTON_SECONDARY != 0
-    }
-    #[inline]
-    pub fn teriary(self) -> bool {
-        self.0 & ffi::AMOTION_EVENT_BUTTON_TERTIARY != 0
-    }
```
MarijnS95, line 354: Note that this PR will fix the `teriary` -> `TERTIARY` typo as pointed out in https://github.com/rust-mobile/android-activity/pull/163, by freshly copying the constants from the header file.

MarijnS95:
We should also do this for `ndk::configuration::DiffResult`, which most likely benefits from the autogenerated `Debug` implementation too.

PR #487, opened 2024-09-03
sys: Regenerate with `bindgen 0.71.1`

MarijnS95:
This can now emit compile-time assertions rather than runtime `#[test]`s, but only if the `--rust-target` is set to `1.77` or higher. We leave it at 1.60 for now to not have to bump our MSRV too aggressively.


MarijnS95:
This requires an MSRV bump to `1.77.0`.

PR #493, opened 2025-02-20
Bump jni-sys and thiserror

becmer:


MarijnS95:
We've already been pursuing the `jni-sys 0.4` upgrade in https://github.com/rust-mobile/ndk/pull/433 but won't commit to it until the matching `jni 0.22` drops to match. As you can tell from the download graph at https://crates.io/crates/jni-sys no-one uses `0.4.0` yet `0.3.0` still sees upwards of 80k downloads every working day.

Without justification I see no need to bump `thiserror` - which would be very hard to get done at once across the entire ecosystem - and will instead spend some time investigating whether we can support both using version ranges.

PR #495, opened 2025-02-22
Add bindings for querying and setting Android System Properties

MarijnS95:
The Android System Property API might not be very useful for most apps even though it still provides a read-only view of all properties but no rights to set any.  Since this is accessible via the NDK this crate should provide a clean and complete (including UB-free) wrapper implementation either way.

### DIFF
--- a/ndk/src/system_properties.rs
+++ b/ndk/src/system_properties.rs
@@ -1,0 +1,371 @@
+//! Bindings for [System Properties]
+//!
+//! [System Properties]: https://source.android.com/docs/core/architecture/configuration/add-system-properties
+
+use std::{
+    ffi::{c_char, c_void, CStr, CString, FromBytesWithNulError, FromVecWithNulError},
+    fmt,
+    mem::MaybeUninit,
+    ptr::NonNull,
+    str::{FromStr, Utf8Error},
+};
+#[cfg(feature = "api-level-26")]
+use std::{num::NonZeroU32, time::Duration};
+
+use thiserror::Error;
+
+use crate::utils::abort_on_panic;
+
+/// Possible failures returned by [`get_raw()`].
+#[derive(Debug, Error)]
+pub enum GetRawError {
+    #[error("Property is missing or empty")]
+    MissingOrEmpty,
+    #[error("Property value does not include a terminating NUL")]
+    ValueMissingNul(#[source] FromVecWithNulError),
+    #[error("`__system_property_get()` or `__system_property_read()` failed")]
+    Failed,
+}
+
+/// Internal helper to deduplicate the implementation between [`get_raw()`] and
+/// [`Property::read_raw()`].
+fn process_owned(get: impl FnOnce(*mut c_char) -> i32) -> Result<CString, GetRawError> {
+    // Pre-allocate a `Vec` which we can move to the user with the result
+    let mut value = Vec::<u8>::with_capacity(ffi::PROP_VALUE_MAX as usize);
+    let ret = get(value.as_mut_ptr().cast());
+    match ret {
+        0 => Err(GetRawError::MissingOrEmpty),
+        -1 => Err(GetRawError::Failed),
+        1.. => {
+            // TODO: This "smart" implementation leaves the user with a 92-byte allocation since
+            // set_len() currently doesn't shrink.  Any such operation would likely reallocate,
+            // making this have no advantage over the stack-local variant that allocates after the
+            // fact.
+            unsafe { value.set_len(ret as usize + 1) }
+            Ok(CString::from_vec_with_nul(value).map_err(GetRawError::ValueMissingNul)?)
+        }
+        _ => unreachable!("Status is unexpected integer {ret}"),
+    }
+}
+
+/// Returns the property value as an owned [`CString`] with possibly invalid UTF-8 [but no interior
+/// NULs].  The maximum length can be up to 92 ([`ffi::PROP_VALUE_MAX`]) including NUL terminator.
+///
+/// [but no interior NULs]: GetRawError::ValueMissingNul
+///
+/// See [`get()`] for a more convenient API that validates this string for UTF-8 and directly parses
+/// it into a [`FromStr`]-compatible type.
+///
+/// # Deprecation
+/// Deprecated since Android O (API level 26), use [`Property::find()`] with
+/// [`Property::read_callback()`] instead which does not have a limit on `value` nor `name` length.
+#[doc(alias = "__system_property_get")]
+pub fn get_raw(name: &CStr) -> Result<CString, GetRawError> {
+    process_owned(|value| unsafe { ffi::__system_property_get(name.as_ptr(), value) })
+}
+
+/// Possible failures returned by [`get()`].
+#[derive(Debug, Error)]
+pub enum GetError<T> {
+    #[error("Property is missing or empty")]
+    MissingOrEmpty,
+    #[error("Property value does not include a terminating NUL")]
+    ValueMissingNul(#[source] FromBytesWithNulError),
+    #[error("Property value does not contain valid UTF-8")]
+    Utf8Error(#[from] Utf8Error),
+    #[error("`__system_property_get()` or `__system_property_read()` failed")]
+    Failed,
+    #[error(transparent)]
+    ParseError(T),
+}
+
+/// Internal helper to deduplicate the implementation between [`get()`] and [`Property::read()`].
+fn process_parse<T: FromStr>(get: impl FnOnce(*mut c_char) -> i32) -> Result<T, GetError<T::Err>> {
+    let mut value = [0u8; ffi::PROP_VALUE_MAX as usize];
+    let ret = get(value.as_mut_ptr().cast());
+    match ret {
+        0 => Err(GetError::MissingOrEmpty),
+        -1 => Err(GetError::Failed),
+        1.. => {
+            let c_str = CStr::from_bytes_with_nul(&value[..ret as usize + 1])
+                .map_err(GetError::ValueMissingNul)?;
+            c_str.to_str()?.parse().map_err(GetError::ParseError)
+        }
+        _ => unreachable!("Status is unexpected integer {ret}"),
+    }
+}
+
+/// Returns the property value as a [`FromStr`]-parsed type from a source string of at most 92
+/// ([`ffi::PROP_VALUE_MAX`]) characters, including NUL terminator.
+///
+/// # Implementation details
+/// This is implemented without any up-front allocations like [`get_raw()`], but requires a trip
+/// through [`CStr`] and [`str`] (for calling [`FromStr::from_str()`]) meaning the resulting
+/// string has to be compliant with [`CStr`] ([no interior NULs]) and [`str`] ([valid UTF-8]).  In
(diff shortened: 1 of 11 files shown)


ndk/src/system_properties.rs
```diff
@@ -0,0 +1,327 @@
...
+/// Deprecated since Android O (API level 26), use [`Property::find()`] with
+/// [`Property::read_callback()`] instead which does not have a limit on `value` nor `name` length.
+#[doc(alias = "__system_property_get")]
+pub fn get_raw(name: &CStr) -> Result<CString, GetRawError> {
+    process_owned(|value| unsafe { ffi::__system_property_get(name.as_ptr(), value) })
+}
+
+/// Possible failures returned by [`get()`].
+#[derive(Debug, Error)]
+pub enum GetError<T> {
```
MarijnS95, line 70: Yikes, our MSRV is too old and this Rust version doesn't emit the `Debug` implementation behind a `T: Debug` boundary. Instead requiring us to force a `T: Debug` and `<T as FromStr>>::Err: Debug` bound throughout the codebase or not have `.unwrap()` capabilities at all (which requires the `Error` type to be `Debug`able):

https://github.com/rust-mobile/ndk/actions/runs/13476131028/job/37655556180?pr=495#step:6:108

MarijnS95: This also makes it inconvenient/impossible to use the type inside the error wrapper for https://github.com/rust-mobile/ndk/pull/479.

ndk/src/system_properties.rs
```diff
@@ -0,0 +1,326 @@
...
+            // set_len() currently doesn't shrink.  Any such operation would likely reallocate,
+            // making this have no advantage over the stack-local variant that allocates after the
+            // fact.
+            unsafe { value.set_len(ret as usize + 1) }
+            Ok(CString::from_vec_with_nul(value)?)
+        }
+    }
+}
+
+/// Returns the property value as an owned [`CString`] with possibly invalid UTF-8 [but no interor
```
Athosvk, line 50:
```suggestion
/// Returns the property value as an owned [`CString`] with possibly invalid UTF-8 [but no interior
```

ndk/src/system_properties.rs
```diff
@@ -0,0 +1,326 @@
...
+/// NULs].  The maximum length can be up to 92 ([`ffi::PROP_VALUE_MAX`]) including NUL terminator.
+///
+/// [but no interor NULs]: GetRawError::NulError
```
Athosvk, line 53:
```suggestion
/// [but no interior NULs]: GetRawError::NulError
```

ndk/src/system_properties.rs
```diff
@@ -0,0 +1,326 @@
...
+/// string has to be compliant with [`CStr`] ([no interior NULs]) and [`str`] ([valid UTF-8]).  In
+/// other words, parsing into a [`String`] will never contain interior NULs (and it is unknown and
+/// unlikely whether the property API allows for this).
+///
+/// [no interior NULs]: GetError::NulError
+/// [valid UTF-8]: GetError::Utf8Error
+///
+/// # Deprecation
+/// Deprecated since Android O (API level 26), use [`Property::find()`] with
+/// [`Property::read_callback()`] instead which does not have a limit on `value` nor `name` length.
```
Athosvk, line 111: Should the docs here maybe also mention the limit on what the limit on returned value is, like is done for the property name?

ndk/src/system_properties.rs
```diff
@@ -0,0 +1,326 @@
...
+
+/// Sets system property `name` to `value`, creating the system property if it doesn't already
+/// exist.
+#[doc(alias = "__system_property_set")]
+pub fn set(name: &CStr, value: &CStr) -> std::io::Result<()> {
+    let ret = unsafe { ffi::__system_property_set(name.as_ptr(), value.as_ptr()) };
+    match ret {
+        0 => Ok(()),
+        ..=-1 => Err(Error::from_raw_os_error(-ret)),
+        1.. => panic!("Unexpected non-zero non-negative return value `{ret}`"),
```
Athosvk, line 125: Can't you use `status_to_io_result` here?

ndk/src/system_properties.rs
```diff
@@ -0,0 +1,326 @@
...
+///
+/// [no interior NULs]: GetError::NulError
+/// [valid UTF-8]: GetError::Utf8Error
+///
+/// # Deprecation
+/// Deprecated since Android O (API level 26), use [`Property::find()`] with
+/// [`Property::read_callback()`] instead which does not have a limit on `value` nor `name` length.
+#[doc(alias = "__system_property_get")]
+pub fn get<T: FromStr>(name: &CStr) -> Result<T, GetError<T::Err>> {
+    process_parse(|value| unsafe { ffi::__system_property_get(name.as_ptr(), value) })
```
Athosvk, line 114: One thing we could add here is that we actually assert on `name.len() <= PROP_NAME_MAX`. What do you think?

ndk/src/system_properties.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+/// Sets system property `name` to `value`, creating the system property if it doesn't already
+/// exist.
+#[doc(alias = "__system_property_set")]
+pub fn set(name: &CStr, value: &CStr) -> std::io::Result<()> {
+    let ret = unsafe { ffi::__system_property_set(name.as_ptr(), value.as_ptr()) };
```
Athosvk, line 121: If I check the source for `__system_property_set`, this is not allowed to exceed both `PROP_VALUE_MAX` and `PROP_NAME_MAX` lens. Probably should specify that in the docs and maybe assert as well.

If we don't assert, `__system_property_set` jjust returns `-1` which is cascaded to a generic error

Athosvk (commented):
Looks really helpful. The only thing for the raw APIs that have limtis is maybe considering to add some asserts, since those specific errors is not something we can catch from what is returned by the ndk itself

ndk/src/system_properties.rs
MarijnS95: This limit only applied pre-Android-O and only for out-allocated strings (because they don't have a "query length" -> "allocate" -> "query data" API). This API can definitely be used to query properties with names longer than `32` chars:

```
RustStdoutStderr: [ndk-examples/src/lib.rs:12:5] get_raw(c"dev.mnt.rootdisk.data_mirror.storage_area") = Ok(
RustStdoutStderr:     "sda",
RustStdoutStderr: )
```

Reading any **value** that is longer than 91 gives this instead:

```
libc    : The property "ro.build.version.known_codenames" has a value with length 301 that is too large for __system_property_get()/__system_property_read(); use __system_property_read_callback() instead
...
RustStdoutStderr: [ndk-examples/src/lib.rs:36:17] pi.name = "ro.build.version.known_codenames"
RustStdoutStderr: [ndk-examples/src/lib.rs:37:17] p.read_raw() = Ok(
RustStdoutStderr:     "Must use __system_property_read_callback() to read",
RustStdoutStderr: )
```

ndk/src/system_properties.rs
MarijnS95: That's already covered in the first paragraph / "`Returns xxx` title" of this function?

Perhaps I should remove `` and `name` `` here as we've never introduced a limit on `name` length in this function and as demonstrated elsewhere, there is none for `get()` (only for `__system_property_read()`).

ndk/src/system_properties.rs
Athosvk: Ah nevermind, misunderstood that part then, clear!

ndk/src/system_properties.rs
```diff
@@ -0,0 +1,326 @@
...
+    /// directly parses it into a [`FromStr`]-compatible type.
+    ///
+    /// # Deprecation
+    /// Deprecated since Android O (API level 26), use [`Self::read_callback()`] instead which does
+    /// not have a limit on `value` nor `name` length.
+    #[doc(alias = "__system_property_read")]
+    pub fn read_raw(&self) -> Result<CString, GetRawError> {
+        process_owned(|value| unsafe {
+            // TODO: should we return the name of ffi::PROP_NAME_MAX?
+            ffi::__system_property_read(self.0.as_ptr(), std::ptr::null_mut(), value)
```
MarijnS95, line 217: Notice that I still have to consider whether to always allocate + move a name here in `read_raw()` and `read()`.

This may seem silly when using `let prop = Property::find(c"we.know.the.name")`, but it's the only API to get names out of `Property::foreach(|prop| prop.how_to_get_the_name())` before Android O where one can call `prop.read_callback()` and get the name (without limitations on length).

ndk/src/system_properties.rs
MarijnS95: Yeah with `set()` this `last_os_error()` is actually [`13` which is `EACCESS`](https://github.com/torvalds/linux/blob/d082ecbc71e9e0bf49883ee4afd435a77a5101b6/include/uapi/asm-generic/errno-base.h#L17) (`Permission Denied`, not `EPERM` which is `Operation not permitted`).

ndk/src/system_properties.rs
MarijnS95: As it turns out this wasn't a broken bound on `Debug` but on the generated `impl`s from `thiserror::Error`. This was fixed a couple years ago but our MSRV CI test intentionally downgrades our crates: we just had to bump the minimum bound a little to `1.0.29`. As a bonus we not only no longer

ndk/src/system_properties.rs
MarijnS95: I've taken a dependency on `nix` just to clear the error in this very infrequently used path. Sensible or overdone?

ndk/src/system_properties.rs
MarijnS95: The most nice and trivial implementation uses an initialized slice of `ffi::PROP_NAME_MAX` characters, so that we can safely call `CStr::from_bytes_until_nul()` and copy into a host allocation. But that requires MSRV 1.69, which must be an acceptable bump but I still haven't set up a proper MSRV policy to follow (besides chugging along with `winit`).

ndk/src/system_properties.rs
MarijnS95: Since the latest `nix 0.29` is also requiring an MSRV bump to `1.89` and this function is very unlikely to be used by most `ndk` consumers, I've instead reverted it and created a wrapper type over `io::Error` specifically for `set()`, documenting and format-printing that its value may be stale/bogus.

ndk/src/system_properties.rs
MarijnS95: Replaced this with an ordinary call to `CStr::from_ptr()` on a `MaybeUninit` slice, like we do in other places where I described to use `CStr::from_bytes_until_nul()` when the MSRV is bumped.

Notice also that this doesn't require a `NulError` for the name, and in hindsight I'm considering to remove it for the `value` too: the contract with the NDK specifies that it should have written a NUL terminator within the property limit, and returned the length of the string (excluding NUL terminator). If that's violated, which should pretty much never happen, we might as well `unwrap` and panic rather than bubbling a "pretty much unreachable" error up to the user to match in `thiserror`.

At least that's the guidance I've been following over the years when deciding to prefer returning `Result::Err` over a `panic!()`.

PR #499, opened 2025-03-14
ndk: Add `ASurfaceControl` bindings

MarijnS95:
Draft to ensure visibility into the impending addition of these bindings. This first requires a few pre-PRs to update the NDK version `ndk-sys` was generated from (which now contains header fixes that make us able to parse `surface_control.h` as valid C), together with also immediately binding the new API made available in such updates.


phiresky:
Is there any updates on this? :) I just made a mistake while hand-writing a binding to ASurfaceTransaction_setGeometry

Issue #309, opened 2022-07-02
Unify lifetime handling for `NativeWindow` and `HardwareBuffer`


MarijnS95:
See https://github.com/rust-windowing/android-ndk-rs/pull/296#issuecomment-1153288154, there are multiple ways we can approach this.

Something based on borrows (Like how `String` + `&str` work together) has my preference, but we can't give any sensible lifetime to for example what `fn native_window()` returns (which we currently have to forcibly `_acquire` to make it valid). Alas, we'll probably remove those `static` getters anyway in light of recent suggestions and to support multi-`Activity`, so this specific case wouldn't affect us anymore. At which point a borrow-based approach might be even better.

MarijnS95:
Looking into this again, handing out `&'_ Type` would be preferred (i.e. to match Rust's `&str` vs `String` deisgn), but we need storage for `Type` in this case which is impossible. Hence `HardwareBufferRef` solves this by simply owning a `HardwareBuffer` and implementing `Drop` for itself.

We could possibly improve this API by adding a lifetime to `HardwareBuffer`, which `HardwareBufferRef` hides by setting it to `'static` internally because _it_ controls the lifetime. That would allow us to derive `Clone, Copy` on the non-refcounted `HardwareBuffer`, as long as `Deref for HardwareBufferRef` returns a target of `(&'a self) -> &'a HardwareBuffer<'a>` and _not_ `'static`. On the other hand the current implementation (without lifetime on `HardwareBuffer`) already captures a very similar semantic when `&'_ HardwareBuffer` exists, except that `HardwareBuffer::from_ptr()` gives you an owned and lifetime-less `HardwareBuffer`.

PR #503, opened 2025-04-25
input: Split out `InputEventJava` into `Motion` and `Key` with `Deref`

### LINKED ISSUE #502: Separate `InputEventJava` into `Key`/`Motion` `Java` structs to `Deref` into (and provide methods on) their native variants
Currently InputEventJava doesn't have any available methods, can we modify it to an enumeration?

```rust
pub enum InputEventJava {
    MotionEvent(MotionEventJava),
    KeyEvent(KeyEventJava)
}

pub struct MotionEventJava(MotionEvent);

impl Deref for MotionEventJava {
    type Target = MotionEvent;

    fn deref(&self) -> &Self::Target {
        todo!()
    }
}
impl Drop for MotionEventJava {
    fn drop(&mut self) {
        todo!()
    }
}

pub struct KeyEventJava(KeyEvent);
impl Deref for KeyEventJava {
    type Target = KeyEvent;

    fn deref(&self) -> &Self::Target {
        todo!()
    }
}
impl Drop for KeyEventJava {
    fn drop(&mut self) {
        todo!()
    }
}

```

MarijnS95:
Fixes #502

When I implemented the `from_java()` constructors for `MotionEvent` and `KeyEvent` in #456, I used a single `InputEventJava` wrapper since that could wrap the existing `enum InputEvent` and only need a single destructor calling `AInputEvent_release()` (which must only be called when the input event was created from Java).

This however requires existing callers to `MotionEvent::from_java()` and `KeyEvent::from_java()` to _unwrap_/unpack that nested `enum InputEvent` again in order to get access to the underlying native methods, despite already calling a specific constructor method (since Android makes a distinction between both types).

Not that that is even reachable, since the nested `InputEvent` member was private and there was no `Deref` into `&InputEvent` anywhere making it impossible to use this API in any meaningful way.

Solve both issues by splitting the `struct` into a `Motion` and `Key` variant, and implement `Deref` on both to their respective type.  No wrapper `InputEventJava` remains since there does not appear to be any reason to pass both of these types into a single API.

### DIFF
--- a/ndk/src/event.rs
+++ b/ndk/src/event.rs
@@ -10,7 +10,7 @@
 //! [`android.view.MotionEvent`]: https://developer.android.com/reference/android/view/MotionEvent
 //! [`android.view.KeyEvent`]: https://developer.android.com/reference/android/view/KeyEvent
 
-use std::ptr::NonNull;
+use std::{ops::Deref, ptr::NonNull};
 
 #[cfg(feature = "api-level-31")]
 use jni_sys::{jobject, JNIEnv};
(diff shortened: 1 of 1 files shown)


MarijnS95:
I'm going to refactor this PR a little bit further, at least to remove the Java moniker and change it to a proper `*Ref` type (as we should also do with some other types: #309). As it turns out _at least_ `SurfaceControl`'s new `AInputReceiver` (that I'm adding on top of #499 / #500) callbacks are documented to require a call to `AInputEvent_release()` on the key/motion events that these callbacks receive:

https://developer.android.com/ndk/reference/group/native-activity#ainputreceiver_onkeyevent
https://developer.android.com/ndk/reference/group/native-activity#ainputreceiver_onmotionevent

MarijnS95:
I may or may not end up renaming it in a followup. `AInputEvent_release()` still documents it specifically takes Java `Key/MotionEvent`s, so perhaps these callbacks specifically wrap Java components. After all, the necessary `AInputTransferToken` is also specific to Java-land.

PR #505, opened 2025-08-25
Fix link to ndk-examples in README.md

AnirudhVyas:
This is a minor documentation change.
NDK examples link was broken, added it.


AnirudhVyas:
I don't understand this was made with the main branch, but there are clippy issues. Should I patch? let me know

PR #507, opened 2025-10-30
ndk/audio: swap AAUDIO_DIRECTION_{INPUT,OUTPUT} doc comments

pr2502:
the doc comments were accidentally swapped when written in Rust, swap them to match the oficial documentation at <https://developer.android.com/ndk/reference/group/audio#enumerations_1> and the actual behaviour of the AAudio interface

### DIFF
--- a/ndk/src/audio.rs
+++ b/ndk/src/audio.rs
@@ -111,10 +111,10 @@ pub enum AudioContentType {
 #[doc(alias = "aaudio_direction_t")]
 #[non_exhaustive]
 pub enum AudioDirection {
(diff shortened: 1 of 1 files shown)


MarijnS95:
That happened in https://github.com/rust-mobile/ndk/pull/66, likely because the `enum` variants were listed in opposite order when compared to the NDK header files.

MarijnS95:
New Rust lints will be fixed on `master` later today after 1.91 releases, there's nothing you'll have to do for this to be merged :+1:

PR #509, opened 2025-12-16
Exclude development scripts from published package

weiznich:
During a dependency review we noticed that the ndk-sys crate includes various development scripts. These development scripts shouldn't be there as they might, at some point become problematic. As of now they prevent any downstream user from enabling the `[bans.build.interpreted]` option of cargo deny.

I opted for using an explicit include list instead of an exclude list to prevent these files from being included in the published packages to make sure that everything that's included is an conscious choice.


weiznich:
@MarijnS95 Any chance to get your look at this?

MarijnS95:
This package will get rewritten on the next release to also accommodate for various upstream header changes. No direct plans to re-release with just this packaging change.

